PR #441, opened 2017-01-05
STREAMLINE-575: Support event time based windowing by extracting timestamp from streamline event

arunmahadevan:
Depends on https://github.com/apache/storm/pull/1820

This is same as PR - https://github.com/hortonworks/streamline/pull/385 that was merged earlier and reverted.


harshach:
Merged into master

PR #480, opened 2017-01-20
Refactor StreamlineEvent and avoid duplicate copies

arunmahadevan:
1. Provide a builder api that can be directly used to put the key-values without the need for a separate map.
2. Avoid duplicate copies as far as possible.
3. Refactor StreamlineEventImpl usages.


harshach:
Merged into master

Issue #483, opened 2017-01-25
Docker Support


jdye64:
Having Docker support for Streamline would be convenient for both development and customers. Since a proper environment requires RDBMS, Storm, Kafka, etc it makes sense that most customers would have a docker-compose script to orchestrate deployment. The scope of this would be simply to provide the isolated image for Streamline. Would be best to use the Docker Maven plugin so that a profile can be created for building Docker images. This would allow people not interested in Docker to ignore it and not require the docker engine on their machine for building.

PR #484, opened 2017-01-25
Github #483 - Docker Support

jdye64:
Initial support for building Docker images via Maven. The image can be
built as part of the build process by using the Maven Profile
“-Pdocker” and does expect that “streamline-dist” has been previously
built. The easiest way to ensure this is always including “-Pdist” with
the Maven command as well. I'm not sure of this teams contribution
guidelines so I apologize if you want a Github issue or JIRA attached
to this let me know and I'll be glad to add one. This was the same
approach we used for NiFi and find it pretty handy. This PR is for
Github Issue #483


harshach:
Thanks @jdye64 . Merged into master

PR #492, opened 2017-02-06
(refactor) Move out metadata and hint provider impls to 'streamline-cluster' module

HeartSaVioR:
They're not used from 'streamline-catalog' and actually not related to.

I'm planning to work on STREAMLINE-545 on top of this branch, but if someone would like to keep as it is, we can discuss and STREAMLINE-545 may be rebased.


harshach:
@HeartSaVioR can you upmerge .

HeartSaVioR:
@harshach 
#504 contains this. Could we close this, or do you want to apply this PR before applying #504?

PR #498, opened 2017-02-10
STREAMLINE-550: Convert streamcatalogstorable into annotations

arunmahadevan:
1. Add StreamlineEntity annotation and marked streamline entities.
2. Load streamline entities from StreamlineApplication
3. Remove *storable.props and *CatalogService specific loading


harshach:
merged into master.

PR #503, opened 2017-02-15
Streamline 643: Postgresql support

arunmahadevan:
Fixes on top of Harsha's patch https://github.com/hortonworks/streamline/pull/502

Tested bootstrap, topology creation and deployment with different components and works fine.

@harshach please review, also you may close your PR since this patch builds on top of those changes.


bootstrap/sql/postgresql/create_tables.sql
```diff
@@ -0,0 +1,344 @@
+-- THE NAMES OF THE TABLE COLUMNS MUST MATCH THE NAMES OF THE CORRESPONDING CLASS MODEL FIELDS;
+
+CREATE TABLE IF NOT EXISTS dashboard (
```
priyank5485: In one of the other PRs you have renamed these tables and few classes to a standard. Do we want to do that here too?

storage/core/src/main/java/com/hortonworks/streamline/storage/impl/jdbc/JdbcStorageManager.java
```diff
@@ -233,6 +234,9 @@ public void init(Map<String, Object> properties) {
             case "mysql":
```
priyank5485: Do we still want to keep mysql here?

storage/tool/src/main/java/com/hortonworks/streamline/storage/tool/StorageProviderConfiguration.java
```diff
@@ -38,6 +40,7 @@ private StorageProviderConfiguration(String driverClass, String url, String user
         this.delimiter = delimiter;
     }
 
+
     public static StorageProviderConfiguration mysql(String driverClass, String url) {
```
priyank5485: Should we remove mysql related varialbes and methods?

storage/tool/src/main/java/com/hortonworks/streamline/storage/tool/StorageProviderConfigurationReader.java
```diff
@@ -103,4 +108,24 @@ private static StorageProviderConfiguration readPhoenixProperties(Map<String, Ob
...
+     * storageProviderConfiguration:
+     *   providerClass: "com.hortonworks.streamline.storage.impl.jdbc.JdbcStorageManager"
+     *   properties:
+     *     db.type: "postgresql"
+     *     queryTimeoutInSecs: 30
+     *     db.properties:
+     *     dataSourceClassName: "org.postgresql.ds.PGSimpleDataSource"
+     *     dataSource.url: "jdbc:postgresql://localhost/test"
+     *     dataSource.user: "postgres"
+     *     dataSource.password: "postgres"
```
priyank5485: Should we add this commented configuration to yaml file?

priyank5485:
+1 once the comments are addressed.

bootstrap/sql/postgresql/create_tables.sql
arunmahadevan: Old names were used for compatibility with Pojos and for testing. After https://github.com/hortonworks/streamline/pull/512 is merged, I will rename these.

storage/tool/src/main/java/com/hortonworks/streamline/storage/tool/StorageProviderConfiguration.java
arunmahadevan: I think we still want to keep mysql support, so we can retain this.

storage/tool/src/main/java/com/hortonworks/streamline/storage/tool/StorageProviderConfigurationReader.java
arunmahadevan: Similar config was earlier there but was removed from the yaml file so I did not add it again.

storage/core/src/main/java/com/hortonworks/streamline/storage/impl/jdbc/JdbcStorageManager.java
arunmahadevan: I think we still want to keep mysql support, so we can retain this.

arunmahadevan:
@priyank5485 after #512 is merged, I will rename the postgres tables and update this PR. Then we can merge this.

HeartSaVioR:
Would we want to add `streamline.postgre.yaml` so that users can start from copying and modifying?

bootstrap/sql/postgresql/create_tables.sql
priyank5485: PR 512 is merged. Did you rename these in this PR?

storage/core/src/main/java/com/hortonworks/streamline/storage/impl/jdbc/JdbcStorageManager.java
priyank5485: Ok. In that case there was some other place in code where mysql was removed as part of changeset of this PR.  I think it was an "if" check which enforced that only postgre is supported. Not sure if we need to change that but can you can double check we have it right to support mysql?

priyank5485:
@HeartSaVioR Yes. We should add streamline.postgre.yaml file.

arunmahadevan:
Closing this PR since new commits are not getting updated here.

Raised, https://github.com/hortonworks/streamline/pull/514

PR #506, opened 2017-02-18
(POC) STREAMLINE-381 Test Source, Test Sink components

HeartSaVioR:
TODO
----
- [x] : add unit tests

Summary
----
* Introduce 'test run' for topology actions
  * replace sources with test sources and sinks with test sinks from topology DAG
    * test source just emits test records (user input) to the downstream
    * test sink just writes received events to local file (generated uuid)
  * run topology via local mode, with 30s shutdown
  * create API endpoints for requesting test run and providing results

API endpoint
----
1. Run topology on test mode

- path: `/topologies/:topologyId/actions/testrun`
- method: POST
- input param: JSON (application/json)

```
{
  "<source name>": {
    "<stream name>": [
      <json format of record 1>,
      <json format of record 2>,
      ...
      <json format of record N>
    ],
    ...
    "<stream name>": [
      ...
    ]
  },
  ...
}
```

The JSON input should contain all sources in topology.

- Response

```
{
  "<sink name>": {
    "uuid": "<generated uuid value>",
    "eventDumpLink": "<full url path for retrieving result events for that sink>"
  },
  ...
}
```

The JSON response will contain all sinks in topology.

2. Retrieve result events of specific sink on topology test run

- path: `/topologies/actions/testrun/result/:uuid`
- method: GET

Full URI will be provided from API response of 1.

- Response

```
{
  "entities": [
    <json format of event 1>,
    <json format of event 1>,
    ...
    <json format of event N>,
  ]
}
```

Note
----
This is a patch for first iteration of [STREAMLINE-381](https://hwxiot.atlassian.net/browse/STREAMLINE-381). I'll file issues for each iterations soon.

There're something to consider and discuss:

1.
Since we're running test topology as local mode, run command doesn't return unless local cluster shutdown is complete. Same thing applies to the API endpoint, so that API will take about 1 minute (or more) to respond, and even hang if Storm has a bug on local cluster shutdown (e.g. non-daemon threads alive).

I'm thinking about asynchronous run, but then we need to have state of running status (maybe need to store to a new table), and UI needs to check the status periodically (or even users need to check it periodically)

2.
Ideally process should start shut down when all test records get processed, but Flux runner doesn't have the feature, and even we implement our own, specific signal should be made for accomplish this.
This patch just sets 30 seconds to shutdown local cluster (Flux provides this feature) which is enough for many cases, but still wondering whether we need to expose this so that users can change waiting seconds.

### DIFF
--- a/conf/streamline-dev.yaml
+++ b/conf/streamline-dev.yaml
@@ -11,6 +11,8 @@ modules:
       #change the below to the path on your local machine
       streamlineStormJar: /tmp/streamline-runtime-storm-0.1.0-SNAPSHOT.jar
       stormHomeDir: /usr/local/Cellar/storm/0.10.0/
+      # directory to store the results of topology test run
+      topologyTestRunResultDir: /tmp
       # schema registry configuration
       schemaRegistryUrl: "http://localhost:9090/api/v1"
       mavenRepoUrl: "hwx-public^http://repo.hortonworks.com/content/groups/public/,hwx-private^http://nexus-private.hortonworks.com/nexus/content/groups/public/"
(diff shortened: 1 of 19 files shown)


HeartSaVioR:
I removed and re-forked the repo, so the PR is invalid now. I'll create new PR.

PR #508, opened 2017-02-22
STREAMLINE-720: Script and endpoint to update topology component bund…

priyank5485:
…les.


harshach:
+1. Merged into master

PR #491, opened 2017-02-06
STREAMLINE-631 Pick runtime jar in libs dir to the default streamline…

HeartSaVioR:
…StormJar config

* Replace `${STREAMLINE_HOME}` in streamlineStormJar to the streamline home directory
  * streamline-server-start.sh finds home directory and pass via Java system property

PR #510, opened 2017-02-24
STREAMLINE-631 Pick runtime jar in libs dir to the default streamlineStormJar config

HeartSaVioR:
* Replace ${STREAMLINE_HOME} in streamlineStormJar to the streamline home directory
  * streamline-server-start.sh finds home directory and pass via Java system property
* when streamlineStormJar is not present, it will take default path

@harshach This is updated PR of #491. Please review.

### DIFF
--- a/bin/streamline-server-start.sh
+++ b/bin/streamline-server-start.sh
@@ -19,7 +19,22 @@ then
         echo "USAGE: $0 [-daemon] streamline.yaml"
         exit 1
 fi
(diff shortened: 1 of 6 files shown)


priyank5485:
+1 Can you please make your repo public or raise another PR by forking from public repo?

HeartSaVioR:
@priyank5485 This is from re-forked repo. Could you try merging and let me know if you can't?

PR #504, opened 2017-02-16
STREAMLINE-545 Add services to service pool manually (following topology component bundle)

HeartSaVioR:
NOTE: This PR is on top of #492.

This PR is modification of #496: introducing `service bundle` which is similar to `topology component bundle`. Bundles for essential services will be registered via bootstrap.sh.

Please refer below to see how to register service bundle / register service manually with using new API endpoints. (Picked from changeset of Cluster.md)

---
### Create Service Bundle

`POST /api/v1/catalog/servicebundles`

**Sample Input**

```json
{
  "name": "STORM",
  "registerClass": "com.hortonworks.streamline.streams.cluster.register.impl.StormServiceRegisterer",
  "serviceUISpecification": {
    "fields": [
      {
        "uiName": "UI server hostname",
        "fieldName": "uiServerHostname",
        "isOptional": false,
        "tooltip": "Hostname of Storm UI server",
        "type": "string"
      },
      {
        "uiName": "Nimbuses hostnames",
        "fieldName": "nimbusesHostnames",
        "isOptional": false,
        "tooltip": "Hostnames of Nimbuses",
        "type": "array.string"
      },
      {
        "uiName": "storm.yaml",
        "fieldName": "storm.yaml",
        "isOptional": false,
        "tooltip": "Upload storm.yaml file which Storm cluster is using",
        "type": "file"
      }
    ]
  }
}
```
   
**Success Response**

    HTTP/1.1 201 Created
    Content-Type: application/json
    
```json
{
  "id": 1,
  "name": "STORM",
  "timestamp": 1487228012093,
  "serviceUISpecification": {
    "fields": [
      {
        "uiName": "UI server hostname",
        "fieldName": "uiServerHostname",
        "isUserInput": true,
        "tooltip": "Hostname of Storm UI server",
        "isOptional": false,
        "type": "string"
      },
      {
        "uiName": "Nimbuses hostnames",
        "fieldName": "nimbusesHostnames",
        "isUserInput": true,
        "tooltip": "Hostnames of Nimbuses",
        "isOptional": false,
        "type": "array.string"
      },
      {
        "uiName": "storm.yaml",
        "fieldName": "storm.yaml",
        "isUserInput": true,
        "tooltip": "Upload storm.yaml file which Storm cluster is using",
        "isOptional": false,
        "type": "file"
      }
    ]
  },
  "registerClass": "com.hortonworks.streamline.streams.cluster.register.impl.StormServiceRegisterer"
}
```

### List Services Bundle

[...]

PR #511, opened 2017-02-24
STREAMLINE-545 Add services to service pool manually (following topology component bundle)

HeartSaVioR:
NOTE: This PR is on top of #492. This is same as #504 but from re-forked repo.

This PR is modification of #496: introducing `service bundle` which is similar to `topology component bundle`. Bundles for essential services will be registered via bootstrap.sh.

Please refer below to see how to register service bundle / register service manually with using new API endpoints. (Picked from changeset of Cluster.md)

---
### Create Service Bundle

`POST /api/v1/catalog/servicebundles`

**Sample Input**

```json
{
  "name": "STORM",
  "registerClass": "com.hortonworks.streamline.streams.cluster.register.impl.StormServiceRegisterer",
  "serviceUISpecification": {
    "fields": [
      {
        "uiName": "UI server hostname",
        "fieldName": "uiServerHostname",
        "isOptional": false,
        "tooltip": "Hostname of Storm UI server",
        "type": "string"
      },
      {
        "uiName": "Nimbuses hostnames",
        "fieldName": "nimbusesHostnames",
        "isOptional": false,
        "tooltip": "Hostnames of Nimbuses",
        "type": "array.string"
      },
      {
        "uiName": "storm.yaml",
        "fieldName": "storm.yaml",
        "isOptional": false,
        "tooltip": "Upload storm.yaml file which Storm cluster is using",
        "type": "file"
      }
    ]
  }
}
```
   
**Success Response**

    HTTP/1.1 201 Created
    Content-Type: application/json
    
```json
{
  "id": 1,
  "name": "STORM",
  "timestamp": 1487228012093,
  "serviceUISpecification": {
    "fields": [
      {
        "uiName": "UI server hostname",
        "fieldName": "uiServerHostname",
        "isUserInput": true,
        "tooltip": "Hostname of Storm UI server",
        "isOptional": false,
        "type": "string"
      },
      {
        "uiName": "Nimbuses hostnames",
        "fieldName": "nimbusesHostnames",
        "isUserInput": true,
        "tooltip": "Hostnames of Nimbuses",
        "isOptional": false,
        "type": "array.string"
      },
      {
        "uiName": "storm.yaml",
        "fieldName": "storm.yaml",
        "isUserInput": true,
        "tooltip": "Upload storm.yaml file which Storm cluster is using",
        "isOptional": false,
        "type": "file"
      }
    ]
  },
  "registerClass": "com.hortonworks.streamline.streams.cluster.register.impl.StormServiceRegisterer"
}
```

### List Services Bundle

`GET /api/v1/catalog/servicebundles`

### Get Service Bundle

`GET /api/v1/catalog/servicebundles/{id}`

### Get Service Bundle (by service name)

`GET /api/v1/catalog/servicebundles/name/{serviceName}`

### Delete Service Bundle

`DELETE /api/v1/catalog/servicebundles/{id}`

---
### Register Service manually

Streamline also provides bundle-registered service and relevant components manually.
User should specify existing cluster to make new service belong to.

`POST /api/v1/clusters/:clusterId/services/register/:serviceName`

`Content-Type` of the API is `multipart/form-data`, since user needs to upload config files for service.

It receives two types of parameters - `config` and `config files`:

* config: component information JSON, same structure of topology component bundle
* config files: service relevant configuration files to upload

Each service bundle defines its specification (config), similar to topology component bundle.
Unless the type of field is `file`, the value of field should be passed to config json.

The sample format for config:

```json
{
  "properties": {
    "uiServerHostname": "jlim-streamlin-3.openstacklocal", 
    "nimbusesHostnames": "jlim-streamlin-2.openstacklocal,jlim-streamlin-3.openstacklocal"
  }
}
```

If the specification defines the type of field as `file`, you need to provide the value from separate parameter.
The parameter name is same as field name on definition.

Here is the sample STORM service registration via curl:

```
curl -v -H "Content-Type: multipart/form-data" -X POST \
-F config='{"properties": {"uiServerHostname": "storm-1", "nimbusesHostnames": "storm-2,storm-3"}}' \
-F storm.yaml=@./storm.yaml http://localhost:8080/api/v1/catalog/clusters/1/services/register/STORM
```

@harshach @shahsank3t 
This is follow-up PR for STREAMLINE-545 which follows the style `topology component bundle`. Please let me know which is better. (#496 or this) 
Thanks in advance!


HeartSaVioR:
Upmerged.

harshach:
@HeartSaVioR sorry missed your comment. Can you please upmerge one more time will merge it in right away.

HeartSaVioR:
@harshach Rebased. Also squashed commits but separating commits between refactor and STREAMLINE-545.

HeartSaVioR:
@harshach Ping for kindly reminder.

bootstrap/services/storm-bundle.json
```diff
@@ -0,0 +1,29 @@
...
+        "uiName": "Nimbuses hostnames",
+        "fieldName": "nimbusesHostnames",
+        "isOptional": false,
+        "tooltip": "Hostnames of Nimbuses",
+        "type": "array.string"
+      },
+      {
+        "uiName": "storm.yaml",
+        "fieldName": "storm.yaml",
+        "isOptional": false,
```
arunmahadevan: Why is this mandatory ? If users dont want to override cant they run with defaults? 

Also I am  not clear on how and where this storm.yaml file would be used.

HeartSaVioR: It will be used for extracting the Storm cluster's configuration to let metadata API utilizes the config (so as bundle hint). It doesn't be used for overriding Storm config for topology submission.
Port informations are extracted from the config file.

HeartSaVioR:
@arunmahadevan 

Looks like comment doesn't seem to be added.

1. It just borrows the naming pattern and not related to topology component bundle.

2. Yes that's the goal of this issue.

3. That is actually similar approach for importing Ambari cluster. Simple to input but someone might feel bothered to upload the file since it might be only placed to cluster nodes.

We could define mandatory fields for each service, but it also means that we need to maintain the list of mandatory fields, and we should find a way to force user to input the value when mandatory fields are added later. So it's ideal but not that simple to handle.

Btw, we still need to receive HDFS / HBase / Hive config file since it should be passed to the topology submission.

hmcl:
@HeartSaVioR i would like to review this patch since it my metadata task is going to depend on this. I will do it first thing tomorrow.

HeartSaVioR:
@hmcl Sure please go ahead. This change is needed mainly for metadata and component bundle hint. 
Please let me know which configurations are needed for accessing secured services. If you don't know about them for now, let's force to put configuration file and change it later.

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/HBaseServiceRegisterer.java
```diff
@@ -0,0 +1,73 @@
...
+            }
+            return false;
+        }).count();
+
+        return validConfigFileCount == 1;
+    }
+
+    @Override
+    protected boolean validateServiceConfiguationsViaFlattenMap(Map<String, String> configMap) {
+        // FIXME: which configuration KVs are mandatory?
```
hmcl: let's file a JIRA for this and remove the FIXME, or fix it.

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/KafkaServiceRegisterer.java
```diff
@@ -0,0 +1,103 @@
...
+    protected boolean validateServiceConfiguationsViaFlattenMap(Map<String, String> configMap) {
+        // requirements
+        // 1. zookeeper.connect should be available in kafka-broker
+        // if it exists, it should be within kafka-broker since we are allowing only one 'kafka-broker'
+        return configMap.containsKey(KAFKA_PROPERTY_ZOOKEEPER_CONNECT);
+    }
+
+    private Component createKafkaBrokerComponent(Config config, Map<String, String> flattenConfigMap) {
+        if (!config.contains(PARAM_KAFKA_BROKER_HOSTNAMES)) {
+            throw new IllegalArgumentException("Required parameter " + PARAM_KAFKA_BROKER_HOSTNAMES + " not presented.");
```
hmcl: present

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/StormServiceRegisterer.java
```diff
@@ -0,0 +1,129 @@
...
+
+    @Override
+    protected boolean validateServiceConfiguationsViaFlattenMap(Map<String, String> configMap) {
+        // for now, every requirements will be checked from components
+        return true;
+    }
+
+    private Component createStormUIServerComponent(Config config, Map<String, String> flattenConfigMap) {
+        if (!config.contains(PARAM_STORM_UI_SERVER_HOSTNAME)) {
+            throw new IllegalArgumentException("Required parameter " + PARAM_STORM_UI_SERVER_HOSTNAME + " not presented.");
```
hmcl: present

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/StormServiceRegisterer.java
```diff
@@ -0,0 +1,129 @@
...
+        }
+
+        Component stormUiServer = new Component();
+        stormUiServer.setName(COMPONENT_STORM_UI_SERVER);
+        stormUiServer.setHosts(Collections.singletonList(stormUiServerHost));
+        environmentService.injectProtocolAndPortToComponent(flattenConfigMap, stormUiServer);
+        return stormUiServer;
+    }
+
+    private Component createNimbusComponent(Config config, Map<String, String> flattenConfigMap) {
```
hmcl: flatConfigMap

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/StormServiceRegisterer.java
```diff
@@ -0,0 +1,129 @@
...
+        if (!config.contains(PARAM_NIMBUS_HOSTNAMES)) {
+            throw new IllegalArgumentException("Required parameter " + PARAM_NIMBUS_HOSTNAMES + " not presented.");
```
hmcl: present

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/ZookeeperServiceRegisterer.java
```diff
@@ -0,0 +1,94 @@
...
+
+    @Override
+    protected boolean validateServiceConfiguationsViaFlattenMap(Map<String, String> configMap) {
+        // for now, every requirements will be checked from components
+        return true;
+    }
+
+    private Component createZookeeperServerComponent(Config config, Map<String, String> flattenConfigMap) {
+        if (!config.contains(PARAM_ZOOKEEPER_SERVER_HOSTNAMES)) {
+            throw new IllegalArgumentException("Required parameter " + PARAM_ZOOKEEPER_SERVER_HOSTNAMES + " not presented.");
```
hmcl: present

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/service/EnvironmentService.java
```diff
@@ -434,6 +439,85 @@ public NamespaceServiceClusterMapping addOrUpdateServiceClusterMapping(
...
+
+                    if (!protocol.isEmpty()) {
+                        component.setProtocol(protocol);
+                    }
+                    if (!portStr.isEmpty()) {
+                        try {
+                            component.setPort(Integer.parseInt(portStr));
+                        } catch (NumberFormatException e) {
+                            LOG.warn(
+                                    "Protocol/Port information [{}] for component {} doesn't seem to known format [{}]."
```
hmcl: has illegal format

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/service/EnvironmentService.java
```diff
@@ -434,6 +439,85 @@ public NamespaceServiceClusterMapping addOrUpdateServiceClusterMapping(
...
+                                            + "skip assigning...", value, component.getName(), confMap.getParsePattern());
```
hmcl: skipping assignment

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/service/EnvironmentService.java
```diff
@@ -434,6 +439,85 @@ public NamespaceServiceClusterMapping addOrUpdateServiceClusterMapping(
...
+    }
+
+    public Collection<ServiceBundle> listServiceBundles(List<QueryParam> params) {
+        return dao.find(SERVICE_BUNDLE_NAMESPACE, params);
+    }
+
+    public ServiceBundle getServiceBundleByName(String serviceName) {
+        Collection<ServiceBundle> serviceBundles = listServiceBundles(Lists.newArrayList(new QueryParam("name", serviceName)));
+        if (serviceBundles.size() > 1) {
+            LOG.warn("Multiple service bundles have same name: {} returning first match.", serviceName);
```
hmcl: with the same name

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/service/EnvironmentService.java
```diff
@@ -434,6 +439,85 @@ public NamespaceServiceClusterMapping addOrUpdateServiceClusterMapping(
...
+            return serviceBundles.iterator().next();
+        } else if (serviceBundles.size() == 1) {
+            return serviceBundles.iterator().next();
+        }
+        return null;
```
hmcl: shouldn't it throw exception?

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/HBaseServiceRegisterer.java
```diff
@@ -0,0 +1,73 @@
...
+import com.hortonworks.streamline.streams.catalog.ServiceConfiguration;
+import org.apache.commons.lang3.StringUtils;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+import static com.hortonworks.streamline.streams.cluster.discovery.ambari.ConfigFilePattern.HBASE_SITE;
+
+public class HBaseServiceRegisterer extends AbstractServiceRegisterer {
```
hmcl: Registrar

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/HDFSServiceRegisterer.java
```diff
@@ -0,0 +1,75 @@
...
+import org.apache.commons.lang3.StringUtils;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+import static com.hortonworks.streamline.streams.cluster.discovery.ambari.ConfigFilePattern.CORE_SITE;
+import static com.hortonworks.streamline.streams.cluster.discovery.ambari.ConfigFilePattern.HDFS_SITE;
+
+public class HDFSServiceRegisterer extends AbstractServiceRegisterer {
```
hmcl: Registrar

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/HiveServiceRegisterer.java
```diff
@@ -0,0 +1,75 @@
...
+import com.hortonworks.streamline.streams.catalog.ServiceConfiguration;
+import org.apache.commons.lang3.StringUtils;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+import static com.hortonworks.streamline.streams.cluster.discovery.ambari.ConfigFilePattern.HIVE_SITE;
+
+public class HiveServiceRegisterer extends AbstractServiceRegisterer {
```
hmcl: Registrar

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/KafkaServiceRegisterer.java
```diff
@@ -0,0 +1,103 @@
...
+import org.apache.commons.lang3.StringUtils;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+import static java.util.stream.Collectors.toList;
+
+public class KafkaServiceRegisterer extends AbstractServiceRegisterer {
```
hmcl: Registrar

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/HiveServiceRegisterer.java
```diff
@@ -0,0 +1,75 @@
...
+            }
+            return false;
+        }).count();
+
+        return validConfigFileCount == 1;
+    }
+
+    @Override
+    protected boolean validateServiceConfiguationsViaFlattenMap(Map<String, String> configMap) {
+        // FIXME: which configuration KVs are mandatory?
```
hmcl: remove this

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/HiveServiceRegisterer.java
```diff
@@ -0,0 +1,75 @@
...
+        // no need to check components
+        return true;
+    }
+
+    @Override
+    protected boolean validateServiceConfigurations(List<ServiceConfiguration> serviceConfigurations) {
+        // requirements
+        // 1. hive-site.xml should be provided
+
+        // FIXME: hive-metastore.xml in Ambari should be included to the proper place (maybe hive-site.xml?)
```
hmcl: remove this

streams/cluster/src/test/java/com/hortonworks/streamline/streams/cluster/register/impl/AbstractServiceRegistererTest.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    protected void resetEnvironmentService() {
+        dao = new InMemoryStorageManager();
+        environmentService = new EnvironmentService(dao);
+    }
+
+    protected T initializeServiceRegisterer() {
+        try {
+            T registerer = testClazz.newInstance();
```
hmcl: registrar

streams/cluster/src/test/java/com/hortonworks/streamline/streams/cluster/register/impl/HBaseServiceRegistererTest.java
```diff
@@ -0,0 +1,86 @@
...
+        super(HBaseServiceRegisterer.class);
+    }
+
+    @Before
+    public void setUp() {
+        resetEnvironmentService();
+    }
+
+    @Test
+    public void testRegister_happyCase() throws Exception {
```
hmcl: let's get a meaningful name for this. happyCase doesn't say if it is a validConfig, or what it is.

streams/cluster/src/test/java/com/hortonworks/streamline/streams/cluster/register/impl/HDFSServiceRegistererTest.java
```diff
@@ -0,0 +1,106 @@
...
+        super(HDFSServiceRegisterer.class);
+    }
+
+    @Before
+    public void setUp() {
+        resetEnvironmentService();
+    }
+
+    @Test
+    public void testRegister_happyCase() throws Exception {
```
hmcl: same comment as above

streams/cluster/src/test/java/com/hortonworks/streamline/streams/cluster/register/impl/KafkaServiceRegistererTest.java
```diff
@@ -0,0 +1,106 @@
...
+        super(KafkaServiceRegisterer.class);
+    }
+
+    @Before
+    public void setUp() {
+        resetEnvironmentService();
+    }
+
+    @Test
+    public void testRegister_happyCase() throws Exception {
```
hmcl: same comment as above

streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/exception/ServiceComponentNotFoundException.java
```diff
@@ -32,7 +28,7 @@ public ServiceComponentNotFoundException(Throwable cause) {
         super(cause);
     }
 
-    public ServiceComponentNotFoundException(Long clusterId, ServiceConfigurations service, ComponentPropertyPattern component) {
-        this(String.format("Component [%s] not found for service [%s] in cluster with id [%d]", component.name(), service.name(), clusterId));
+    public ServiceComponentNotFoundException(Long clusterId, String serviceName, String componentName) {
```
HeartSaVioR: I think I answered the comment earlier too. It's due to avoid cross reference between modules.

bootstrap/services/hbase-bundle.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "HBASE",
+  "registerClass": "com.hortonworks.streamline.streams.cluster.register.impl.HBaseServiceRegisterer",
```
HeartSaVioR: Yes right. Thanks for finding.

bootstrap/services/hbase-bundle.json
```diff
@@ -0,0 +1,15 @@
...
+  "serviceUISpecification": {
+    "fields": [
+      {
+        "uiName": "hbase-site.xml",
+        "fieldName": "hbase-site.xml",
+        "isOptional": false,
+        "tooltip": "Upload hbase-site.xml file which HBase cluster is using",
```
HeartSaVioR: Will address

bootstrap/services/hdfs-bundle.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "HDFS",
+  "registerClass": "com.hortonworks.streamline.streams.cluster.register.impl.HDFSServiceRegisterer",
```
HeartSaVioR: Will address

bootstrap/services/hdfs-bundle.json
```diff
@@ -0,0 +1,22 @@
...
+  "serviceUISpecification": {
+    "fields": [
+      {
+        "uiName": "core-site.xml",
+        "fieldName": "core-site.xml",
+        "isOptional": false,
+        "tooltip": "Upload core-site.xml file which Hadoop cluster is using",
```
HeartSaVioR: Will address

bootstrap/services/hdfs-bundle.json
```diff
@@ -0,0 +1,22 @@
...
+        "type": "file"
+      },
+      {
+        "uiName": "hdfs-site.xml",
+        "fieldName": "hdfs-site.xml",
+        "isOptional": false,
+        "tooltip": "Upload hdfs-site.xml file which Hadoop cluster is using",
```
HeartSaVioR: Will address

bootstrap/services/hive-bundle.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "HIVE",
+  "registerClass": "com.hortonworks.streamline.streams.cluster.register.impl.HiveServiceRegisterer",
```
HeartSaVioR: Will address

bootstrap/services/hive-bundle.json
```diff
@@ -0,0 +1,15 @@
...
+  "serviceUISpecification": {
+    "fields": [
+      {
+        "uiName": "hive-site.xml",
+        "fieldName": "hive-site.xml",
+        "isOptional": false,
+        "tooltip": "Upload hive-site.xml file which Hive cluster is using",
```
HeartSaVioR: Will address

bootstrap/services/kafka-bundle.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "KAFKA",
+  "registerClass": "com.hortonworks.streamline.streams.cluster.register.impl.KafkaServiceRegisterer",
```
HeartSaVioR: Will address

bootstrap/services/kafka-bundle.json
```diff
@@ -0,0 +1,22 @@
...
+        "fieldName": "brokersHostnames",
+        "isOptional": false,
+        "tooltip": "Hostnames of Kafka Brokers",
+        "type": "array.string"
+      },
+      {
+        "uiName": "server.properties",
+        "fieldName": "server.properties",
+        "isOptional": false,
+        "tooltip": "Upload server.properties file which Kafka cluster is using",
```
HeartSaVioR: Will address

bootstrap/services/storm-bundle.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "STORM",
+  "registerClass": "com.hortonworks.streamline.streams.cluster.register.impl.StormServiceRegisterer",
```
HeartSaVioR: Will address

bootstrap/services/zookeeper-bundle.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "ZOOKEEPER",
+  "registerClass": "com.hortonworks.streamline.streams.cluster.register.impl.ZookeeperServiceRegisterer",
```
HeartSaVioR: Will address

bootstrap/services/zookeeper-bundle.json
```diff
@@ -0,0 +1,22 @@
...
+        "fieldName": "zkServersHostnames",
+        "isOptional": false,
+        "tooltip": "Hostnames of Zookeeper servers",
+        "type": "array.string"
+      },
+      {
+        "uiName": "zoo.cfg",
+        "fieldName": "zoo.cfg",
+        "isOptional": false,
+        "tooltip": "Upload zoo.cfg file which Zookeeper cluster is using",
```
HeartSaVioR: Will address

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/service/EnvironmentService.java
```diff
@@ -434,6 +439,85 @@ public NamespaceServiceClusterMapping addOrUpdateServiceClusterMapping(
         return newMapping;
     }
 
+    public void injectProtocolAndPortToComponent(Map<String, String> configurations, Component component) {
```
HeartSaVioR: Copying my previous comment:
Not sure. Still not clear which is better: set vs add vs inject. Could you please describe why?

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/service/EnvironmentService.java
HeartSaVioR: Will address

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/service/EnvironmentService.java
HeartSaVioR: Will address

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/service/EnvironmentService.java
HeartSaVioR: Will address, but feel very minor

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/service/EnvironmentService.java
HeartSaVioR: It should be consistent between getXXX with DAO, and I didn't see any of them throwing exception.

streams/cluster/src/test/java/com/hortonworks/streamline/streams/cluster/register/impl/HBaseServiceRegistererTest.java
HeartSaVioR: Then just removing keyword `happyCase` would work for me, since it checks when everything with parameters is correct. What do you think?

streams/cluster/src/test/java/com/hortonworks/streamline/streams/cluster/register/impl/HBaseServiceRegistererTest.java
```diff
@@ -0,0 +1,86 @@
...
+        HBaseServiceRegisterer registerer = initializeServiceRegisterer();
+
+        try (InputStream is = getClass().getClassLoader().getResourceAsStream(HBASE_SITE_XML_FILE_PATH)) {
+            ManualServiceRegisterer.ConfigFileInfo hiveSiteXml = new ManualServiceRegisterer.ConfigFileInfo(HBASE_SITE_XML, is);
+            registerer.register(cluster, new Config(), Lists.newArrayList(hiveSiteXml));
+        }
+    }
+
+    @Test
+    public void testRegister_requiredPropertyNotPresented() throws Exception {
```
HeartSaVioR: Will address

streams/cluster/src/test/java/com/hortonworks/streamline/streams/cluster/register/impl/StormServiceRegistererTest.java
```diff
@@ -0,0 +1,129 @@
...
+        super(StormServiceRegisterer.class);
+    }
+
+    @Before
+    public void setUp() {
+        resetEnvironmentService();
+    }
+
+    @Test
+    public void testRegister_happyCase() throws Exception {
```
HeartSaVioR: Same here

streams/cluster/src/test/resources/register/happycase/core-site.xml
```diff
@@ -0,0 +1,134 @@
+  <configuration>
```
HeartSaVioR: OK agreed on this, and `correct_config` and `invalid_config` look good to me. Will address.

streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/cluster/ServiceBundle.java
```diff
@@ -0,0 +1,210 @@
...
+
+    @Override
+    @JsonIgnore
+    public StorableKey getStorableKey () {
+        return new StorableKey(getNameSpace(), getPrimaryKey());
+    }
+
+    @Override
+    public Map toMap () {
+        ObjectMapper mapper = new ObjectMapper();
```
HeartSaVioR: Agreed on static field. It shouldn't be an instance field since it would be on every entities.

streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/cluster/ServiceBundle.java
```diff
@@ -0,0 +1,210 @@
...
+        map.put(REGISTER_CLASS, registerClass);
+        return map;
+    }
+
+    @Override
+    public Storable fromMap (Map<String, Object> map) {
+        id = (Long) map.get(ID);
+        name = (String)  map.get(NAME);
+        timestamp = (Long) map.get(TIMESTAMP);
+        ObjectMapper mapper = new ObjectMapper();
```
HeartSaVioR: Same as above

streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/configuration/ConfigFileReader.java
```diff
@@ -0,0 +1,97 @@
...
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.HashMap;
+import java.util.Map;
+
+import static java.util.stream.Collectors.toMap;
+
+/**
+ * Reader that read configuration stream based on the file type.
```
HeartSaVioR: We might want to move the comment to the method, given that we are describing with detail. Will address.

streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/configuration/ConfigFileReader.java
```diff
@@ -0,0 +1,97 @@
...
+    private Map<String, String> readConfigFromYamlType(InputStream configFileStream) throws IOException {
+        ObjectReader reader = new YAMLMapper().reader();
+        Map<String, Object> confMap = reader.forType(new TypeReference<Map<String, Object>>() {})
+                .readValue(configFileStream);
+        return confMap.entrySet().stream()
+                .collect(toMap(Map.Entry::getKey, e -> convertValueAsString(e.getValue())));
+    }
+
+    private String convertValueAsString(Object value) {
+        if (value == null) {
```
HeartSaVioR: It is not a special case, but just one kind of the values.

streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/configuration/ConfigFileType.java
```diff
@@ -22,12 +22,12 @@
...
  */
 public enum ConfigFileType {
-  HADOOP_XML, PROPERTIES, YAML;
+  HADOOP_XML, PROPERTIES, YAML, ZOOKEEPER_CFG;
 
   public static ConfigFileType getFileTypeFromFileName(String fileName) {
     String fileExt = Files.getFileExtension(fileName);
     if (fileExt.isEmpty()) {
-      return null;
+      throw new IllegalArgumentException("Not supported config file extension: " + fileExt);
```
HeartSaVioR: 1. I just picked `null` as return value but not relevant with NPE. It's not happening because of referencing null.
2. I'm not in favor of throwing custom NPE since NPE doesn't have any meaning other than referencing null, and no one would want to catch NPE in normal.

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/discovery/ambari/AmbariServiceNodeDiscoverer.java
```diff
@@ -181,9 +183,29 @@ public void init(Map<String, String> conf) throws ConfigException {
...
+      return confName;
+    }).collect(toList());
+  }
+
+  private String getOriginConfigTypeName(String confItemType) {
+    ServiceConfigTypeAmbariToOriginalPattern pattern = ServiceConfigTypeAmbariToOriginalPattern.findByAmbariConfType(confItemType);
+    if (pattern != null) {
+      return pattern.originConfType();
+    }
+    return confItemType;
```
HeartSaVioR: I think it should not happen, but we could apply defensive programming: place precondition check here, and I'm in favor of doing this. Will fix.

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/discovery/ambari/AmbariServiceNodeDiscoverer.java
```diff
@@ -171,7 +171,9 @@ public void init(Map<String, String> conf) throws ConfigException {
                 confNameList, items);
 
         for (ServiceConfigurationItem confItem : confToItem.values()) {
-          configurations.put(confItem.getType(), getProperties(confItem));
+          // convert Ambari type to the actual Service's configuration file
+          String type = getOriginConfigTypeName(confItem.getType());
+          configurations.put(type, getProperties(confItem));
```
HeartSaVioR: Left comment below.

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/discovery/ambari/ServiceConfigTypeAmbariToOriginalPattern.java
```diff
@@ -0,0 +1,64 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
+package com.hortonworks.streamline.streams.cluster.discovery.ambari;
+
+/**
+ * This enum defines how Ambari configuration type is converted to original service configuration type.
```
HeartSaVioR: Yes right.

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/discovery/ambari/ServiceConfigTypeAmbariToOriginalPattern.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    public static ServiceConfigTypeAmbariToOriginalPattern findByAmbariConfType(String ambariConfType) {
+        for (ServiceConfigTypeAmbariToOriginalPattern value : values()) {
+            if (value.ambariConfType.equals(ambariConfType)) {
+                return value;
+            }
+        }
+
+        return null;
```
HeartSaVioR: It just means that matching entity doesn't exist. Changing this to `Optional` is OK to me. Will address, and also below `findByOriginalConfType`.

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/model/ServiceWithComponents.java
```diff
@@ -0,0 +1,52 @@
...
+import com.hortonworks.streamline.streams.catalog.Component;
+import com.hortonworks.streamline.streams.catalog.Service;
+import com.hortonworks.streamline.streams.catalog.ServiceConfiguration;
+
+import java.util.Collection;
+
+public class ServiceWithComponents {
+    private Service service;
+    private Collection<ServiceConfiguration> configurations;
+    private Collection<Component> components;
```
HeartSaVioR: ServiceConfiguration is a part of configurations for specific service, and Component is an one of components for specific service. They are not related to each other.

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/ManualServiceRegisterer.java
```diff
@@ -0,0 +1,50 @@
...
+import com.hortonworks.streamline.common.Config;
+import com.hortonworks.streamline.streams.catalog.Cluster;
+import com.hortonworks.streamline.streams.catalog.Service;
+import com.hortonworks.streamline.streams.cluster.service.EnvironmentService;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.List;
+
+public interface ManualServiceRegisterer {
```
HeartSaVioR: Will address

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/AbstractServiceRegisterer.java
```diff
@@ -0,0 +1,124 @@
...
+import com.hortonworks.streamline.streams.cluster.service.EnvironmentService;
+import org.apache.commons.io.FilenameUtils;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public abstract class AbstractServiceRegisterer implements ManualServiceRegisterer {
```
HeartSaVioR: Will address

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/AbstractServiceRegisterer.java
```diff
@@ -0,0 +1,124 @@
...
+
+    protected abstract String getServiceName();
+
+    protected abstract List<Component> createComponents(Config config, Map<String, String> flattenConfigMap);
+
+    protected abstract boolean validateComponents(List<Component> components);
+
+    protected abstract boolean validateServiceConfigurations(List<ServiceConfiguration> serviceConfigurations);
+
+    protected abstract boolean validateServiceConfiguationsViaFlattenMap(Map<String, String> configMap);
```
HeartSaVioR: Will address

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/AbstractServiceRegisterer.java
```diff
@@ -0,0 +1,124 @@
...
+
+    @Override
+    public void init(EnvironmentService environmentService) {
+        this.environmentService = environmentService;
+    }
+
+    @Override
+    public Service register(Cluster cluster, Config config, List<ConfigFileInfo> configFileInfos) throws IOException {
+        Service service = environmentService.initializeService(cluster, getServiceName());
```
HeartSaVioR: It creates new Service entity, which is also bound to specific Cluster. It should be created per each request.

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/AbstractServiceRegisterer.java
```diff
@@ -0,0 +1,124 @@
...
+
+        List<ServiceConfiguration> configurations = new ArrayList<>();
+        Map<String, String> flattenConfigMap = new HashMap<>();
+
+        ObjectMapper objectMapper = new ObjectMapper();
```
HeartSaVioR: Will address.

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/HBaseServiceRegisterer.java
HeartSaVioR: Will address.

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/HBaseServiceRegisterer.java
HeartSaVioR: Will remove this, and add once we find out more.

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/HDFSServiceRegisterer.java
HeartSaVioR: Will address.

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/HiveServiceRegisterer.java
HeartSaVioR: Will address.

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/HiveServiceRegisterer.java
HeartSaVioR: Will address.

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/HiveServiceRegisterer.java
HeartSaVioR: Will address.

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/KafkaServiceRegisterer.java
HeartSaVioR: Will address.

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/KafkaServiceRegisterer.java
HeartSaVioR: Will address.

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/StormServiceRegisterer.java
HeartSaVioR: Will address.

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/StormServiceRegisterer.java
HeartSaVioR: Will address.

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/StormServiceRegisterer.java
HeartSaVioR: Will address.

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/ZookeeperServiceRegisterer.java
HeartSaVioR: Will address.

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/AbstractServiceRegisterer.java
```diff
@@ -0,0 +1,124 @@
...
+    protected abstract List<Component> createComponents(Config config, Map<String, String> flattenConfigMap);
+
+    protected abstract boolean validateComponents(List<Component> components);
+
+    protected abstract boolean validateServiceConfigurations(List<ServiceConfiguration> serviceConfigurations);
+
+    protected abstract boolean validateServiceConfiguationsViaFlattenMap(Map<String, String> configMap);
+
+    @Override
+    public void init(EnvironmentService environmentService) {
```
HeartSaVioR: We're initializing actual class dynamically so having init() method is necessary for passing EnvironmentService for all of classes.

streams/cluster/src/test/java/com/hortonworks/streamline/streams/cluster/register/impl/AbstractServiceRegistererTest.java
HeartSaVioR: Will address.

streams/cluster/src/test/java/com/hortonworks/streamline/streams/cluster/register/impl/HDFSServiceRegistererTest.java
HeartSaVioR: Will address.

streams/cluster/src/test/java/com/hortonworks/streamline/streams/cluster/register/impl/KafkaServiceRegistererTest.java
HeartSaVioR: Will address.

HeartSaVioR:
@hmcl Thanks for thoughtful review. Addressed comments or left comments for each.

streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/exception/ServiceConfigurationNotFoundException.java
```diff
@@ -31,7 +28,7 @@ public ServiceConfigurationNotFoundException(Throwable cause) {
         super(cause);
     }
 
-    public ServiceConfigurationNotFoundException(Long clusterId, ServiceConfigurations service, String configurationName) {
-        this(String.format("Configuration [%s] not found for service [%s] in cluster with id [%d]", configurationName, service.name(), clusterId));
+    public ServiceConfigurationNotFoundException(Long clusterId, String serviceName, String configurationName) {
```
HeartSaVioR: Same as above.

bootstrap/services/storm-bundle.json
priyank5485: @HeartSaVioR Want to clarify your last comment. You said it will not be used for overriding storm config for topology submission. Should not we be using this for submitting topology, now that we have this?

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/ManualServiceRegistrar.java
```diff
@@ -0,0 +1,50 @@
...
+import com.hortonworks.streamline.streams.catalog.Service;
+import com.hortonworks.streamline.streams.cluster.service.EnvironmentService;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.List;
+
+public interface ManualServiceRegistrar {
+
+    void init(EnvironmentService environmentService);
```
priyank5485: Should we add a line or two documenting the interface methods and may be EnvironmentService as well?

streams/cluster/src/test/java/com/hortonworks/streamline/streams/cluster/register/impl/HBaseServiceRegistrarTest.java
```diff
@@ -0,0 +1,86 @@
...
+    public void testRegister() throws Exception {
+        Cluster cluster = getTestCluster(1L);
+
+        HBaseServiceRegistrar registrar = initializeServiceRegistrar();
+
+        try (InputStream is = getClass().getClassLoader().getResourceAsStream(HBASE_SITE_XML_FILE_PATH)) {
+            ManualServiceRegistrar.ConfigFileInfo hiveSiteXml = new ManualServiceRegistrar.ConfigFileInfo(HBASE_SITE_XML, is);
+            registrar.register(cluster, new Config(), Lists.newArrayList(hiveSiteXml));
+        }
+    }
```
priyank5485: Does it make sense to add a assertEquals after the register call? Probably query EnvironmentService and assert it to be equal to the expected object we create in the test case if i understand it correctly?

priyank5485:
@HeartSaVioR This looks great. Left some comments/questions. +1 once they are addressed. Sorry for the late review.

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/ManualServiceRegistrar.java
HeartSaVioR: Yeah that would be better. Thanks for finding missed spots.

streams/cluster/src/test/java/com/hortonworks/streamline/streams/cluster/register/impl/HBaseServiceRegistrarTest.java
HeartSaVioR: I'd rather not depending on the content of the test resource, but simple verification can be done without depending on that too much. Will address.

bootstrap/services/storm-bundle.json
HeartSaVioR: @priyank5485 
Port informations are extracted from the config file. 
Maybe we can just let user input ports too, but given that we should support secured Storm cluster as well, I'd rather leave this until we define all mandatory and optional fields.

HeartSaVioR:
@priyank5485 Thanks for reviewing. Addressed.

priyank5485:
@arunmahadevan @hmcl Let know if there is something you still want to be addressed. If not I will merge this in

hmcl:
@priyank5485 taking a look.

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/service/EnvironmentService.java
hmcl: The problem is that most DAO methods currently return null and in my opinion they are all incorrect. They should all throw an unchecked exception. For example if a method like getComponentByName, getServiceByName, getServiceConfigurationByName returns null, it means something unexpected occurred. You were looking for something that wasn't there. The client code should be alerted for that, not simply return a null value. A null value is simply a default value. Why is null different from -1, or a string "Not Found" ? For example, the Java Persistence API, when looking up an entity in an underlying data storage (database) and that entity is not found, it throws [EntityNotFoundException](https://docs.oracle.com/javaee/7/api/javax/persistence/EntityNotFoundException.html)

streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/configuration/ConfigFileReader.java
hmcl: I am not arguing one way or another here as I don't know the whole context.

streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/configuration/ConfigFileType.java
hmcl: This is a good option. You can also consider IllegalStateException, or create something like InvalidConfigException, if we don't already have something similar to it in the Streamline codebase.

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/model/ServiceWithComponents.java
hmcl: So this class is just a wrapper for all `ServiceConfiguration` and `Component` of a `Service` ?

streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/configuration/ConfigFileReader.java
```diff
@@ -32,11 +32,15 @@
 import static java.util.stream.Collectors.toMap;
 
 /**
- * Reader that read configuration stream based on the file type.
+ * Configuration reader that based on the file type.
```
hmcl: nit: probably you want to delete "that"

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/AbstractServiceRegistrar.java
```diff
@@ -83,7 +85,7 @@ public Service register(Cluster cluster, Config config, List<ConfigFileInfo> con
             throw new IllegalArgumentException("Validation failed for service configurations.");
         }
 
-        if (!validateServiceConfiguationsViaFlattenMap(flattenConfigMap)) {
+        if (!validateServiceConfiguationsAsFlattenedMap(flattenConfigMap)) {
```
hmcl: nit: perhaps a better name would be: `isValidServiceConfiguationsAsFlattenedMap(...)`

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/HBaseServiceRegistrar.java
```diff
@@ -66,8 +66,7 @@ protected boolean validateServiceConfigurations(List<ServiceConfiguration> servi
     }
 
     @Override
-    protected boolean validateServiceConfiguationsViaFlattenMap(Map<String, String> configMap) {
-        // FIXME: which configuration KVs are mandatory?
+    protected boolean validateServiceConfiguationsAsFlattenedMap(Map<String, String> configMap) {
```
hmcl: nit: perhaps a better name would be: `isValidServiceConfiguationsAsFlattenedMap(...)`

streams/cluster/src/test/java/com/hortonworks/streamline/streams/cluster/register/impl/HBaseServiceRegistrarTest.java
```diff
@@ -53,6 +56,11 @@ public void testRegister() throws Exception {
             ManualServiceRegistrar.ConfigFileInfo hiveSiteXml = new ManualServiceRegistrar.ConfigFileInfo(HBASE_SITE_XML, is);
             registrar.register(cluster, new Config(), Lists.newArrayList(hiveSiteXml));
         }
+
+        Service hbaseService = environmentService.getServiceByName(cluster.getId(), HBaseServiceRegistrar.SERVICE_NAME_HBASE);
+        assertNotNull(hbaseService);
```
hmcl: it's beyond the scope of this patch, but I would really vote that we get rid of all of these nulls. Do you think we should file a JIRA for that ?

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/model/ServiceWithComponents.java
HeartSaVioR: Yes right.

streams/cluster/src/test/java/com/hortonworks/streamline/streams/cluster/register/impl/HBaseServiceRegistrarTest.java
HeartSaVioR: Yes I think you need to file an issue, and before that, need to make consensus among team. Personally I'm OK to make it Optional but not sure to throw specific exception.

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/service/EnvironmentService.java
HeartSaVioR: While I agree that `null` doesn't represent such meaning, but `Optional` provides that, so throwing exception is not the only option for us. Please start discussion to reach a consensus if you think this is a problem.

HeartSaVioR:
@hmcl 
Sorry I lost track for this. Left comments for your latest review comments.

PR #513, opened 2017-02-28
STREAMLINE-664 Handle tick tuples in all storm components of a streamline topology

HeartSaVioR:
* adds TupleUtils.isTick for each Bolt implementations

I've seen some Storm Bolt implementations also don't handle tick tuples. I'll make a patch for Storm side too.
If we want to wrap the Bolt in Streamline and fix in Streamline, please let me know.


priyank5485:
@HeartSaVioR Is there a reason for not adding it to RulesBolt? I checked other bolts and i think thats the only one we are missing. +1 after addressing that comment.

priyank5485:
@HeartSaVioR After reviewing storm PR, I think we should handle it in streamline the same way as well.

HeartSaVioR:
@priyank5485 
No I just missed it. Addressed. I'm not clear that we can utilize the class from Storm @satishd proposed. Do we want to have duplicated class from Streamline side, given that we can not get it fairly soon since I'm against to include any non-blocker things to Storm 1.1.0?

satishd:
@HeartSaVioR Agree on having a duplicate class in streamline till we get the respective storm changes into streamline.

HeartSaVioR:
@satishd @priyank5485 OK updated.

satishd:
+1 LGTM

PR #515, opened 2017-03-03
STREAMLINE-723 JdbcStorageManager#add(Storable) should call insert in…

satishd:
JdbcStorageManager#add(Storable) should call insert instead of insertOrUpdate and it should throw AlreadyExistsException when it encounters any constraint validation exception.

### DIFF
--- a/storage/core/src/main/java/com/hortonworks/streamline/storage/impl/jdbc/JdbcStorageManager.java
+++ b/storage/core/src/main/java/com/hortonworks/streamline/storage/impl/jdbc/JdbcStorageManager.java
@@ -36,6 +36,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.sql.Connection;
+import java.sql.SQLException;
+import java.sql.SQLIntegrityConstraintViolationException;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
@@ -61,14 +63,17 @@ public class JdbcStorageManager implements StorageManager {
     @Override
     public void add(Storable storable) throws AlreadyExistsException {
         log.debug("Adding storable [{}]", storable);
-        final Storable existing = get(storable.getStorableKey());
 
-        if(existing == null) {
-            addOrUpdate(storable);
-        } else if (!existing.equals(storable)) {
-            throw new AlreadyExistsException("Another instance with same id = " + storable.getPrimaryKey()
-                    + " exists with different value in namespace " + storable.getNameSpace()
-                    + ". Consider using addOrUpdate method if you always want to overwrite.");
+        try {
+            queryExecutor.insert(storable);
+        } catch (StorageException ex) {
+            if (ex.getCause() instanceof SQLIntegrityConstraintViolationException) {
+                throw new AlreadyExistsException("Another instance with same id = " + storable.getPrimaryKey()
+                                                         + " exists with different value in namespace " + storable.getNameSpace()
+                                                         + ". Consider using addOrUpdate method if you always want to overwrite.");
+            }
+
+            throw ex;
         }
     }
 
--- a/storage/core/src/main/java/com/hortonworks/streamline/storage/impl/jdbc/JdbcStorageManager.java
+++ b/storage/core/src/main/java/com/hortonworks/streamline/storage/impl/jdbc/JdbcStorageManager.java
@@ -36,8 +36,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.sql.Connection;
-import java.sql.SQLException;
-import java.sql.SQLIntegrityConstraintViolationException;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


satishd:
@harshach pinging for reviewing this PR.

storage/core/src/main/java/com/hortonworks/streamline/storage/impl/jdbc/JdbcStorageManager.java
```diff
@@ -61,14 +63,17 @@ public JdbcStorageManager(QueryExecutor queryExecutor) {
...
-        } else if (!existing.equals(storable)) {
-            throw new AlreadyExistsException("Another instance with same id = " + storable.getPrimaryKey()
-                    + " exists with different value in namespace " + storable.getNameSpace()
-                    + ". Consider using addOrUpdate method if you always want to overwrite.");
+
+        try {
+            queryExecutor.insert(storable);
+        } catch (StorageException ex) {
+            if (ex.getCause() instanceof SQLIntegrityConstraintViolationException) {
+                throw new AlreadyExistsException("Another instance with same id = " + storable.getPrimaryKey()
```
HeartSaVioR: http://docs.oracle.com/javase/7/docs/api/java/sql/SQLIntegrityConstraintViolationException.html

This is only true when insert is failing due to the violation for PK uniqueness, neither foreign key nor unique key.

So we still need to query for existence while we are defining foreign key.

satishd: Javadoc says that 
> The subclass of SQLException thrown when the SQLState class value is '23', or under vendor-specified conditions. This indicates that an integrity constraint (foreign key, primary key or unique key) has been violated.
Please consult your driver vendor documentation for the vendor-specified conditions for which this Exception may be thrown.

So, this includes foreign key, primary key or unique key if the driver supports to throw this Exception.

This is kind of best effort to throw `AlreadyExistsException` by catching `SQLIntegrityConstraintViolationException`. We should not check for `get` and then execute `addOrUpdate` as `addOrUpdate` updates on duplicate keys and we do not want to update on duplicate keys for `add` method. It also avoids two calls to DB for inserting data. If the vendor throws some other Exception then that error is thrown back to the caller.

HeartSaVioR: I'm worrying about throwing wrong message: AlreadyExistsException for foreign key violation. It makes user harder to find out the root reason, since we would just respond with HTTP STATUS 409 conflict, which origin issue is that record for FK doesn't exist.

satishd: I would prefer to go with right behavior. We can add more error logging about the error details which may be helpful in finding the root cause.

HeartSaVioR: If we are OK to tolerate false positive like this, how about selecting again when SQLIntegrityConstraintViolationException is thrown?

If "select query" succeeds, we could treat previous exception as unexpected exception since the exception would mean foreign key or unique key violation.
If "select query" fails, we could treat it as AlreadyExistsException. 

There's false positive here: "insert query" fails with PK uniqueness violation, and entity is deleted from other before executing "select query". It's a tiny chance so don't hurt much, and even we can retry insertion at that time.

satishd: We can add that even though it is not foolproof as it is a best effort to return right Exception message with details by not sacrificing the right behavior of the method.

satishd: @HeartSaVioR That logic looks convoluted. Are you suggesting the below?
```
try {
            queryExecutor.insert(storable);
        } catch (StorageException ex) {
            Throwable cause = ex.getCause();
            if (cause instanceof SQLIntegrityConstraintViolationException) {
                // below check is the suggestion!!
                if(get(storable.getStorableKey()) == null) {
                    throw new AlreadyExistsException("Another instance with same id = " + storable.getPrimaryKey()
                                                             + " exists with different value in namespace " + storable.getNameSpace()
                                                             + ". Consider using addOrUpdate method if you always want to overwrite.");
                }
            }

            throw ex;
        }
```
>If "select query" fails, we could treat it as AlreadyExistsException.
This logic looks confusing as `AlreadyExistsException` is thrown when there is no entry for the given Storable.

arunmahadevan: AlreadyExistsException is a runtime exception and not sure why `add` (and other methods in StorageManager) declares unchecked exception in throws. Nowhere in the code we catch/handle AlreadyExistsException. IMO, we can just log the StorageException and just propagate it up and let the client handle it by looking at the error code if required.

satishd: That is another topic, there was a todo added in [StorageException](https://github.com/hortonworks/streamline/blob/master/storage/core/src/main/java/com/hortonworks/streamline/storage/exception/StorageException.java#L18). It seems it was added in very early days(mid 2015) . It seems storage and cache APIs require some cleanup.

I guess we are trying to give better error message for the client without showing JDBC provider's error message as that looks to be confusing. This is not a critical issue to be addressed but trying to give possible comprehensible error message to user.

arunmahadevan: Then better throw exception with right message based on whether its primary, foreign or unique key violation.

HeartSaVioR:
+1 
That seems to be what we should do.

PR #516, opened 2017-03-03
STREAMLINE-596 Use Schema from registry/common module

satishd:


satishd:
@harshach Pinging for review.

HeartSaVioR:
Just 2 cents: we might want to think about streamline and registry as separate projects.
Schema class is used for more than communicating with registry, like defining schema for storable, which is not relevant to registry. Looks like unneeded coupling.

satishd:
> Just 2 cents: we might want to think about streamline and registry as separate projects.
Schema class is used for more than communicating with registry, like defining schema for storable, which is not relevant to registry. Looks like unneeded coupling.

@HeartSaVioR Actually, it is the other way. Storage is needed by registry also and we want to maintain storage module in registry repo and that module should be used in streamline. It is same with dependent common classes like Schema. There is a task to remove storage module from streamline and use storage module from registry repo.

@harshach You can add more if any.

HeartSaVioR:
Yes there're some more duplicated things between projects. May be better to have common module(s), and ideally common module(s) might be better to move to the another project. 
As I commented, it's just 2 cents, and we could move them out of later whenever necessary.

harshach:
+1

PR #514, opened 2017-03-01
STREAMLINE-643: Postgresql support

arunmahadevan:
This is copy of https://github.com/hortonworks/streamline/pull/503 plus,

1. Renamed tables
2. Added streamline.postgre.yaml

PR #519, opened 2017-03-08
STREAMLINE-727 bootstrap-stroage script should allow create only and …

HeartSaVioR:
…keep deletion as an option

* bootstrap-storage.sh receives one optional argument
  * default value is "create"
  * "create": create tables
  * "drop": drop tables
  * "drop-create": drop then create tables
  * "check-connection": only check connection
* create new class StreamlineTableInitializer to keep SQLScriptRunner usable for general case
  * new class will handle only the case what bootstrap-storage provides
  * refactor SQLScriptRunner to let new class uses features

Regarding concurrent execution, since we're using `IF NOT EXISTS` for creating table and `IF EXISTS` for dropping table so it should be safe.

### DIFF
--- a/bootstrap/bootstrap-storage.sh
+++ b/bootstrap/bootstrap-storage.sh
@@ -32,6 +32,7 @@ done
 BOOTSTRAP_DIR=`dirname ${PRG}`
 CONFIG_FILE_PATH=${BOOTSTRAP_DIR}/../conf/streamline.yaml
 MYSQL_JAR_URL_PATH=https://dev.mysql.com/get/Downloads/Connector-J/mysql-connector-java-5.1.40.zip
+SCRIPT_ROOT_DIR="${BOOTSTRAP_DIR}/sql"
 
 # Which java to use
 if [ -z "${JAVA_HOME}" ]; then
@@ -40,7 +41,7 @@ else
   JAVA="${JAVA_HOME}/bin/java"
 fi
 
-SCRIPT_RUNNER_MAIN_CLASS=com.hortonworks.streamline.storage.tool.SQLScriptRunner
+TABLE_INITIALIZER_MAIN_CLASS=com.hortonworks.streamline.storage.tool.TablesInitializer
 for file in "${BOOTSTRAP_DIR}"/lib/*.jar;
 do
     CLASSPATH="$CLASSPATH":"$file"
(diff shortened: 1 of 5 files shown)


storage/tool/src/main/java/com/hortonworks/streamline/storage/tool/SQLScriptRunner.java
```diff
@@ -39,44 +42,99 @@
...
+                    System.out.println(String.format("######## SQL Query:  %s ", query));
+                    connection.createStatement().execute(query);
+                    System.out.println("######## Query executed");
+                }
             }
         }
     }
-    
+
+    public void downloadMySQLJarIfNeeded(String mysqlJarUrl) throws Exception {
```
satishd: I guess this class should not be specific to MySql. It is more specific to any JDBC supported storage. PR #514 already supports Postgres. It is better to clean this up.

bootstrap/bootstrap-storage.sh
```diff
@@ -40,34 +41,60 @@ else
...
 
-function streamlineBootstrapStorage {
-    SCRIPT_DIR="${BOOTSTRAP_DIR}/sql/<dbtype>"
-    FILE_OPT="-f ${SCRIPT_DIR}/drop_tables.sql -f ${SCRIPT_DIR}/create_tables.sql"
+function dropStreamlineTables {
+    ${JAVA} -Dstreamline.bootstrap.dir=$BOOTSTRAP_DIR  -cp ${CLASSPATH} ${STREAMLINE_TABLE_INITIALIZER_MAIN_CLASS} -m ${MYSQL_JAR_URL_PATH} -c ${CONFIG_FILE_PATH} --drop-script ${SCRIPT_DIR}/drop_tables.sql --drop
+}
+
+function createStreamlineTables {
+    ${JAVA} -Dstreamline.bootstrap.dir=$BOOTSTRAP_DIR  -cp ${CLASSPATH} ${STREAMLINE_TABLE_INITIALIZER_MAIN_CLASS} -m ${MYSQL_JAR_URL_PATH} -c ${CONFIG_FILE_PATH} --create-script ${SCRIPT_DIR}/create_tables.sql --create
```
satishd: This script should not be specific to MySql. We are going to support Postgres with #514 
You may want to make appropriate changes.

bootstrap/bootstrap-storage.sh
```diff
@@ -40,34 +41,60 @@ else
...
+}
 
-    echo "Script files option: $FILE_OPT"
-    exec ${JAVA} -Dstreamline.bootstrap.dir=$BOOTSTRAP_DIR  -cp ${CLASSPATH} ${SCRIPT_RUNNER_MAIN_CLASS} -m ${MYSQL_JAR_URL_PATH} -c ${CONFIG_FILE_PATH} ${FILE_OPT}
+function checkStorageConnection {
+    ${JAVA} -Dstreamline.bootstrap.dir=$BOOTSTRAP_DIR  -cp ${CLASSPATH} ${STREAMLINE_TABLE_INITIALIZER_MAIN_CLASS} -m ${MYSQL_JAR_URL_PATH} -c ${CONFIG_FILE_PATH} --check-connection
```
satishd: Same as earlier comment.

storage/tool/src/main/java/com/hortonworks/streamline/storage/tool/SQLScriptRunner.java
HeartSaVioR: Makes sense. I'll move out this method to new utility class, and let callers call that instead.

bootstrap/bootstrap-storage.sh
HeartSaVioR: It is not that easy to make the change from here. The script doesn't know which storage type the configuration sets up. Previously the script knows about storage type (via executing another Java class), but we abstracted it to the Java class to reduce two operations into one.

It shouldn't hurt to PostgreSQL storage setup, too.

bootstrap/bootstrap-storage.sh
HeartSaVioR: Same here.

storage/tool/src/main/java/com/hortonworks/streamline/storage/tool/SQLScriptRunner.java
HeartSaVioR: Btw, like below it shouldn't hurt to PostgreSQL storage setup.

HeartSaVioR:
@satishd 
I might be misunderstanding your comment. Where do you think mysql specific logic should be placed?
Please let me know that you're thinking about moving out `mysql lib download` to the bootstrap script.

satishd:
@HeartSaVioR Respective storage type scripts like mysql and postgres can have their own scripts (`bootstrap-storage-<mysql/postgres>.sh`) which can share common functionality from other script like `bootstrap-storage-common.sh`.

bootstrap/bootstrap-storage.sh
harshach: @HeartSaVioR given that we have util class to read the streamline configs. We should use that figure out the db.type from config.

harshach: Also we are expecting sql files to be under bootstrap/sql. Instead of passing all these options I would recommend we follow a syntax.
i.e store all the sql files in this manner
bootstrap/sql/db_type/create_tables.sql , bootstrap/sql/db_type/drop_tables.sql.
we can change the db_type mysql or postegres, phoenix etc..
Instead of having to pass drop_tables.sql or create_tables.sql . You only pass the parent directory under which sql scripts are stored. In this case, it would be "bootstrap/sql".
Your Java Class then can read the streamline  config to determine the right db.type and use that to invoke right db_type sql script and also it can take an argument such as "--create" or "--drop" like you are doing already but no need to pass the specific script to do so.

harshach:
@satishd @HeartSaVioR we are waiting on this to unblock ambari changes and QE. I added my comments to make this generic. Let me know if you've any questions regarding this.

HeartSaVioR:
@satishd @harshach 
I feel we need to arrange this. Let's look at the history of change.

1. For the first time, this was implemented to two classes: config reader / sql runner. (IOT-441, #265)
Config reader reads config and figures out connection information, and sql runner runs the scripts for given dbtype.

2. After that we changed config reader to an utility class (hence not called by script), and utilize config reader class into sql runner. (STREAMLINE-612, #419) 

3. We needed downloading MySQL driver because of license issue, and added the feature: download MySQL driver to sql runner. (STREAMLINE-643, #449)
Before this change it was generic.

If we want to make the thing generic, let's just exclude mysql driver download to the outside of sql runner, and maybe, even outside of bootstrap-storage. Sql runner was generic before handling mysql driver download.
(Users might need to execute separate script for MySQL for the first time. If we think it hurts UX and don't want to do that, bootstrap-storage or sql runner should have it as exceptional case and can't be generic.)

What do you think?

harshach:
@HeartSaVioR mysql driver will be downloaded only when we configure the db.type to be "mysql". why we need to change that?

HeartSaVioR:
@harshach 
What @satishd commented about moving out specific implementation is all related to mysql download. (Please look through where @satishd left comments.)
So my understanding of review comments are allowing non-generic vs generic and move out mysql driver download. It doesn't represent that I want to move out the logic, just thinking about the way to make it db-type generic. Either (keep as it is or change it) is fine for me.

@satishd Please follow up the thread and let us know your thought.

storage/tool/src/main/java/com/hortonworks/streamline/storage/tool/SQLScriptRunner.java
harshach: @satishd My understanding is postgres library is ASLv2 complaint where as mariadb or mysql driver is not. Hence we only added for mysql one. For postgres we should ship the library in classpath and shouldn't require a download.

bootstrap/bootstrap-storage.sh
HeartSaVioR: @harshach Yes new class is used only for bootstrapping so we could put the sql file placement information to the class. Great point. I'll apply it.

HeartSaVioR: @harshach Addressed.

satishd:
@HeartSaVioR @harshach I am fine with the change as if it is one off the case.

bootstrap/bootstrap-storage.sh
harshach: @satishd @HeartSaVioR are we in agreement with the above approach.

harshach: @HeartSaVioR thanks

satishd:
+1

PR #521, opened 2017-03-10
STREAMLINE-726: Authorization for streamline REST apis

arunmahadevan:
Added authorization for the various streamline resources.

Note: This includes changes from https://github.com/hortonworks/streamline/pull/517


streams/authorizer/src/main/java/com/hortonworks/streamline/streams/security/Roles.java
```diff
@@ -0,0 +1,31 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
+package com.hortonworks.streamline.streams.security;
+
+public class Roles {
+    public static final String ROLE_ADMIN = "ROLE_ADMIN";
+    public static final String ROLE_SECURITY_ADMIN = "ROLE_SECURITY_ADMIN";
```
harshach: we should reduce the number of roles here. We can keep the "Admin", "Developer", "OP" and treat notifications and files as resources

harshach: I guess these are like umbrella permissions set right. i.e a user can have one or more of this to get broader access

harshach:
@arunmahadevan This looks great.  +1.

streams/authorizer/src/main/java/com/hortonworks/streamline/streams/security/Roles.java
arunmahadevan: @harshach these are fine grained roles checked for respective operations. E.g. a user should have CLUSTER_ADMIN role to create a service pool, NOTIFIER_ADMIN for uploading notifier jars and so on. We can group multiple roles and create higher level application roles like DEV, OPS etc and assign them to users. We could do this at bootstrap time as well.

PR #522, opened 2017-03-14
STREAMLINE:472 Scheme class customization for kafka sources

satishd:


streams/runners/storm/layout/src/main/java/com/hortonworks/streamline/streams/layout/storm/KafkaSpoutFluxComponent.java
```diff
@@ -113,15 +116,49 @@ private String addSpoutConfigComponent () {
...
+        String mapClassId = "map-" + UUID_FOR_COMPONENTS;
+        List<Pair<String, Object>> entries =
+                Lists.newArrayList(Pair.of(KafkaSourceScheme.TOPIC_KEY,
+                                           conf.get(TopologyLayoutConstants.JSON_KEY_TOPIC)),
+                                   Pair.of(KafkaSourceScheme.SCHEMA_REGISTRY_URL_KEY,
+                                           conf.get(TopologyLayoutConstants.SCHEMA_REGISTRY_URL)),
+                                   Pair.of(KafkaSourceScheme.DATASOURCE_ID_KEY,
+                                           kafkaSource != null ? kafkaSource.getId() : ""));
+        List<Map<String, Object>> configMethods = new ArrayList<>();
+        for (Map.Entry<String, Object> entry : entries) {
```
satishd: Initially I had that code and it did not look good to me. Avoided multiple statements to set configMethods by having a list with 3 elements.

streams/runners/storm/layout/src/main/java/com/hortonworks/streamline/streams/layout/storm/KafkaSourceScheme.java
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Scheme to be implemented for Kafka sources to deserialize bytebuffer record into tuples.
+ */
+public interface KafkaSourceScheme extends MultiScheme {
+
+    /**
+     * Key for topic property.
+     */
+    String TOPIC_KEY = "topic";
```
satishd: >The constants can be moved to some other class like KafkaSpoutFluxComponent.

These constants are more specific to `KafkaSourceScheme` and I would like them to be declared here.

>We can accept MultiScheme implementations, instead of forcing them to implement KafkaSourceScheme always so that users can pass existing multi schemes. If the user has not provided any scheme, by default we can instantiate AvroKafkaSpoutScheme while building the flux component. If the users pass an instance of KafkaSourceScheme we invoke the init with the config.

Kafka source expects that user should register schema for a given topic and it does not allow configuring the source till the schema is available for the given topic. So, respective `MultiScheme` should be used for deserializing the messages. We can allow other MultiSchemes but those schemes may handle opaque data like string/binary without any structure defined by schema. I guess it is better to restrict users but there may be cases where producers may not be really using any schema registry and sends opaque data to respective topics. In that case it is good to allow user to configure respective MultiSchemes. I am fine in supporting this kind of usecase.

@harshach What is your opinion on this?

arunmahadevan: >These constants are more specific to KafkaSourceScheme and I would like them to be declared here.

How do you enforce all the implementors of this interface to use these constants? Then its better to have these as methods of the interface like `getTopic`, `getSchemaRegistryUrl` etc. If this is specific to AvroKafkaSpoutScheme then better move it there.

satishd: We are not trying to enforce all the implementors to use these constants. They can use if they want to. The reason why this configuration is kept as opaque without introducing get/set methods is to avoid adding any methods to interface later like we may want to add reader schema version  also. Other way to do is to have abstract class and have getXXX methods implemented and implementors can use that abstract class if they want to.

satishd: I may prefer passing a Cofig object defined with all the fields instead of an opaque map to `init` method.

harshach:
@arunmahadevan can you take a look again.

arunmahadevan:
@harshach With the proposed patch users need to use `AvroKafkaSpoutScheme` or implement `KafkaSourceScheme`. If we do not plan to support existing Multischeme implementations, this patch looks ok.

harshach:
@arunmahadevan as we've tight integration into registry they have to use AvroScheme for now

harshach:
+1

PR #524, opened 2017-03-14
STREAMLINE-732: Escape column names in mysql select and insert query

arunmahadevan:


shahsank3t:
+1 fixes the issue UI was facing

satishd:
+1

PR #526, opened 2017-03-16
STREAMLINE-561: MySQL drop table script easy to break due to foreign key and order of drop table

arunmahadevan:
@HeartSaVioR pls check. 
Tested by running bootstrap storage script.


HeartSaVioR:
+1 assuming it's running well with both mysql and postgresql .

PR #527, opened 2017-03-17
STREAMLINE-728 Ambari cluster import should work when Ambari is on SS…

HeartSaVioR:
…L or Kerberos

* Add options to use custom Truststore
  * This enables importing self-certificated SSL Ambari cluster
  * Also CAs which default JDK truststore doesn't cover

REST APIs on Kerberized Ambari cluster still use basic authentication (user/pass) so it doesn't require change. 

This only covers using custom truststore. Given that this option is only needed for specific case and there're lots of documents describing how to create and maintain custom truststore, I'm not sure which thing is needed for making it easier to maintain single truststore. (Also left comment on origin issue.)

Please let me know if you have ideas for that.

### DIFF
--- a/bin/import-certificate-into-truststore.sh
+++ b/bin/import-certificate-into-truststore.sh
@@ -1,0 +1,94 @@
+#!/usr/bin/env bash
+
+#
+# Copyright 2017 Hortonworks.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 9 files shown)


HeartSaVioR:
Never mind. I added script to download and import certificates to truststore.

harshach:
+1

PR #528, opened 2017-03-20
STREAMLINE-721: Topology deployment should be a state machine

arunmahadevan:


streams/actions/src/main/java/com/hortonworks/streamline/streams/actions/TopologyActionContext.java
```diff
@@ -0,0 +1,15 @@
+package com.hortonworks.streamline.streams.actions;
```
HeartSaVioR: Missing license comment.

streams/actions/src/main/java/com/hortonworks/streamline/streams/actions/topology/state/TopologyStates.java
```diff
@@ -0,0 +1,165 @@
+package com.hortonworks.streamline.streams.actions.topology.state;
```
HeartSaVioR: Missing license

streams/actions/src/test/java/com/hortonworks/streamline/streams/actions/topology/state/TopologyStateFactoryTest.java
```diff
@@ -0,0 +1,18 @@
+package com.hortonworks.streamline.streams.actions.topology.state;
```
HeartSaVioR: Missing license

streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/topology/state/TopologyState.java
```diff
@@ -0,0 +1,79 @@
+package com.hortonworks.streamline.streams.catalog.topology.state;
```
HeartSaVioR: Missing license

streams/actions/src/main/java/com/hortonworks/streamline/streams/actions/topology/state/TopologyStates.java
```diff
@@ -0,0 +1,165 @@
...
+                context.setCurrentAction("Topology suspended");
+            } catch (Exception ex) {
+                LOG.error("Error while trying to suspend the topology", ex);
+                context.setCurrentAction("Suspending the topology failed due to: " + ex);
+                throw ex;
+            }
+        }
+    };
+
+    public static final TopologyState TOPOLOGY_STATE_SUSPENDED = new TopologyState() {
```
HeartSaVioR: Would be better to allow `kill` while suspending without going through `resume`.

shahsank3t:
@HeartSaVioR 
Agree with you, deploy API should return success/error immediately and then on, UI can poll deploy state API until success/failure is returned.
This was already discussed with @harshach awhile back.

arunmahadevan:
@HeartSaVioR addressed comments, please check.

HeartSaVioR:
@arunmahadevan Thanks for the quick fix. +1

harshach:
@HeartSaVioR  Yes we need a status API.
@shahsank3t can you change the current deploy dialog box to poll this api to retrieve the progress in deploying the topology.

streams/actions/src/main/java/com/hortonworks/streamline/streams/actions/topology/state/TopologyStates.java
```diff
@@ -0,0 +1,199 @@
...
+import com.hortonworks.streamline.streams.catalog.CatalogToLayoutConverter;
+import com.hortonworks.streamline.streams.catalog.Topology;
+import com.hortonworks.streamline.streams.layout.component.TopologyDag;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * The class captures the different states a 'topology' can be in and the state transitions.
+ */
+public final class TopologyStates {
```
satishd: I do not see state transitions and valid next/previous states for each state. Can we have that documented here and it should be respectively handled in respective state implementations?

satishd:
This change looks more like to find out the current action being taken and respective status message but it is not really about implementing a state machine of the topology deployment life cycle.

streams/actions/src/main/java/com/hortonworks/streamline/streams/actions/topology/state/TopologyStates.java
arunmahadevan: Its pretty clear if you go through the code for the different states below. Each state has a set of valid actions and for each action we move to the appropriate next state.

We can put the different states and transitions in the javadoc, but it will get out of sync with the code soon. So its better we retain it in the code itself since its pretty straightforward here.

satishd: I do not see any checks in `TopologyState#deploy` whether the current state is valid. Those checks should throw IllegalStateException or any other relevant exception.

satishd: I am not particular about putting into javadoc but I do not think we are going to add new states often.

satishd: OK. It seems other invalid operations of current state throw `IllegalStateException`. 

Cause of confusion for me is different states operations are pushed into `TopologyState` class. I would have preferred to have `TopologyState` class not containing states operations but method like `execute/run(TopologyContext)` and the implementation would take care of whether the earlier state is valid or not to execute the current state. With that, no need to add any method in `TopologyState` whenever a new state is added in the existing state flow.

satishd: OK. It seems other invalid operations of current state throw IllegalStateException.

Cause of confusion for me is different states operations are pushed into `TopologyState` class. I would have preferred to have `TopologyState` class not containing states operations but method like `execute/run(TopologyContext)` and the implementation would take care of whether the earlier state is valid or not to execute the current state. With that, no need to add any method in TopologyState whenever a new state is added in the existing state flow but other dependent states need to modify accordingly for the validation.

I am fine with the current approach also as we are not going to add states often and change the implementation.

arunmahadevan: Here I followed the GOF state pattern, which has its advantages (well explained in the book) and a standard OO way of representing states and transitions. Added a note in the comment to call it out.

PR #530, opened 2017-03-28
STREAMLINE-733: Metadata API for TopologyEditor component bar

arunmahadevan:
Right now it stores with a dummy user id, changes needs to be made once https://github.com/hortonworks/streamline/pull/521 is merged to populate the actual user id.

### DIFF
--- a/bootstrap/bootstrap.sh
+++ b/bootstrap/bootstrap.sh
@@ -107,6 +107,7 @@ done
 CATALOG_ROOT_URL_PROPERTY_KEY=catalogRootUrl
 component_dir=${bootstrap_dir}/components
 service_dir=${bootstrap_dir}/services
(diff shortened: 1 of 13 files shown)


harshach:
@arunmahadevan we also need to handle this in non-secure version. In non-secure cluster the changes to the toolbar will be system-wide. Do we handle this case?

arunmahadevan:
@harshach in non-secure mode we map it with a dummy user id and the changes will be system wide. I will make changes to this PR to handle the secure/non-secure modes now that the authorization patch is merged.

arunmahadevan:
@harshach made changes to handle authorization and mapping it to anonymous user in non-secure mode and real user in secure mode.

harshach:
Thanks @arunmahadevan . +1.

PR #531, opened 2017-03-30
STREAMLINE-739: Embed jar name along with UUID

arunmahadevan:
Added 'streamline-functions' and 'notifiers' prefix while storing the udfs and notifiers.

### DIFF
--- a/streams/service/src/main/java/com/hortonworks/streamline/streams/service/UDFCatalogResource.java
+++ b/streams/service/src/main/java/com/hortonworks/streamline/streams/service/UDFCatalogResource.java
@@ -394,7 +394,7 @@ public class UDFCatalogResource {
     private String uploadJar(InputStream is, String udfName) throws IOException {
         String jarFileName;
         if (is != null) {
-            jarFileName = UUID.randomUUID().toString() + ".jar";
+            jarFileName = String.format("streamline-functions-%s.jar", UUID.randomUUID().toString());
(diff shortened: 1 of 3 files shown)


streams/service/src/main/java/com/hortonworks/streamline/streams/service/UDFCatalogResource.java
```diff
@@ -394,7 +394,7 @@ private Method findMethod(Class<?> clazz, String name) {
     private String uploadJar(InputStream is, String udfName) throws IOException {
         String jarFileName;
         if (is != null) {
-            jarFileName = UUID.randomUUID().toString() + ".jar";
+            jarFileName = String.format("streamline-functions-%s.jar", UUID.randomUUID().toString());
```
harshach: can't we take the name from uploaded file param and add UUID to it instead of hard-coding here.
We need to do this for custom processor as well

arunmahadevan: @harshach, if the same jar is uploaded again for another udf, we detect duplicates and refer to the original jar. So in this case it will cause confusion if we name it baed on the first UDF function since the same jar is used and shipped for all the udfs. Same applies to notifier.

For custom processor it already stores the user uploaded file name and does not use UUID so may be we can leave it as is.

harshach:
+1

PR #532, opened 2017-04-03
STREAMLINE-702  Factoring out join functionality from WindowedQueryBolt

roshannaik:
Core logic from WindowedQueryBolt has been moved into Storm's JoinBolt.
WindowedQueryBolt in Streamline has now become an adapter for Storm's JoinBolt... providing any StreamLine specific customizations


harshach:
+1.

PR #533, opened 2017-04-05
STREAMLINE-750 Change comma separated string into an array for servic…

HeartSaVioR:
…eBundle API having type as array.string

@shahsank3t Please pull this and test and let me know. Thanks in advance!


shahsank3t:
@HeartSaVioR 
+1, Tested the PR and it works fine.

PR #543, opened 2017-04-13
STREAMLINE-751 Email as service option in Manual Cluster

HeartSaVioR:
* implements hint provider for Notification sink
* implements service registrar for Email

@shahsank3t 
I can't test it on the UI since there seems no way to provide hints for nested fields. Other bundles don't require that, so this is a new requirement.
Below is the output format which Notification hint provider will provide.

```
{
  "3": {
    "cluster": {
      "id": 3,
      "name": "EMailTest",
      "ambariImportUrl": "",
      "description": "EmailTest",
      "timestamp": 1491991431783
    },
    "hints": {
      "properties.host": "smtp.gmail.com",
      "properties.ssl": "true",
      "properties.port": "587",
      "properties.protocol": "smtp",
      "properties.auth": "true",
      "properties.starttls": "true"
    }
  }
}
```

Please check and modify UI side accordingly. Thanks in advance!

### DIFF
--- a/bootstrap/bootstrap.sh
+++ b/bootstrap/bootstrap.sh
@@ -166,6 +166,7 @@ function add_all_bundles {
     post /servicebundles ${service_dir}/hdfs-bundle.json
     post /servicebundles ${service_dir}/hbase-bundle.json
     post /servicebundles ${service_dir}/hive-bundle.json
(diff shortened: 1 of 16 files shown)


harshach:
+1

PR #544, opened 2017-04-13
STREAMLINE-771 - Bug fix in Join processor for proper handling of implicit "streamline-event." prefix

roshannaik:


harshach:
+1

PR #550, opened 2017-04-21
STREAMLINE-780 Import Druid for auto import from Ambari

HeartSaVioR:
Just a simple patch to activate auto import Druid service for Ambari import.

@harshach Please review and comment. If we want more than providing hint for Druid sink, please let me know which information we need to have. Thanks in advance.


harshach:
Thanks @HeartSaVioR. I'll merge this in. We need to auto-fill cluster details such as overlord and other configs for druid. Similarly like we do in kafka, hbase, hive etc..

HeartSaVioR:
@harshach 
I'm not sure we can give more hints for druid sink. Do we need to provide some informations for superset?

@omkreddy 
Could you please review this and let me know if there're other configurations to be provided to the druid sink?

harshach:
@HeartSaVioR I meant to say auto-filling the configs in Druid config dialog in UI. similar to waht we do with other services

omkreddy:
@HeartSaVioR LGTM.  we have only three configs, which can be auto-filled from ambari.

PR #552, opened 2017-04-21
STREAMLINE-778: Roles API updates

arunmahadevan:
1. Changed the role hierarchy api to accept role name instead of id.
    i.e. POST /roles/{parentRoleName}/children/{childRoleName}
2. Added a new role hierarch api to accept a set of child role names
    i.e. POST /roles/{parentRoleName}/children/
         ["childRoleName", "childRoleName2",..]
3. Added description to Role


shahsank3t:
@arunmahadevan 
If the user wants to rename the role or add/remove more child roles, are we suppose to make same POST call or need to add another PUT API?

arunmahadevan:
@shahsank3t 
1. For renaming the role you can use "PUT /roles/:ID". The parent-child relations are maintained internally on ids so it will not be affected.
2. Right now there is no PUT api for updating children. POST will only add the roles you pass in the input array. (if you pass an already existing child role in POST it will throw an exception). You need to call "DELETE /roles/:PARENT_ID/children/:CHILD_ID" to remove parent-child association. If you really want, I can add a PUT api.
3. There is no key for the roles array. Its just a set of string, so you can POST a json array.

shahsank3t:
@arunmahadevan 
Yes, please add a PUT api for updating children. Mapping & then making POST/Delete API calls to update children will be complicated on UI.

arunmahadevan:
@shahsank3t, added PUT api.

shahsank3t:
@arunmahadevan 
Thanks for PUT API.

One more change we need is a way to identify predefined ones (eg: ROLE_CLUSTER_ADMIN, ROLE_TOPOLOGYY_ADMIN, etc) from the ones user defined. 
I was wondering if you could add children array into the GET API for "roles" which UI can use as a differencing field between predefined & user defined roles. This will also help us reduce 'N' number of GET children API request, where N is the number of users defined roles having children.

arunmahadevan:
@shahsank3t, when the user clicks on the role details, you can query the child roles and display. Not sure if you necessarily have to differentiate between predefined roles and user defined roles.

shahsank3t:
@arunmahadevan, on the UI we are showing predefined roles into one section and custom roles into another. So to clearly differentiate between them, it would be better to get children in the same API else UI needs to make GET children for all roles and then separate them.

arunmahadevan:
@shahsank3t, if you want a way to differentiate system roles from user defined ones, we can add a flag (say isSystemRole). Getting children by default would be expensive in case of multilevel hierarchy and not really required for differentiating.

shahsank3t:
@arunmahadevan I am fine with adding a flag as well.

shahsank3t:
@arunmahadevan, On deleting the custom role, we are getting an error saying: 
"An exception with message [Role has child roles: [Role{id=5, name='UDF_ADMIN', timestamp=1493024164911} com.hortonworks.streamline.streams.security.catalog.Role@ef4d6887, Role{id=4, name='FILE_ADMIN', timestamp=1493024157325} com.hortonworks.streamline.streams.security.catalog.Role@d2504eac]] was thrown while processing request."

Can you please make appropriate changes on the backend which delete's the associated child roles for that custom role rather than UI making all those delete API calls? (similar to add API)

arunmahadevan:
@shahsank3t, added a "system" flag in the roles. Also modified the delete API to remove the child association automatically.

shahsank3t:
+1, Thanks @arunmahadevan

PR #556, opened 2017-04-26
STREAMLINE-782: Need an API to get current user when in secure mode

arunmahadevan:
@shahsank3t , you can test it with "GET api/v1/catalog/users/current" in secure mode.

### DIFF
--- a/common/src/main/java/com/hortonworks/streamline/common/exception/service/exception/request/EntityNotFoundException.java
+++ b/common/src/main/java/com/hortonworks/streamline/common/exception/service/exception/request/EntityNotFoundException.java
@@ -42,6 +42,10 @@ public class EntityNotFoundException extends WebServiceException {
     return new EntityNotFoundException(buildMessageByID(id), cause);
   }
 
+  public static EntityNotFoundException byMessage(String msg) {
+    return new EntityNotFoundException(msg);
+  }
+
   public static EntityNotFoundException byName(String name) {
     return new EntityNotFoundException(buildMessageByName(name));
   }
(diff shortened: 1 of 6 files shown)


harshach:
@shahsank3t why we need this api. We should be able to use the cookie that get stored once auth completes

shahsank3t:
@harshach I believe in the cookie, will only have the user name and not userId. Regards of that, UI still needs to call another API to find that user's role so this API can return all the info in one go, like user information and role information.

arunmahadevan:
@shahsank3t, the proposed api will give you an user id only if the user is registered in the streamline db. For super users that are not registered (only specified as admin user in streamline.yaml) it will return you an "admin" flag indicating if the user is an admin or not. For "admin" users you don't need to check for roles, they will have all access.

shahsank3t:
@arunmahadevan oh, OK. "admin" flag is good.
IMO it would be good if we can also add user & role information in the API if a particular user is registered.

harshach:
@arunmahadevan @shahsank3t +1 on having role info. It will help in showing the right navigation options etc..

shahsank3t:
@arunmahadevan 

API: http://pm-eng1-cluster4.field.hortonworks.com:7777/api/v1/catalog/users/current

Error Response: {"responseMessage":"User 'streamline-hdp/pm-eng1-cluster4.field.hortonworks.com@EXAMPLE.COM' is not in the user database or in the admin principals"}

cc: @harshach

arunmahadevan:
@shahsank3t , i made some fixes and rebased it on top of https://github.com/hortonworks/streamline/pull/564
Can you try again?

PR #558, opened 2017-05-01
STREAMLINE-715: Metadata service should work with secure services (Ready to MERGE)

hmcl:
- Set doAsUser query param for Storm UI from Security Context
  - StormMetadataServiceTest
  - Services execution using Subject.doAs
  - Pass Subject onto ServiceMetadataResources
  - HBaseMetadataService - UserGroupInformation.setConfiguration(config);
  - HBaseMetadataResource, HBaseMetadataService with Subject.doAs(...)

### DIFF
--- a/common/src/main/java/com/hortonworks/streamline/common/function/ConsumerException.java
+++ b/common/src/main/java/com/hortonworks/streamline/common/function/ConsumerException.java
@@ -1,0 +1,5 @@
+package com.hortonworks.streamline.common.function;
+
+public interface ConsumerException<T, E extends Exception> {
+    void accept(T t) throws E;
+}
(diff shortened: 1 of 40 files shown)


hmcl:
This patch is still WIP. Namely, I am still testing several scenarios to confirm all is OK. Once I have done more extensive testing I will remove the WIP tag.

I believe that this change also addresses STREAMLINE-742. I am still testing to see if it handles all the cases.

hmcl:
Currently, and temporarily, some code has an `isSecure` flag set to true because the `SecurityContext#isSecure()` is not currently returning true without the authorizer.  It will be removed and is used only to be able to test the code.

This patch is also currently using the streamline subject in order to authenticate and test the authentication code. It will change and instantiate the appropriate subjects as I test the authorization part for the relevant services

hmcl:
@harshach this patch is now ready to merge

harshach:
@hmcl merged into master. please check & close the pr.

PR #560, opened 2017-05-05
STREAMLINE-708: Provide a search feature for entities

arunmahadevan:
Search API for entities

E.g. GET api/v1/catalog/search?namespace=topology&queryString=test&sort=timestamp&desc=true

Right now supports topology, cluster and namespace as the namespaces.


shahsank3t:
@arunmahadevan 
Currently, UI is making following calls to get all entities:
1. api/v1/catalog/namespaces?detail=true
2. api/v1/catalog/clusters?detail=true
3. api/v1/catalog/topologies?detail=true&sort=last_updated&latencyTopN=3

As you can see, we need to pass "?detail=true" (and other query params for topology) so that we get response appropriately.

Can we add similar kind of parameter to this search API or maybe a way to pass (1 or more) query params down the line to appropriate namespaces(topology/cluster/namespace) and get the response as expected.

cc: @harshach

shahsank3t:
Another thing, I have "ClusterA" and "HDF" as clusters and searching for "f" return's blank array while 'F' return's result. Maybe we should ignore the case and return the results.

arunmahadevan:
@shahsank3t , right now this is generic search api for searching entities across different namespaces. As you have listed in your comment, the query params you want to add can vary based on the namespaces (e.g. latencyTopN applies only for topologies). It will complicate things if we mix up all these into a single search api. Also getting the details for each of the entities may be unnecessary and time consuming while displaying the search results. Once the user clicks on a particular topology/namespace/cluster in the search result, you can invoke the appropriate GET api with detail flag and display the details.

harshach:
@arunmahadevan @shahsank3t I think we can skip the latency and other specific things. But detailed is important to retain the current UX after the search appears. The current one filters the same page, so we want to display the full details like we have right now on topology page.
detail & sort order would be important and can be common across the entities

arunmahadevan:
@shahsank3t , made some changes to display the details in the search results. Can you try with the latest patch ?

shahsank3t:
@arunmahadevan thanks for the changes, works for me.

PR #564, opened 2017-05-10
STREAMLINE-787: Populate the user principal into jax-rs SecurityContext from HttpServletRequest

arunmahadevan:
I did some basic testing with the secure cluster and it appears to work. Need to do more testing.

When the user is not authorized, we return a 401 NotAuthorized response but UI seems to be hanging. Observed this while adding service pool.

Corresponding users for the admin principals listed in the streamline.yaml should be added via the user api before they can create entities. (E.g streamline-hdf user needs to be added in the streamline db before streamline-hdf@EXAMPLE.COM). This can be done by the same admin user.

### DIFF
--- a/streams/authorizer/src/main/java/com/hortonworks/streamline/streams/security/authentication/StreamlineKerberosRequestFilter.java
+++ b/streams/authorizer/src/main/java/com/hortonworks/streamline/streams/security/authentication/StreamlineKerberosRequestFilter.java
@@ -1,0 +1,57 @@
+/**
+ * Copyright 2017 Hortonworks.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+
+ *   http://www.apache.org/licenses/LICENSE-2.0
+
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
+package com.hortonworks.streamline.streams.security.authentication;
+
+import com.hortonworks.streamline.common.exception.service.exception.request.WebserviceAuthorizationException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.ws.rs.container.ContainerRequestContext;
+import javax.ws.rs.container.ContainerRequestFilter;
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.SecurityContext;
+import javax.ws.rs.ext.Provider;
+import java.io.IOException;
+import java.security.Principal;
+
+import static com.hortonworks.streamline.streams.security.authentication.StreamlineSecurityContext.KERBEROS_AUTH;
+
+@Provider
+public class StreamlineKerberosRequestFilter implements ContainerRequestFilter {
+    private static final Logger LOG = LoggerFactory.getLogger(StreamlineKerberosRequestFilter.class);
+
+    @Context
+    private HttpServletRequest httpRequest;
+
+    @Override
+    public void filter(ContainerRequestContext requestContext) throws IOException {
+        Principal principal = httpRequest.getUserPrincipal();
+        String scheme = requestContext.getUriInfo().getRequestUri().getScheme();
+
+        LOG.debug("Method: {}, AuthType: {}, RemoteUser: {}, UserPrincipal: {}, Scheme: {}",
+                httpRequest.getMethod(), httpRequest.getAuthType(),
+                httpRequest.getRemoteUser(), principal, scheme);
+
+        if (principal == null || !httpRequest.getAuthType().equalsIgnoreCase(KERBEROS_AUTH)) {
+            throw new WebserviceAuthorizationException("Not authorized");
+        }
+
+        SecurityContext securityContext = new StreamlineSecurityContext(principal, scheme, KERBEROS_AUTH);
+        LOG.debug("SecurityContext {}", securityContext);
+        requestContext.setSecurityContext(securityContext);
+    }
+}
--- a/streams/authorizer/src/main/java/com/hortonworks/streamline/streams/security/impl/DefaultStreamlineAuthorizer.java
+++ b/streams/authorizer/src/main/java/com/hortonworks/streamline/streams/security/impl/DefaultStreamlineAuthorizer.java
@@ -18,6 +18,7 @@ package com.hortonworks.streamline.streams.security.impl;
 import com.hortonworks.streamline.streams.security.AuthenticationContext;
 import com.hortonworks.streamline.streams.security.AuthorizationException;
 import com.hortonworks.streamline.streams.security.Permission;
+import com.hortonworks.streamline.streams.security.SecurityUtil;
 import com.hortonworks.streamline.streams.security.StreamlineAuthorizer;
 import com.hortonworks.streamline.streams.security.catalog.AclEntry;
 import com.hortonworks.streamline.streams.security.catalog.Role;
@@ -30,6 +31,7 @@ import java.util.EnumSet;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 public class DefaultStreamlineAuthorizer implements StreamlineAuthorizer {
     private static final Logger LOG = LoggerFactory.getLogger(DefaultStreamlineAuthorizer.class);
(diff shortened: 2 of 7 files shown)


harshach:
@arunmahadevan  I think adding user via manual operation going to be confusing and tedious for the admin. Why not we auto-register the user as soon as they login?

harshach:
@arunmahadevan also admin principal should be by default be registerd. This shouldn't be an extra step.  Also we need to do AuthToLocal part which pulls out  the principal name. In Users section we should be able to store both principal name and full principal including the domain as an extra field

streams/authorizer/src/main/java/com/hortonworks/streamline/streams/security/authentication/StreamlineKerberosRequestFilter.java
```diff
@@ -0,0 +1,54 @@
...
+
+import javax.servlet.http.HttpServletRequest;
+import javax.ws.rs.container.ContainerRequestContext;
+import javax.ws.rs.container.ContainerRequestFilter;
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.SecurityContext;
+import java.io.IOException;
+import java.security.Principal;
+
+public class StreamlineKerberosRequestFilter implements ContainerRequestFilter {
```
priyank5485: Is there a reason for adding this as a ContainerRequestFilter rather than a ServletFilter. We already have a method called addServletFilters in StreamlineApplication and it picks up all filters configured in streamline.yaml

webservice/src/main/java/com/hortonworks/streamline/webservice/StreamlineApplication.java
```diff
@@ -252,7 +253,7 @@ private void registerResources(StreamlineConfiguration configuration, Environmen
             authorizerConfig.put(DefaultStreamlineAuthorizer.CONF_CATALOG_SERVICE, securityCatalogService);
             authorizerConfig.put(DefaultStreamlineAuthorizer.CONF_ADMIN_PRINCIPALS, authorizerConf.getAdminPrincipals());
             authorizer.init(authorizerConfig);
-            environment.jersey().register(new StreamlineBasicAuthorizationRequestFilter());
+            environment.jersey().register(new StreamlineKerberosRequestFilter());
```
priyank5485: Does this add the filter at the end in filter chain? If so we are okay. If not we need to make sure the AuthenticationFilter is executed before this one

priyank5485:
@arunmahadevan Just want to confirm about how SecurityContext is populated on requests to different endpoints. We have SecurityContext with @Context annotation in all the resources' endpoints. Does jersey populate that SecurityContext using the filter you added or is UI expected to set that SecurityContext in the request?

harshach:
Deployed the latest master here http://pm-eng2-cluster2.field.hortonworks.com:7777/#/service-pool?_k=w9oe6r
you need to copy /tmp/streamline.service.keytab from that machine onto your local and also copy /etc/krb5.conf 
Now I don't get authorization errors but when I try to create service pool running into this error
**An exception with message [No such user 'streamline-hdf'] was thrown while processing request.
Hide**
Few thing we need to address is
1. Make user registration transparent i.e we don't need to ask admin to create a user again.
2. We need to add authToLocal rule so that we don't need to add the following in config
authorizerConfiguration:
  className: "com.hortonworks.streamline.streams.security.impl.DefaultStreamlineAuthorizer"
  adminPrincipals:
    - "streamline-hdf@EXAMPLE.COM"

it should be just "streamline-hdf".

harshach:
@arunmahadevan  please ignore the second comment part. It does work if I put in "streamline-hdf"

@HeartSaVioR  I tried to import the a cluster using service pool
http://pm-eng2-cluster2.field.hortonworks.com:7777/#/service-pool?_k=s02p9f

right now I am getting "An exception with message [No such user 'streamline-hdf'] was thrown while processing request." 
But this still creates the service pool with empty. When I try to delete it and UI shows that the service pool is used by some environment where as its not being used anywhere. When I try again it does delete it.

hmcl:
@arunmahadevan I read your comment in the [JIRA](https://hwxiot.atlassian.net/browse/STREAMLINE-786?focusedCommentId=26001&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-26001). 

Does this change guarantee that `SecurityContext#isSecure()=true` when using kerberos authentication, or is it only true when using SSL, and for Kerberos, since it uses plain text, it returns false ?

If it returns false for Kerberos, how can I support the mechanism as illustrated in [here](https://github.com/hmcl/streamline-hortonworks/blob/58ffebd7fbfb7858544cff84890e0b8d02cdeee2/streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/service/metadata/HiveMetadataService.java#L130)?

streams/authorizer/src/main/java/com/hortonworks/streamline/streams/security/authentication/StreamlineKerberosRequestFilter.java
arunmahadevan: ContainerRequestFilter is specific to jax-rs and makes it easy to inject context into jersey REST methods via @Context annotation.

webservice/src/main/java/com/hortonworks/streamline/webservice/StreamlineApplication.java
arunmahadevan: This is at jax-rs level so will kick in only after all the servlet filters in the chain are executed.

arunmahadevan:
@priyank5485 
> We have SecurityContext with @Context annotation in all the resources' endpoints. Does jersey populate that SecurityContext using the filter you added or is UI expected to set that SecurityContext in the request?

This is added by the jersey filter. UI doesn't need to do anything.

@harshach 
>Make user registration transparent i.e we don't need to ask admin to create a user again.

Let me see if it can be done without any user action at application startup, if not do it via bootstrap.

@hmcl 
>If it returns false for Kerberos, how can I support the mechanism as illustrated in here?

Security context has a `getAuthenticationScheme` method. I will make some changes to return a 'constant' for kerberos based authentication and you can use this.
https://docs.oracle.com/javaee/7/api/javax/ws/rs/core/SecurityContext.html#getAuthenticationScheme--

harshach:
@arunmahadevan for transparent user registration while we are doing this lets add role "ANONYMOUS" to the non-admin users. This way we can display a error page and admin can filter users who are ANONYMOUS and grant appropriate roles.

arunmahadevan:
@hmcl, you can do 
`securityContext.getAuthenticationScheme() == StreamlineSecurityContext.KERBEROS_AUTH`
to check if the authentication is kerberos.

@harshach we are adding an anonymous user in bootstrap.sh. We can grant anonymous users the required permissions and role. Other user can be added by the admin. We cant automatically add user entry for normal users when they login since it requires "security admin" role to add users and we are executing with the privileges of the logged in user.

arunmahadevan:
Added option to add user entry for the admin user listed in streamline.yaml if not already present.

harshach:
@arunmahadevan I'll be merging this in. I think what we should atleast have is a staging of users.
So that admin can go and check and give permissions to the user. We can discuss this further.
But having admin to be registered via bootstrap is going to be a problem. In general once the admin is listed they should have been registered in the database.
for example lets say if we add "streamline" as admin in the config and during the install time Ambari runs bootstrap once and we get "streamline" registered as user. Later if we add another user "harsha" to admin Ambari is not going to run bootstrap . This will be a bit of expectation mismatch as user might think they configured the "harsha" as admin but they have to do this extra step. 
I can understand this for users but for "admins" we should do better.

streams/authorizer/src/main/java/com/hortonworks/streamline/streams/security/authentication/StreamlineKerberosRequestFilter.java
```diff
@@ -0,0 +1,55 @@
...
+import javax.ws.rs.container.ContainerRequestContext;
+import javax.ws.rs.container.ContainerRequestFilter;
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.SecurityContext;
+import java.io.IOException;
+import java.security.Principal;
+
+import static com.hortonworks.streamline.streams.security.authentication.StreamlineSecurityContext.KERBEROS_AUTH;
+
+public class StreamlineKerberosRequestFilter implements ContainerRequestFilter {
```
hmcl: I am not 100% sure if it is relevant here, but the documentation for `ContainerRequestFilter` states:

"_Filters implementing this interface must be annotated with @Provider to be discovered by the JAX-RS runtime._"

hmcl:
+1 LGTM, just some minor neats. 

@arunmahadevan can you please squash the commits once you hare done addressing all the other reviewers comments.

harshach:
@arunmahadevan with this patch logged in as admin user I am getting this for notifiers
{code}

HTTP/1.1 500 Internal Server Error
Date: Fri, 12 May 2017 02:59:01 GMT
Content-Type: application/json
Content-Length: 104

{"responseMessage":"An exception with message [Permission denied] was thrown while processing request."}
Adding Console notifier
HTTP/1.1 100 Continue

HTTP/1.1 500 Internal Server Error
Date: Fri, 12 May 2017 02:59:01 GMT
Content-Type: application/json
Content-Length: 104

{"responseMessage":"An exception with message [Permission denied] was thrown while processing request."}
{code}

streams/authorizer/src/main/java/com/hortonworks/streamline/streams/security/impl/DefaultStreamlineAuthorizer.java
```diff
@@ -38,14 +40,39 @@
...
+                        return false;
+                    } else {
+                        return true;
+                    }
+                })
+                .forEach(name -> {
+                    User user = new User();
+                    user.setName(name);
+                    user.setEmail(name + "@auto-generated");
+                    User addedUser = catalogService.addUser(user);
```
arunmahadevan: This is used to log the added user below.

arunmahadevan:
@harshach, with this patch the users listed under 'adminPrincipals' in streamline.yaml are automatically added to db when streamline server is started.

Are you getting this error while running the bootstrap ? While uploading notifier we check for role 'ROLE_NOTIFIER_ADMIN' and if not return a 401. Here it seems to throw 500.

arunmahadevan:
@harshach , that error was due to file permssions.

The directory specified in fileStorageConfiguration (/hdp/streamline/jars) was having rwx only for root, but the streamline process is running as "streamline" user.

```
[root@pm-eng1-cluster4 streamline]# ls -l /hdp/streamline/
total 0
drwxr-xr-x. 2 root root 6 May 12 02:18 jars
drwxr-xr-x. 2 root root 6 May 12 02:18 topology_test_results
```

After given rwx for others to the jars directory, I am able to run the bootstrap. If these directories are auto created by some scripts, it can be given the right permissions at creation time.

PR #566, opened 2017-05-11
STREAMLINE-788. bootstrap scripts should work in secure mode.

harshach:


harshach:
tested on both secure and non-secure clusters. Users need to kinit before running this in secure cluster

PR #568, opened 2017-05-12
STREAMLINE-763 Topology should be able to deployed on Secured Storm c…

HeartSaVioR:
…luster

* Also request Storm REST API with Streamline's Subject to access secured Storm cluster

Change-Id: I4d6f24eb5f588ad04a3283624d8e0e503542cce5

This commit was not merged yet since STREAMLINE-763 and STREAMLINE-764 went concurrently.

@harshach I think we need to get this for requesting Storm REST API to the secure cluster.


HeartSaVioR:
Also fixed a bug which is not occurred from IDE run. The changes are also applied to the `security` branch.

PR #572, opened 2017-05-15
STREAMLINE-769 Streamline changes for auto cred

HeartSaVioR:
WIP: I checked that this patch provides necessary configuration to the topology config so I think it's OK to be reviewed, but I didn't do end-to-end system test yet.

This patch is on top of #568 so please check the commits only for STREAMLINE-769.

To activate auto credential, user needs to fill out "principal" and "keytab path" in nimbus node for each cluster to the topology config. Streamline finds Cluster entity by name so user needs to fill out cluster's name as well for each cluster.
(I know filling out cluster name by user is bad but I don't know how we can provide hint for this.)

streamline-topology-config-cluster-security-config

Currently HDFS and HBase are supported. Hive and Kafka should be done after getting auto-creds for Hive and Kafka.

`putAutoTokenDelegationConfig` is the main function which handles the feature.

Please refer to below link to which configuration this patch will push to the Storm topology config:

* https://github.com/apache/storm/blob/1.x-branch/docs/storm-hdfs.md#using-hdfs-delegation-tokens
* https://github.com/apache/storm/blob/1.x-branch/docs/storm-hbase.md
  * anchor is not made, please find "delegation tokens" 

Please review and comment. Thanks in advance!

cc. @arunmahadevan since you refactored `storm-autocreds`.

### DIFF
--- a/bootstrap/components/topology/storm-topology-component.json
+++ b/bootstrap/components/topology/storm-topology-component.json
@@ -55,6 +55,40 @@
             "defaultValue": "hdfs://localhost:9000/tmp/hbase"
           }
         ]
+      },
+      {
+        "uiName": "Clusters Security Config",
+        "fieldName": "clustersSecurityConfig",
+        "tooltip": "Clusters Security Configuration",
+        "isOptional": true,
+        "type": "array.object",
+        "defaultValue": null,
+        "fields": [
+          {
+            "uiName": "Cluster Name",
+            "fieldName": "clusterName",
+            "tooltip": "Name of the cluster (same as service pool)",
+            "isOptional": false,
+            "type": "string",
+            "defaultValue": null
+          },
+          {
+            "uiName": "Principal",
+            "fieldName": "principal",
+            "tooltip": "The principal for auto delegation token",
+            "isOptional": false,
+            "type": "string",
+            "defaultValue": null
+          },
+          {
+            "uiName": "Keytab Path",
+            "fieldName": "keytabPath",
+            "tooltip": "The path of keytab file in Nimbus node to log in with provided principal",
+            "isOptional": false,
+            "type": "string",
+            "defaultValue": null
+          }
+        ]
       }
     ]
   }
--- a/streams/runners/storm/actions/src/main/java/com/hortonworks/streamline/streams/actions/storm/topology/ServiceConfigurationReader.java
+++ b/streams/runners/storm/actions/src/main/java/com/hortonworks/streamline/streams/actions/storm/topology/ServiceConfigurationReader.java
@@ -30,9 +30,11 @@ public class ServiceConfigurationReader implements ServiceConfigurationReadable 
 
     @Override
     public Map<Long, Map<String, String>> readAllClusters(String serviceName) {
-        // FIXME: assertions or some validations
-
         Namespace namespace = environmentService.getNamespace(namespaceId);
+
+        if (namespace == null) {
+            throw new IllegalArgumentException("Namespace " + namespaceId + " doesn't exist.");
+        }
 
         Long namespaceId = namespace.getId();
 
@@ -85,10 +87,11 @@ public class ServiceConfigurationReader implements ServiceConfigurationReadable 
     @Override
     public Map<String, String> read(String clusterName, String serviceName) {
         Collection<Cluster> clusters = environmentService.listClusters(
-                Lists.newArrayList(new QueryParam("namespaceId", String.valueOf(namespaceId)),
-                        new QueryParam("name", clusterName)));
+                Lists.newArrayList(new QueryParam("name", clusterName)));
 
-        // FIXME: assertions or some validations
+        if (clusters.isEmpty()) {
+            throw new IllegalArgumentException("Cluster with name " + clusterName + " doesn't exist.");
+        }
 
         Cluster cluster = clusters.iterator().next();
         return read(cluster.getId(), serviceName);
--- a/streams/runners/storm/actions/src/main/java/com/hortonworks/streamline/streams/actions/storm/topology/ServiceConfigurationReader.java
+++ b/streams/runners/storm/actions/src/main/java/com/hortonworks/streamline/streams/actions/storm/topology/ServiceConfigurationReader.java
@@ -1,0 +1,96 @@
+package com.hortonworks.streamline.streams.actions.storm.topology;
+
+import com.google.common.collect.Lists;
+import com.hortonworks.streamline.common.QueryParam;
+import com.hortonworks.streamline.streams.catalog.Cluster;
+import com.hortonworks.streamline.streams.catalog.Namespace;
+import com.hortonworks.streamline.streams.catalog.NamespaceServiceClusterMapping;
+import com.hortonworks.streamline.streams.catalog.ServiceConfiguration;
+import com.hortonworks.streamline.streams.cluster.service.EnvironmentService;
+import com.hortonworks.streamline.streams.storm.common.ServiceConfigurationReadable;
+
+import java.io.IOException;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+public class ServiceConfigurationReader implements ServiceConfigurationReadable {
+
+    private final EnvironmentService environmentService;
+    private final long namespaceId;
+
+    public ServiceConfigurationReader(EnvironmentService environmentService, long namespaceId) {
+        this.environmentService = environmentService;
+        this.namespaceId = namespaceId;
+    }
+
+    @Override
+    public Map<Long, Map<String, String>> readAllClusters(String serviceName) {
+        // FIXME: assertions or some validations
+
+        Namespace namespace = environmentService.getNamespace(namespaceId);
+
+        Long namespaceId = namespace.getId();
+
+        Collection<NamespaceServiceClusterMapping> mappings = environmentService.listServiceClusterMapping(namespaceId, serviceName);
+
+        List<Long> clusters = mappings.stream()
+                .map(mapping -> mapping.getClusterId())
+                .collect(Collectors.toList());
+
+        Map<Long, Map<String, String>> retMap = new HashMap<>();
+        clusters.forEach(c -> {
+            Map<String, String> flattenConfig = read(c, serviceName);
+            retMap.put(c, flattenConfig);
+        });
+
+        return retMap;
+    }
+
+    @Override
+    public Map<String, String> read(Long clusterId, String serviceName) {
+        Collection<NamespaceServiceClusterMapping> mappings = environmentService.listServiceClusterMapping(namespaceId, serviceName);
+        boolean associated = mappings.stream().anyMatch(map -> map.getClusterId().equals(clusterId));
+        if (!associated) {
+            return Collections.emptyMap();
+        }
+
+        Long serviceId = environmentService.getServiceIdByName(clusterId, serviceName);
+        if (serviceId == null) {
+            throw new IllegalStateException("Cluster " + clusterId + " is associated to the service " + serviceName +
+                " for namespace " + namespaceId + ", but actual service doesn't exist.");
+        }
+
+        Collection<ServiceConfiguration> serviceConfigurations = environmentService.listServiceConfigurations(serviceId);
+
+        Map<String, String> flattenConfig = new HashMap<>();
+
+        // FIXME: should resolve the priority between configurations if exists
+        serviceConfigurations.forEach(sc -> {
+            try {
+                Map<String, String> configurationMap = sc.getConfigurationMap();
+                flattenConfig.putAll(configurationMap);
+            } catch (IOException e) {
+                throw new RuntimeException("Can't read configuration from service configuration - ID: " + sc.getId());
+            }
+        });
+
+        return flattenConfig;
+    }
+
+    @Override
+    public Map<String, String> read(String clusterName, String serviceName) {
+        Collection<Cluster> clusters = environmentService.listClusters(
+                Lists.newArrayList(new QueryParam("namespaceId", String.valueOf(namespaceId)),
+                        new QueryParam("name", clusterName)));
+
+        // FIXME: assertions or some validations
+
+        Cluster cluster = clusters.iterator().next();
+        return read(cluster.getId(), serviceName);
+    }
+}
--- a/streams/runners/storm/actions/src/main/java/com/hortonworks/streamline/streams/actions/storm/topology/StormTopologyActionsImpl.java
+++ b/streams/runners/storm/actions/src/main/java/com/hortonworks/streamline/streams/actions/storm/topology/StormTopologyActionsImpl.java
@@ -16,8 +16,17 @@
 package com.hortonworks.streamline.streams.actions.storm.topology;
 
 import com.google.common.base.Joiner;
+import com.google.common.collect.Lists;
 import com.hortonworks.streamline.common.exception.service.exception.request.TopologyAlreadyExistsOnCluster;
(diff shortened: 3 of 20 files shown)


bootstrap/components/topology/storm-topology-component.json
```diff
@@ -55,6 +55,40 @@
             "defaultValue": "hdfs://localhost:9000/tmp/hbase"
           }
         ]
+      },
+      {
```
arunmahadevan: Is this always going to be shown or only if streamline is running in kerberos mode ? Is it optional in non-secure mode ?

bootstrap/components/topology/storm-topology-component.json
```diff
@@ -55,6 +55,40 @@
...
+        "uiName": "Clusters Security Config",
+        "fieldName": "clustersSecurityConfig",
+        "tooltip": "Clusters Security Configuration",
+        "isOptional": true,
+        "type": "array.object",
+        "defaultValue": null,
+        "fields": [
+          {
+            "uiName": "Cluster Name",
```
arunmahadevan: Is there someway to auto suggest the list of cluster names? For now maybe user entering the values is fine.

HeartSaVioR: I'm not sure. It might need UI side change.

@shahsank3t 
We would like to get hints for cluster names. 
If we can provide hints for `clustersSecurityConfig. clusterName` (please note that clustersSecurityConfig is an array) we can change the type of `clustersSecurityConfig. clusterName` to `enumstring` and put cluster names to the options.
Please let me know the possibility of implementing this.

bootstrap/components/topology/storm-topology-component.json
HeartSaVioR: It's always shown since field list are not dynamic. 
And yes, it's optional. Users don't need to add the config if topology doesn't contain secured cluster or auto token delegation is not needed.

streams/runners/storm/actions/src/main/java/com/hortonworks/streamline/streams/actions/storm/topology/StormTopologyActionsImpl.java
```diff
@@ -490,6 +507,101 @@ private String createYamlFile (TopologyLayout topology) throws Exception {
...
+
+                            String clusterKey = clusterKeyPrefix + clusterName;
+                            topologyConfig.put(clusterKey, conf);
+                            clusterKeys.add(clusterKey);
+                        }
+                    });
+
+            topologyConfig.put(credentialConfigKeyName, clusterKeys);
+
+            Optional<List<String>> autoCredentialsOptional = topologyConfig.getAnyOptional("topology.auto-credentials");
```
HeartSaVioR: Here we also add `topology.auto-credentials` when we add credential config key for the service.

streams/runners/storm/actions/src/main/java/com/hortonworks/streamline/streams/actions/storm/topology/StormTopologyActionsImpl.java
```diff
@@ -490,6 +507,101 @@ private String createYamlFile (TopologyLayout topology) throws Exception {
...
+                                                    Map<String, Map<String, String>> clusterToConfiguration,
+                                                    List<Class<?>> outputComponentClasses,
+                                                    List<Class<?>> inputComponentClasses,
+                                                    String serviceName, String clusterKeyPrefix,
+                                                    String keytabPathKeyName, String principalKeyName,
+                                                    String credentialConfigKeyName,
+                                                    String topologyAutoCredentialClassName) {
+
+        boolean componentExists = false;
+        for (OutputComponent outputComponent : topologyDag.getOutputComponents()) {
```
HeartSaVioR: Here we check if topology has one or more sources related to such service.

streams/runners/storm/actions/src/main/java/com/hortonworks/streamline/streams/actions/storm/topology/StormTopologyActionsImpl.java
```diff
@@ -490,6 +507,101 @@ private String createYamlFile (TopologyLayout topology) throws Exception {
...
+            for (Class<?> clazz : outputComponentClasses) {
+                if (outputComponent.getClass().isAssignableFrom(clazz)) {
+                    componentExists = true;
+                    break;
+                }
+            }
+        }
+
+        if (!componentExists) {
```
HeartSaVioR: And here we check if topology has one or more sinks related to such service only when topology doesn't have any source related to such service.

streams/runners/storm/actions/src/main/java/com/hortonworks/streamline/streams/actions/storm/topology/StormTopologyActionsImpl.java
```diff
@@ -490,6 +507,101 @@ private String createYamlFile (TopologyLayout topology) throws Exception {
...
+                        break;
+                    }
+                }
+            }
+        }
+
+        if (componentExists) {
+            List<String> clusterKeys = new ArrayList<>();
+
+            clusterToConfiguration.keySet()
```
HeartSaVioR: We are using Streamline topology security config, so if users don't set anything for that, we don't add anything to Storm topology config.

streams/runners/storm/actions/src/main/java/com/hortonworks/streamline/streams/actions/storm/topology/ServiceConfigurationReader.java
```diff
@@ -0,0 +1,99 @@
...
+                throw new RuntimeException("Can't read configuration from service configuration - ID: " + sc.getId());
+            }
+        });
+
+        return flattenConfig;
+    }
+
+    @Override
+    public Map<String, String> read(String clusterName, String serviceName) {
+        Collection<Cluster> clusters = environmentService.listClusters(
```
HeartSaVioR: Here we find cluster entity via its name.

streams/runners/storm/actions/src/main/java/com/hortonworks/streamline/streams/actions/storm/topology/ServiceConfigurationReader.java
```diff
@@ -0,0 +1,99 @@
...
+                Lists.newArrayList(new QueryParam("name", clusterName)));
+
+        if (!clusters.isEmpty()) {
+            throw new IllegalArgumentException("Cluster with name " + clusterName + " doesn't exist.");
+        }
+
+        Cluster cluster = clusters.iterator().next();
```
HeartSaVioR: We're assuming we only have one matched cluster entity.

streams/runners/storm/actions/src/main/java/com/hortonworks/streamline/streams/actions/storm/topology/ServiceConfigurationReader.java
```diff
@@ -0,0 +1,99 @@
...
+            throw new IllegalStateException("Cluster " + clusterId + " is associated to the service " + serviceName +
+                " for namespace " + namespaceId + ", but actual service doesn't exist.");
+        }
+
+        Collection<ServiceConfiguration> serviceConfigurations = environmentService.listServiceConfigurations(serviceId);
+
+        Map<String, String> flattenConfig = new HashMap<>();
+
+        // FIXME: should resolve the priority between configurations if exists
+        serviceConfigurations.forEach(sc -> {
```
HeartSaVioR: Below lines flatten the configurations for such service into a map.

streams/runners/storm/actions/src/main/java/com/hortonworks/streamline/streams/actions/storm/topology/ServiceConfigurationReader.java
```diff
@@ -0,0 +1,99 @@
...
+            return Collections.emptyMap();
+        }
+
+        Long serviceId = environmentService.getServiceIdByName(clusterId, serviceName);
+        if (serviceId == null) {
+            throw new IllegalStateException("Cluster " + clusterId + " is associated to the service " + serviceName +
+                " for namespace " + namespaceId + ", but actual service doesn't exist.");
+        }
+
+        Collection<ServiceConfiguration> serviceConfigurations = environmentService.listServiceConfigurations(serviceId);
```
HeartSaVioR: Here we get all service configurations for such service.

streams/runners/storm/actions/src/main/java/com/hortonworks/streamline/streams/actions/storm/topology/ServiceConfigurationReader.java
```diff
@@ -0,0 +1,99 @@
...
+            Map<String, String> flattenConfig = read(c, serviceName);
+            retMap.put(c, flattenConfig);
+        });
+
+        return retMap;
+    }
+
+    @Override
+    public Map<String, String> read(Long clusterId, String serviceName) {
+        Collection<NamespaceServiceClusterMapping> mappings = environmentService.listServiceClusterMapping(namespaceId, serviceName);
```
HeartSaVioR: Here we get namespace-service mapping, and see whether given cluster is associated for the service. If it isn't, don't retrieve configurations.

HeartSaVioR:
Now this patch also has 42d97a3 which is also available from #574 . I'll rebase this branch when #574 gets merged.

streams/runners/storm/actions/src/main/java/com/hortonworks/streamline/streams/actions/storm/topology/ServiceConfigurationReader.java
arunmahadevan: For hdfs in my test env, I see the following 4 configurations viz. hadoop-policy, core-site, hdfs-site and hadoop-env.

core-site and hdfs-site are fine, but I am not sure about the other two and if adding them will cause any un-intended effect. These key values finally gets inserted into a hadoop Configuration object. (Configuration for hdfs, HbaseConfiguration for hbase etc).

harshach:
@HeartSaVioR can you check with @omkreddy about hive integration.
@priyank5485 should we use topology setting's keytab for kafka as well instead of taking it from source/sink configs?

priyank5485:
@harshach The topology settings keytab here will be the location on nimbus since its doing delegation token for auto creds. It can be the same keytab location on worker as well. It will probably also be for a principal that it is getting the tokens on behalf of the actual user. If its safe to presume that principal getting the delegation token will be same as kafka client then we can use the topology settings keytab. Otherwise we should keep it separate for kafka.

HeartSaVioR:
@harshach 
I'm on testing with only HDFS and facing some issues on this. (auth for kafka and schema registry) 
I'll address all of issues facing while testing first, and address Hive side. Please let me know if you would like to see the change for Hive first.

bootstrap/components/topology/storm-topology-component.json
shahsank3t: @HeartSaVioR yes, we can change `"clusterName"` type from "string" to "enum" and pass the names to options, but I am wondering whether this cluster names option will be pre-defined into this JSON file or will be coming via some API's depending on cluster and environment?

HeartSaVioR: @shahsank3t It should be provided dynamically (on demand) like hint. That's why I'm wondering it could be possible.

harshach:
@HeartSaVioR check with @priyank5485 he will have answers for the kafka related stuff.

HeartSaVioR:
@harshach I've got resolved Kafka & SR related issues (will submit patches soon) but now facing HDFS issue.

HeartSaVioR:
UPDATE: I've resolved Kafka side & SR side.

Below patch is needed for resolving Kafka side: #576
For SR we should add SR jass config manually for each storm_jass.conf

I'm now stuck with HDFS auto creds and trying to resolve the issue. @arunmahadevan is also looking into this.

HeartSaVioR:
UPDATE: just found and fixed a bug regarding HDFS autocreds.

- JIRA: https://issues.apache.org/jira/browse/STORM-2520
- PR
  - https://github.com/apache/storm/pull/2123 (master)
  - https://github.com/apache/storm/pull/2124 (1.x)

I fixed another bug on Storm while testing, http://issues.apache.org/jira/browse/STORM-2518

After applying two patches, I can see topology configured to `Kafka -> Rule -> HDFS` working well.

I'll continue testing on HBase side tomorrow.

HeartSaVioR:
UPDATE: Just finished testing HBase side. Fixed issues while testing.

We need to have another quick fix for HBase because of conflict between slf4j-log4j12 and log4j-over-slf4j. Since the quick fix should be addressed regardless of security, I'll create another PR for that.

HeartSaVioR:
Quick fix for HBase is available at #578.

harshach:
@HeartSaVioR not sure if you are running the master but recently merged in @hmcl metadata patch that might fix the hbase issue.
@hmcl can you confirm

HeartSaVioR:
@harshach 
Metadata patch is already applied. I had a quick look and found we also need to modify HintProvider to get security context and/or subject, and pass them to HBaseMetadata. So maybe some more works to do.

HeartSaVioR:
UPDATE: Finally applied and tested Hive auto creds. 

Also tested with secure HDFS / HBase / Hive in same cluster altogether.

Btw, storm-hive `1.0.2.3.0.0.0-310` doesn't contain the change for autocreds, so I need to update bundle for hive sink manually to storage. (changed to `1.1.0.2.6.1.0-94`)

It will be resolved when both target HDP and HDF versions apply Hive storm-creds patch.

HeartSaVioR:
Btw, Kafka sasl jaas config doesn't work with "_HOST" representation of principal.
(Failed to login with keytab which is for principal with actual hostname as host part.)

While testing I just used actual hostname on principal and try to locate worker to the specific node, which doesn't make sense in production.

@priyank5485 @harshach Any ideas on this? Is it up to the users to take care of?

HeartSaVioR:
Maybe implementing delegation token support on Kafka would be best, though it would take numerous efforts.

HeartSaVioR:
#581 addresses hint providers to handle SecurityContext and Subject. I applied the patch for testing env. and succeeded to get table lists.

priyank5485:
@HeartSaVioR When you say _HOST representation of principal do you mean a principal of the form user/hostname@REALM ? What is the value you used for principal? When you do a list_principals in kadmin does it return the exact same principal? I think typically principals with hostnames are service principals. I dont think we need to worry about the scenario you mentioned but will still wait for your response to think  more about it.

HeartSaVioR:
@priyank5485 

For example, I can use `hive/_HOST@EXAMPLE.COM` to the cluster principal of topology config even cluster keytab actually has `hive/<FQDN>@EXAMPLE.COM`. So even if the cluster uses Nimbus H/A (meaning we can't predict which node AutoCreds will run), it will work clearly.

Kafka doesn't work with such approach. It requires principal should be same as keytab has.
We don't know which node workers containing Kafka spout will be launched, and even worse, Kafka spout will be launched from multiple nodes.
So users should make sure that user should prepare principal without FQDN part for every worker nodes and corresponding keytab, in order to make principal same across the nodes and we can finally put principal to the Kafka spout config.

Please note that Hive seems to disallow principal which doesn't have FQDN part, at least for getting delegation token. So what Kafka requires and what Hive requires are opposite and user should be aware of this.

HeartSaVioR:
@priyank5485 
Ah OK. I'm not familiar with Kerberos auth and principal convention, so thought that FQDN part is highly recommended to have.
So you're saying principal will not have FQDN part so no need to worry about that, right? Then we don't need to worry about that. One thing to know that is that Hive threw error when I set principal to non-FQDN when getting delegation token. You may want to test this and check my observation for Hive.

PR #573, opened 2017-05-15
STREAMLINE-778: Authorizer changes

arunmahadevan:
1. Auto assign ADMIN role for admin principals listed in streamline.yaml
2. Change bootstrap.sh to handle admin ROLE being auto created during server startup
3. Introduce owner and grant fields so that owners of entities can grant access to others without requiring SECURITY_ADMIN role
4. Three higher level roles added. ROLE_ADMIN, ROLE_OPERATIONS, ROLE_DEVELOPER
5. Added a field within role metadata to indicate which menu items UI needs to display based on the user ROLE
6. Removed FILE_ADMIN role
7. ROLE_ADMIN is now system level role and has privileges to all operations
8. List ACLs will filter and display relevant acl entries for the currently logged in user. This can be further filetered by passing query params.
9. ADD/UPDATE/DELETE/GET Acls will now work based on ACL entities of the currently logged in user (or SECURITY_ADMIN role which has access to all acls)

@harshach @shahsank3t , these are the authorizer changes based on our discussion. I haven't tested it out extensively, but you can review and see if I missed something and try it out in some env if possible.


harshach:
@arunmahadevan need some changes this. I sent you an email.

1. We need to add for Operations, a blanket all access on topologies even if they not created by them
2. Developer blanket read access on all the other topologies that not created by them. Should've read access on all topology resources 
3. Let's add Role_Analyst they should've blanket read access only on all topologies.

Other changes proposed was to remove adding custom roles, we will remove the link from UI. 
Move Resources permissions to Users.
When a user creates we should store that in the Applicaiton metadata as owner of that topology and they can give access to others by sharing the topology.

PR #575, opened 2017-05-16
Move kerberos login and renewal thread related code to common-auth in SR

priyank5485:
Remove code from kafka spout translator and streamline server to do
Subject.doAs for schema registry client. Rather SchemaRegistryClient
will handle the kerberos login part not. This change is dependent on
https://github.com/hortonworks/registry/pull/113/files

Change-Id: I12ed9df0e0e08014bce61055d0e1e6eb660f7277


harshach:
+1

PR #576, opened 2017-05-17
Quick Fix: Force using kafka-clients to 0.10.2.1 to use dynamic SASL jaas config

HeartSaVioR:
We're using dynamic SASL jaas config which is introduced at 0.10.2.0 for Kafka sources and sinks, but storm-kafka-client pulls kafka-clients 0.10.0.0 so the configuration is ignored in runtime.

This patch will fix the issue.


harshach:
+1

priyank5485:
+1

PR #577, opened 2017-05-17
STREAMLINE-789: Kafka Metadata should return broker hosts and all the…

hmcl:
… listeners

   - Added broker host and port information grouped by security protocol to the JSON security entry

### DIFF
--- a/streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/bundle/impl/KafkaBundleHintProvider.java
+++ b/streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/bundle/impl/KafkaBundleHintProvider.java
@@ -41,6 +41,7 @@ public class KafkaBundleHintProvider extends AbstractBundleHintProvider {
     public static final String FIELD_NAME_BROKER_ZK_PATH = "zkPath";
     public static final String FIELD_NAME_ZK_SERVERS = "zkServers";
     public static final String FIELD_NAME_ZK_PORT = "zkPort";
+    public static final String FIELD_NAME_BROKER_LISTENERS_PROTOCOL = "protocol";
 
     @Override
     public Map<String, Object> getHintsOnCluster(Cluster cluster) {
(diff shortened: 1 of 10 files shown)


hmcl:
@harshach @priyank5485 can you take a look in particular at KafkaBundleHintProviderTest.java and KafkaBundleHintProvider.java to make sure the hint info is being added properly. 

Currently the KafkaMetadataResource rest calls have the following output (i.e. protocol info is only added to the brokers json, but not to topics).
`
{
  "brokers": [
    {
      "host": "pm-eng1-cluster2.field.hortonworks.com",
      "port": 6667
    },
    {
      "host": "pm-eng1-cluster3.field.hortonworks.com",
      "port": 6667
    },
    {
      "host": "pm-eng1-cluster4.field.hortonworks.com",
      "port": 6667
    }
  ],
  "security": {
    "authentication": {
      "enabled": true,
      "scheme": "kerberos"
    },
    "authorizer": {
      "invoked": false,
      "msg": "Authorization not enforced. Every authenticated user has access to all metadata info"
    }
  },
  "protocol": {
    "SASL_PLAINTEXT": [
      "pm-eng1-cluster2.field.hortonworks.com:6667",
      "pm-eng1-cluster3.field.hortonworks.com:6667",
      "pm-eng1-cluster4.field.hortonworks.com:6667"
    ]
  }
}
`

`
{
  "topics": [
    "truck_test",
    "ambari_kafka_service_check",
    "truck_events",
    "truck_events_stream",
    "test",
    "__consumer_offsets"
  ],
  "security": {
    "authentication": {
      "enabled": true,
      "scheme": "kerberos"
    },
    "authorizer": {
      "invoked": false,
      "msg": "Authorization not enforced. Every authenticated user has access to all metadata info"
    }
  }
}

`

priyank5485:
@hmcl There are two hint provider classes KafkaSinkBundleHintProvider and KafkaBundleHintProvider for sink and source respectively. We need to handle bot source and sinks. Also, if I understand it correctly the way hints are returned is that per cluster its a json object. Each json object has a key which is the ui field name in the json spec for that component and then suggestions of values for that field name. After the latest change to switch to storm-kafka-client we have some new fields like bootstrapServers,  serviceName, etc. I was expecting to see the hint response to have keys for those fields and values as suggestions. Am i missing anything?

priyank5485:
@hmcl Just had a call with @shahsank3t Can we have the hint response for kafka source and sink have the following fields 
``` json
{
  "1": {
    "cluster": {
      "id": 1,
      "name": "clusterA",
      "ambariImportUrl": "http://172.22.125.40:8080/api/v1/clusters/clusterA",
      "description": "This is an auto generated description",
      "timestamp": 1495174176866
    },
    "hints": {
      "bootstrapServers": {
        "PLAINTEXT":
          "gk-sl-17jan-1.openstacklocal:6667"
      },
      "topic": [
        "ambari_kafka_service_check",
        "truck_events_stream"
      ],
      "securityProtocol": ["PLAINTEXT"],
      "kafkaServiceName": "kafka"
    }
  }
}
```

hmcl:
The changes in this PR were included as part of PR https://github.com/hortonworks/streamline/pull/590

PR #584, opened 2017-05-19
BUG-81055: Add null checks to handle null fields if directly creating via API

arunmahadevan:


arunmahadevan:
@satishd , QE is invoking the backend apis to create entities and not passing values for all the fields (so some fields are null). After that when they clone a topology its throwing NPE. So here I am treating copy ctor invoked with null as a no-op so that qe can run their tests with minimal fields.

PR #585, opened 2017-05-19
ISSUE-583: 'readerSchemaVersion' in Kafka config key is marked as opt…

### LINKED ISSUE #583: 'readerSchemaVersion' in Kafka config key is marked as optional but effectively required due to Flux
Kafka Source has 'readerSchemaVersion' as `optional` config key and its default value is set to `null`.
But since Flux can't handle `null` value and throws an error, the key is effectively `required`.

Needs to change `readerSchemaVersion` to be `required`, or add a new constructor for AvroKafkaSpoutTranslator which excludes `readerSchemaVersion` from parameters so that it can be handled as `null`.

Moreover, we need to handle `empty string` as `null` for `readerSchemaVersion` as well since it throws NumberFormatException.

HeartSaVioR:
…ional but effectively required due to Flux

* add a new constructor for AvroKafkaSpoutTranslator which excludes readerSchemaVersion from parameters
* handle empty string as null for readerSchemaVersion as well since it throws NumberFormatException

This closes #583

### DIFF
--- a/streams/runners/storm/layout/src/main/java/com/hortonworks/streamline/streams/layout/storm/KafkaSpoutFluxComponent.java
+++ b/streams/runners/storm/layout/src/main/java/com/hortonworks/streamline/streams/layout/storm/KafkaSpoutFluxComponent.java
@@ -144,7 +144,12 @@ public class KafkaSpoutFluxComponent extends AbstractFluxComponent {
                 .JSON_KEY_TOPIC}));
         constructorArgs.add(kafkaSource != null ? kafkaSource.getId() : "");
         constructorArgs.add(conf.get(TopologyLayoutConstants.SCHEMA_REGISTRY_URL));
-        constructorArgs.add(conf.get("readerSchemaVersion") != null ? Integer.parseInt((String) conf.get("readerSchemaVersion")) : null);
+
+        // add readerSchemaVersion to constructor arg only when it's not null and not empty
+        String readerSchemaVersion = (String) conf.get("readerSchemaVersion");
+        if(readerSchemaVersion != null && !readerSchemaVersion.isEMpty()) {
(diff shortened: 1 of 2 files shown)


streams/runners/storm/layout/src/main/java/com/hortonworks/streamline/streams/layout/storm/KafkaSpoutFluxComponent.java
```diff
@@ -144,7 +144,15 @@ private String addRecordTranslator () {
                 .JSON_KEY_TOPIC}));
         constructorArgs.add(kafkaSource != null ? kafkaSource.getId() : "");
         constructorArgs.add(conf.get(TopologyLayoutConstants.SCHEMA_REGISTRY_URL));
-        constructorArgs.add(conf.get("readerSchemaVersion") != null ? Integer.parseInt((String) conf.get("readerSchemaVersion")) : null);
+
+        // add readerSchemaVersion to constructor arg only when it's not null and not empty
+        Object readerSchemaVersionObj = conf.get("readerSchemaVersion");
+        if (readerSchemaVersionObj != null) {
+            String readerSchemaVersion = (String) readerSchemaVersionObj;
+            if (!readerSchemaVersion.isEmpty()) {
```
satishd: nit: This can be simplified like below without having multiple ifs
```
String readerSchemaVersion = (String) conf.get("readerSchemaVersion");
if(readerSchemaVersion != null && !readerSchemaVersion.isEMpty()) {
    constructorArgs.add(Integer.parseInt(readerSchemaVersion))
}
```

HeartSaVioR: Thanks for the guide. Addressed.

PR #587, opened 2017-05-19
BUG-81157: Add Keytab and Principal Information To Metadata Service Security JSON Object

hmcl:


harshach:
@hmcl file a github issue for this and use ISSUE-XXX in the commit log . Lets not BUG-XXX.

hmcl:
@harshach @priyank5485 this PR sits on top of https://github.com/hortonworks/streamline/pull/577 which should be merged first. I can take care of this once we have +1.

PR #596, opened 2017-05-21
ISSUE-595: StormMetadataResource endpoint returns 401 Unauthorized in…

hmcl:
… secured cluster setup

Change-Id: Ifab5788400e33080cd1e41b4b3101758efc155eb


HeartSaVioR:
Seems like we're putting similar efforts for both side, Metadata (its own handling) and other places (StormRestAPIClient).
May be better to use StormRestAPIClient also for StormMetadataResource, and put effort to StormRestAPIClient if it lacks some features.

priyank5485:
@hmcl This seems to be covered as part of the other PR. I already merged that. Hence closing this. you can reopen if thats not the case.

priyank5485:
@hmcl I cant close it. Can you close it?

PR #603, opened 2017-05-21
ISSUE-593, ISSUE-594, ISSUE-595

hmcl:
@harshach This PR has the 3 commits, one per issue. I rebased it on top of master, and did mvn clean install. All tests passed for me. I also tested with secure and insecure mode, as well as the hints. The only thing I didn't test was the StormHintProvider because I don't really know all the intricacies of bundling that hint provider with the UI. I believe @shahsank3t is supposed to do this. IF not let me know and I will look into it.

### DIFF
--- a/streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/service/metadata/StormMetadataService.java
+++ b/streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/service/metadata/StormMetadataService.java
@@ -25,6 +25,7 @@ import com.hortonworks.streamline.streams.catalog.exception.ServiceNotFoundExcep
 import com.hortonworks.streamline.streams.cluster.discovery.ambari.ComponentPropertyPattern;
 import com.hortonworks.streamline.streams.cluster.discovery.ambari.ServiceConfigurations;
 import com.hortonworks.streamline.streams.cluster.service.EnvironmentService;
+import com.hortonworks.streamline.streams.cluster.service.metadata.common.EnvironmentServiceUtil;
 import com.hortonworks.streamline.streams.cluster.service.metadata.common.HostPort;
 import com.hortonworks.streamline.streams.cluster.service.metadata.json.Authorizer;
 import com.hortonworks.streamline.streams.cluster.service.metadata.json.Keytabs;
@@ -40,6 +41,8 @@ import java.io.IOException;
 import java.security.PrivilegedActionException;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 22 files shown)


streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/service/metadata/StormMetadataService.java
```diff
@@ -153,45 +177,49 @@ private Client newHttpClient() {
...
                 }
-            }
-        }
-        return new StormTopologies(topologies, new Security(securityContext, new Authorizer(false), getPrincipals(), getKeytabs()));
+        );
+    }
+
+    private <T, E extends Exception> T executeSecure(SupplierException<T, E> action) throws PrivilegedActionException, E {
+        return action.get();    // For now it executes without security context. Uncomment bellow for security context.
+//        return SecurityUtil.execute(action, securityContext, subject);
```
priyank5485: @hmcl  Will this work for secured cluster as well? In the other PR just for this issue you had subject passed in and called SecurityUtil.execute. Just checking if you forgot to uncomment the line or not?

priyank5485:
@hmcl Just one question. Other than that it looks good. Will merge it in once you confirm no change is needed or push a fix

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/service/metadata/StormMetadataService.java
hmcl: @priyank5485 I left it like this because for my tests in secure (and insecure) mode it is the only way it works. With subject.doAs() oddly it does not work. Perhaps it is related with the comment that @HeartSaVioR made in [here](https://github.com/hortonworks/streamline/pull/596) as a follow up to his [patch](https://github.com/hortonworks/streamline/pull/598)


Is the code with the line currently commented out working for everyone? If so I can revet it to match this [change](https://github.com/hortonworks/streamline/pull/596), and will identify why it does not work for my tests.

priyank5485: @hmcl When you say does not work what is the issue?Anyway, if you are sure the way it is works in both modes i will merge it in

hmcl: I would say let's just merge it. If it does not work, the fix is to comment the line above and uncomment this line. I will try to find the stack trace. I have to re-run the tests.

Bottom line is, until the discussion that happened yesterday about things stopped working in insecure cluster, all the time things were as they are now (in this patch). So this as is is no different from what we have had the last 2/3 weeks..

priyank5485: @hmcl I can merge it in. Can you check one thing for me? @harshach had applied changes from one of the other PRs that you have referenced here in one of his secured clusters. Without that patch clusters/{clusterId}/services/storm/topologies" was returning Unauthorized response. With your patch it worked fine. Since you said it works in both secured and non secured clusters. Can you do a GET once in a secured setup to be sure that it works, unless you already tested that and confirm that it works.

harshach:
Deployed this PR under http://pm-eng1-cluster4.field.hortonworks.com:7777. 
@shahsank3t once your verify if the responses are good please update it here

shahsank3t:
verified the response on the cluster, looks good.

PR #605, opened 2017-05-22
ISSUE-604 Use DRIVER CLASS NAME in JDBC sink for as its name represents

HeartSaVioR:
* since some storages including Phoenix doesn't support javax.sql.DataSource
* modify all fields to conform to driverClassName (we no longer use datasource config)

Please refer https://github.com/brettwooldridge/HikariCP for field descriptions.


priyank5485:
+1

PR #615, opened 2017-05-23
ISSUE-607 JDBC sink should build UPSERT query instead of using JdbcIn…

### LINKED ISSUE #607: JDBC sink should build UPSERT query instead of using JdbcInsertBolt default INSERT query
Currently JDBC sink just uses JdbcInsertBolt and doesn't set custom query. This means the query for insertion would be `INSERT INTO <table> (<column list>) VALUES (<values>)`.

Given that Storm guarantees at-least-once, even there's no duplicate unique key for input tuples of JDBC sink, there could be same tuples received multiple time and it makes duplication of unique key. This means basic query only works with table which don't have unique keys (including primary key) which heavily limits functionality.

Moreover, Phoenix doesn't support non-PK table and support UPSERT instead of INSERT.

So JDBC sink should basically build UPSERT-like query, which format is up to the target DB type.
JDBC sink needs to have supported DB types, and lets users select correct one.

HeartSaVioR:
…sertBolt default INSERT query

* add 'dbType' to JDBC sink config field
* change underlying query to UPSERT instead of INSERT
* fix deadlock between DriverManager static block and specific driver class's static block

This closes #607.

### DIFF
--- a/bootstrap/components/sinks/jdbc-sink-topology-component.json
+++ b/bootstrap/components/sinks/jdbc-sink-topology-component.json
@@ -16,6 +16,18 @@
         "type": "string"
       },
       {
+        "uiName": "Database Type",
+        "fieldName": "dbType",
+        "isOptional": false,
+        "tooltip": "The type of target database.",
+        "type": "enumstring",
+        "options": [
+          "mysql",
+          "postgresql",
+          "phoenix"
+        ]
+      },
+      {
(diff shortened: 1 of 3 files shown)


harshach:
+1

HeartSaVioR:
Merged via 77dd77bd581a2b4f6d19a4d30cd121c25300c57a

PR #635, opened 2017-05-24
ISSUE-625: Service pool sorting by name should be case insensitive

### LINKED ISSUE #625: Service pool sorting by name should be case insensitive
Create a service pool name aaa and Zzz. When you sort it by name, Zzz appears first. Sort by name should be case insensitive.

arunmahadevan:
Fixes #625


harshach:
+1

shahsank3t:
+1 looks good, tested and verified

PR #689, opened 2017-05-29
ISSUE-688: Sum and Count UDAF does not work with nested fields

### LINKED ISSUE #688: Sum and Count UDAF does not work with nested fields
The bootstrap for Sum needs to be updated to use NumberSum which works for nested fields.

arunmahadevan:
Fixes #688


harshach:
+1

PR #712, opened 2017-06-01
ISSUE-710: Enhance custom processor interface to be able to use in di…

priyank5485:
…fferent contexts

Change-Id: I73704a41763dcc572366be9c6e87b941f6af0381


priyank5485:
@sureshms Please review

sureshms:
+1 with one change. Let's pick the field from custom processor output instead of input event field in case of name conflict

PR #723, opened 2017-06-03
BUG-82171 NullPointer exception when using PostgreSQL as backend storage

HeartSaVioR:
While we get the value from CHAR(5) which contains "true" (hence 1 space left), rightmost whitespace is trimmed for MySQL but unfortunately not for PostgreSQL.

Change-Id: Iac69fea23975949d5b1a08040878943078ae5a77


priyank5485:
+1

PR #729, opened 2017-06-05
Simple fix to Custom Processor

HeartSaVioR:
* console custom processor doesn't seem to conform new interface
* add verification for new field

@priyank5485 Please review this. Thanks in advance.


priyank5485:
+1

PR #730, opened 2017-06-05
ISSUE-728 pass nimbus.thrift.max_buffer_size from Environments/Storm during the deployment

### LINKED ISSUE #728: pass nimbus.thrift.max_buffer_size from Environments/Storm during the deployment
During the deployment of larger topologies we are running into issues with nimbus.thrift.max_buffer_size being low. This can't be solved by just updating on the cluster configs as this config is being used in NimbusClient , so on a machine where we only have Streamline installed , Ambari won't push this config to storm client hence we it depends on config from "defaults.yaml" which has a value 1MB. We should pick this from storm config and pass it during the deployment.

HeartSaVioR:
Closes #728 

@harshach Please review. Thanks in advance.


harshach:
+1

PR #760, opened 2017-06-26
ISSUE-751 Drop support Phoenix as Streamline storage type

### LINKED ISSUE #751: Drop support Phoenix as Streamline storage type
Streamline already relies on Relational DB characteristics, which Phoenix doesn't support. We've implemented features for Phoenix in Streamline, but the implementation requires additional operations hence slowing down whole operations. 

We're supporting MySQL and PostgreSQL officially, which would be sufficient for most of users. So we could just drop supporting Phoenix and clean up storage module.

HeartSaVioR:
Removed sql script for phoenix and phoenix support feature from storage module. Also some integration tests.

This closes #751


streamline-dist/VERSION
```diff
@@ -0,0 +1,6 @@
+#Created by build system. Do not modify
```
arunmahadevan: Should this be checked in?

satishd:
@arunmahadevan @HeartSaVioR  +1 for dropping support for phoenix and add transactions support in storage-manager abstractions which uses underlying storage engine's transactions.

@harshach It was proposed earlier that registry repo will have storage module and it should be shared by both streamline and registry. It would be good to have all these changes to go into a single place. I guess we earlier had a discussion on where to put these common modules like common-auth/storage-cache etc. We can decide and stick to that. 

I am not saying that these modules should be placed in registry. I am fine with any reasonable location of the common modules. I brought up this discussion here to avoid  any duplicate work to be done.

streamline-dist/VERSION
HeartSaVioR: I guess not. Will remove this.

HeartSaVioR: Removed

HeartSaVioR:
Raised #763

HeartSaVioR:
Btw, adding transaction to storage module will add complexity, so we might be better to have some POCs around some implementations simultaneously. JOOQ could be one of candidates, and if someone is familiar with any other things, maybe we could try.

harshach:
@satishd @HeartSaVioR the initial idea with the webserver modules we can co-locate under one jetty container and hence the tighter integration of registry & streamline.

We already have auth module in registry and we can maintain the storage the module there. Unless if we want to spin up another project for the common-modules. It will be confusing for everyone and any change would need to touch multiple projects. 
Given these are core capabilities , we can continue to have these under registry under common and move it out of streamline.

arunmahadevan:
@HeartSaVioR , to start with we can add support for beginTransaction, commit, rollback to Storage manager and start using it in the resource/service class methods. We could also consider frameworks like JOOQ or Spring but I think that would be a bigger refactoring.

HeartSaVioR:
Yes that would become a huge work, but I also think we added codes for making it just work for a long period. Would like to have some time to think about making it better.

HeartSaVioR:
Btw let's move the discussions to the proper places: 

- #763 regarding adding transaction (maybe also considering POC)
- new issue or discussion channel regarding having common module

HeartSaVioR:
@satishd I'll just merge this as it is. Whether we want to have common module or not, we want to drop supporting phoenix from both projects so this change should be checked in.

PR #766, opened 2017-06-27
ISSUE-761 Move standalone mode of Storm SQL to Streamline

### LINKED ISSUE #761: Move standalone mode of Storm SQL to Streamline
Storm SQL has standalone mode which is supposed to be used only Streamline. 
It forces Streamline to refer SNAPSHOT version of Storm, and also have possibility of Calcite version mismatch issue between Storm SQL and Streamline.
And since SQL feature is coupled with processor, which means that Streamline depends on Storm though users choose other stream framework as runner (when possible).
(Currently `streamline-runtime` depends on `storm-sql-core` and `storm-sql-runtime`.)

This issue tracks moving Storm SQL standalone mode to Streamline. It's OK to reimplement if it's simpler.

HeartSaVioR:
* most of the codes are from Apache Storm, hence keeping Apache license header there
* rename some files to conform to Streamline

Please note that we're completely removing `storm-sql-core` and `storm-sql-runtime` dependency, and also `storm-core` in `streamline-runtime` so that runner libraries can be isolated to the runner itself.

I'm not a lawyer so I'm not clear if it's OK to borrow some Storm codes (I guess this is OK) and modify a bit, and rename some classes. Heron has been doing it, but not sure they're modifying the borrowed code.

Maybe we can try reimplementing this with Calcite's feature like interpreter (and I think this is ideal), but it might need POC and some efforts, so leave it to the further improvement.

Another benefit of this patch for Storm side is that we can get rid of standalone mode in Storm SQL. I don't think someone uses it except Streamline.

Manually tested.

This closes #761.

---
This change is [Reviewable](https://reviewable.io/reviews/hortonworks/streamline/766)


harshach:
Lets merge it in after the 3.0.1 branch created. We can get this into master after.

HeartSaVioR:
Yes sure. That's a kind of improvement and not urgent unless Storm removes standard mode of Storm SQL or we start to support another streaming engines.

HeartSaVioR:
Given that we have HDF-3.0-maint, I'd like to see this reviewed and finally merged, so that I can poll around Storm user/dev mailing list or just craft a patch for removing standalone mode of Storm SQL.

harshach:
This PR is ready to merge. @arunmahadevan can you review

HeartSaVioR:
@arunmahadevan Thanks, I addressed review comments, find Storm from module and replace by hand.

arunmahadevan:
+1

PR #775, opened 2017-06-29
ISSUE-773: Ignore null values for a field in avro record to fix runti…

priyank5485:
…me exceptions

Change-Id: Ifb6cf88a2e0e6e777d6eb46226a212f13d2c6f5c

### DIFF
--- a/streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/spout/AvroStreamsSnapshotDeserializer.java
+++ b/streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/spout/AvroStreamsSnapshotDeserializer.java
@@ -65,7 +65,10 @@ public class AvroStreamsSnapshotDeserializer extends AvroSnapshotDeserializer {
             List<Schema.Field> fields = indexedRecord.getSchema().getFields();
             ImmutableMap.Builder<String, Object> keyValues = ImmutableMap.builder();
             for (Schema.Field field : fields) {
-                keyValues.put(field.name(), convertValue(indexedRecord.get(field.pos())));
+                Object currentValue = convertValue(indexedRecord.get(field.pos()));
+                if (currentValue != null) {
+                    keyValues.put(field.name(), currentValue);
+                }
             }
             value = keyValues.build();
 
@@ -86,7 +89,10 @@ public class AvroStreamsSnapshotDeserializer extends AvroSnapshotDeserializer {
             Map<Object, Object> map = (Map<Object, Object>) deserializedObj;
             ImmutableMap.Builder<String, Object> keyValues = ImmutableMap.builder();
             for (Map.Entry entry : map.entrySet()) {
-                keyValues.put(entry.getKey().toString(), convertValue(entry.getValue()));
+                Object currentValue = convertValue(entry.getValue());
+                if (currentValue != null) {
+                    keyValues.put(entry.getKey().toString(), currentValue);
+                }
             }
             value = keyValues.build();
 
@@ -94,7 +100,10 @@ public class AvroStreamsSnapshotDeserializer extends AvroSnapshotDeserializer {
             Collection<Object> collection = (Collection<Object>) deserializedObj;
             ImmutableList.Builder<Object> values = ImmutableList.builder();
             for (Object obj : collection) {
-                values.add(convertValue(obj));
+                Object currentValue = convertValue(obj);
+                if (currentValue != null) {
+                    values.add(currentValue);
+                }
             }
             value = values.build();
 


streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/spout/AvroStreamsSnapshotDeserializer.java
```diff
@@ -65,7 +65,10 @@ private Object convertValue(Object deserializedObj) {
             List<Schema.Field> fields = indexedRecord.getSchema().getFields();
             ImmutableMap.Builder<String, Object> keyValues = ImmutableMap.builder();
             for (Schema.Field field : fields) {
-                keyValues.put(field.name(), convertValue(indexedRecord.get(field.pos())));
+                Object currentValue = convertValue(indexedRecord.get(field.pos()));
+                if (currentValue != null) {
```
harshach: can you add debug log level statement saying ignoring null message with offset from kafka

satishd: We should not ignore null values for a field but the map can be changed to `HashMap` and set the records `value` as unmodifiable map of the earlier HashMap.

streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/spout/AvroStreamsSnapshotDeserializer.java
```diff
@@ -86,15 +89,21 @@ private Object convertValue(Object deserializedObj) {
             Map<Object, Object> map = (Map<Object, Object>) deserializedObj;
             ImmutableMap.Builder<String, Object> keyValues = ImmutableMap.builder();
             for (Map.Entry entry : map.entrySet()) {
-                keyValues.put(entry.getKey().toString(), convertValue(entry.getValue()));
+                Object currentValue = convertValue(entry.getValue());
```
satishd: same as earlier comment.

streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/spout/AvroStreamsSnapshotDeserializer.java
```diff
@@ -86,15 +89,21 @@ private Object convertValue(Object deserializedObj) {
...
             }
             value = keyValues.build();
 
         } else if (deserializedObj instanceof Collection) { // type of array
             Collection<Object> collection = (Collection<Object>) deserializedObj;
             ImmutableList.Builder<Object> values = ImmutableList.builder();
             for (Object obj : collection) {
-                values.add(convertValue(obj));
+                Object currentValue = convertValue(obj);
+                if (currentValue != null) {
```
satishd: We should not ignore null values for a field but the `values` can be changed to `ArrayList` and set the records value as unmodifiable list of the earlier ArrayList.

harshach:
@priyank5485 @satishd should we make this as configurable option and default to true. 
ignore.invalid.messages on all the sources side and also extend this to unparsable event. 
I think for the most users they might want to ignore incase of mal-formed or null events but for some it might important to throw an error an stop processing.  In case of mal-formed event we should also log that.

satishd:
@harshach I guess you intended to put earlier comment for https://github.com/hortonworks/registry/pull/168

streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/spout/AvroStreamsSnapshotDeserializer.java
priyank5485: @harshach Its not null message from kafka at this point. Its a nullable field from avro schema inside the message. We dont have kafka offset at this point in code. This is called after the deserializer.deserialize on kafka has been called for value in kafka payload. That deserialized value is only passed here. If we want to log offset, i will have to change around signatures of methods to pass offset information as well and log it. We can log that we did not receive a value for nullable avro field here. But I think we can probably not do that as well since user should be aware of avro schema. Do you still prefer adding any debug log here?

priyank5485: @satishd Any reason for not ignoring null values for a field? It is after all a nullable field as per avro schema. Can you elaborate on your suggested solution? I did not understand it.

priyank5485:
@harshach Agree with you on having a setting like that. I remember discussing this with @satishd His suggestion was to add a checked exception to avro deserializers in SR. That way SAM source using SR deserializer can have a catch block for the checked exception and based on the config you suggested we can ignore the message totally and log it. However if ignore.invalid.messages was set to false, not sure how to handle that. Need to think more about that.

streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/spout/AvroStreamsSnapshotDeserializer.java
satishd: @priyank5485 In this fix, null value for a key is ignored and that key does not show up in the Map. 
We should always generate a map with all the fields that are received as part of the payload even when their values are null. The reason is that a key containing null value is not equivalent to not having key at all in the payload.

satishd:
@harshach @priyank5485 It may be better to extend this to earlier concept of invalid/unparsed messages stream at source level in which users can get complete control of handling invalid messages. 

@priyank5485 My suggestion was to catch `SerDesException` thrown by deserializer in spouts and take respective action. Deserializer can not catch and log as it has to return some value back to the caller of `deserialize` method.

streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/spout/AvroStreamsSnapshotDeserializer.java
priyank5485: @satishd I understand that a key containing null value is not equivalent to not having key at all. But how does that matter here? Is it going to break anything? Can you give an example? We need this map to be immutable since its going to be sent across in storm. I am still not clear on exact solution you are suggesting? Is there an immutable map that allows null values?

priyank5485:
@satishd Having invalid/unparsed stream sounds like a possible approach. If we decide to do that, it will need more changes and probably some UI changes as well. We can cover it in another issue if both of you are okay. Also, now i remember your suggestion. Just curious cant we have a catch all block in SR deserializer and throw it as a checked exception that is a part of deserialize method? Why does deserialize have to return something?

streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/spout/AvroStreamsSnapshotDeserializer.java
satishd: @priyank5485 User may have logic in UDFs/UDAFs or elsewhere having a check like map contains key which returns false though it should returned true, it is going to break that logic. We should not be filtering any of the received payload unless there is a strong reason todo. Here, I do not see any reason for filtering and send the received payload as it is by source to subsequent processors.

 with this change we are converting the received payload 
```
{
"id": 12322,
"name": "foo",
location: null
}
```
into 
```
{
"id": 12322,
"name":"foo",
}
```
which are not equal.

When avro schema is defined as 
```
{
 "namespace":"hwx",
 "type":"record",
 "name":"",
 "fields":[
    ...
    { "name": "location", "type": [ "string", "null" ] },
    ...
 ]
}
```
then `value` field must exist either as some string value or as null but it should not be missing. With this fix, it would break that constraint and result in error.

>We need this map to be immutable since its going to be sent across in storm. I am still not clear on exact solution you are suggesting? Is there an immutable map that allows null values?

I have mentioned in my earlier [comment](https://github.com/hortonworks/streamline/pull/775#discussion_r125385744) about creating HashMap and return unmodifiable map for immutability which can be done like below.
```
        Map<String, Object> fieldValues = new HashMap<>();
        Object values = convertValue(deserializedObj);
        if (values instanceof Map) {
            fieldValues.putAll((Map) values);
        } else {
            fieldValues.put(StreamlineEvent.PRIMITIVE_PAYLOAD_FIELD, values);
        }

        return Collections.unmodifiableMap(fieldValues);
```

Issue encountered by user:
there was a null value sent by client as part of some field which is defined in schema as it can be either null or non-null value and the current code  is trying to set that but it is throwing NPE because of using `ImmutableMap`. 
So, we should fix that instead of filtering out those valid null values, one way to fix that is as above.

satishd:
>Just curious cant we have a catch all block in SR deserializer and throw it as a checked exception that is a part of deserialize method? Why does deserialize have to return something?

@priyank5485 What I meant is deserializer can not just ignore invalid messages by catching exceptions and log them without throwing exception or returning payloads back to consumer of deserializer. Consumers of deserializer need to handle the logic of handling errors and take respective action. 

I am fine with addressing invalid messages at source with a new issue.

streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/spout/AvroStreamsSnapshotDeserializer.java
harshach: @priyank5485 got it. We don't need to add that.

priyank5485: @satishd Sorry, i did not see your last comments until now. For some reason, I am not always getting notifications.  I read your explanation about avro record with null value. It makes sense. However, here we are not dealing with avro record directly, anywhere. Either in the code that we write in SAM processors/sinks or in custom UDFs, UDAFs, CPs, etc. Ultimately, user understands that an avro record is translated to a StreamlineEvent. How we define the behavior of null avro fields in StreamlineEvent is our decision. One way of doing is putting a null value. Other is not putting an entry at all for that field.  As far as your point about map contains key returning false is concerned, do you have a use case in mind where user would want it to return true? It seems that you are saying user will rely on getting all field names in schema by doing a map.keys() or something on those lines. I dont think they should need to do that from the message payload. They should already be aware of the schema and not rely on the map to get the fields in their schema.

priyank5485:
@satishd Raised this https://github.com/hortonworks/streamline/issues/808 for invalid or messages that dont adhere to schema

streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/spout/AvroStreamsSnapshotDeserializer.java
harshach: @priyank5485 if I understand correctly what @satishd is saying is we might have a schema where the filed might look like this
`driverId: int, null`
With the above field we can have a event with driverId = 101 and another event where driverId = null.
It looks like you are removing the driverId from streamlineEvent all together if its null. So in this case if we have a rules processor where user added a rule that says "driverId > 100". This could cause an issue as the user assumes there will be a filed called driverId in the event where as with this fix we are removing when its null. So a Map lookup for the driverId might not be successful when its null. 
I think it makes sense to keep the driverId even if its null.  Basically we are not being true to the schema thats expressed and can have issues since we are not meeting the expectation that the StreamlineEvent will contain all the keys expressed in a schema.
This also means users need to worry about null values coming in. If they wrote a generic custom processor they have to make sure the field that they are looking is not null before doing any operations on it.

priyank5485: @harshach If we want StreamlineEvent to contain all keys(even for null values) then we have to go with the route suggested by Satish. @satishd I have code changes like you suggested but we will run in to issue further down. StreamlineEventImpl uses ImmutableMap and when an instance of StreamlineEventImpl is constructed it will try to create an ImmutableMap again with an entry that has a null for its value. https://github.com/hortonworks/streamline/blob/master/streams/common/src/main/java/com/hortonworks/streamline/streams/common/StreamlineEventImpl.java#L105 is the line where it will throw an Exception now. I will need to change StreamlineEventImpl as well to not use ImmutableMap since it does not allow null values and it will need more testing.

arunmahadevan: @priyank5485 , if you are changing the implementation you also need to make sure to re-implement all the methods that could potentially modify the map (like the entrySet etc). Also the Immutablemap.CopyOf does a reference copy if the map is already immutable, and these kind of optimization is really needed in our transformations where we always emit a new event. Allowing null values opens up possiblility of NPE so thats one of the reasons why the guava immutable map disallows it. Another option would be to insert something like `Optional.empty()` instead of null value and have it well defined in our contract/javadocs. The receivers can check for `value == Optional.empty()` than `value == null` wherever they want null semantics.

priyank5485: @arunmahadevan All of what you said makes sense. Because changing StreamlineEventImpl is not a trivial change, I will wait for @satishd and @harshach to see if they are okay with using Optional.empty() or if they have any other ideas

harshach: @priyank5485 @satishd @arunmahadevan changing this for minor release seems to be risky, especially changing the immutable map.  Lets file another issue for next major release to keep track of it.
Overall I think @arunmahadevan suggestion looks good to use `Optional.empty()`. I don't think we've clear answer how better to deal with it. On one hand removing the key when value is null breaks the schema contract but having them in there means Users in their code needs to handle this. 
Especially incase of generic UDFs/CPs they have to write the code defensively as they don't know which schema's field might be null.

satishd: @harshach Agree that it is risky to have big changes now for this release. 

@priyank5485 @harshach @arunmahadevan Having Map abstraction exposed with StreamlineEvent is a leaky abstraction and we had discussion on that earlier. Ideally, internal implementation should not dictate the user API. Having Optional.empty() looks fine even though it adds burden on user to have this check instead of null. This would have been seamless if we have StreamlineEvent as Map<String, Optional> (with values always wrapped in Optional) instead of Map<String, Object>.

harshach: @satishd @priyank5485 @arunmahadevan lets look at this for next release. We might have to re-think the StreamlineEvent but changing that means we are going to break the backward compatibility. 
@priyank5485 Lets go with Optional.empty() option for this release.

harshach:
discussed with @priyank5485 . Since the current change is already running in a customer env and proven to be working. Adding Optional.empty() could provide additional issues like here https://github.com/hortonworks/streamline/blob/master/streams/functions/src/main/java/com/hortonworks/streamline/streams/udaf/NumberSum.java#L41 . We need to check across the codebase to make sure we ignore Optional.empty(). 
Lets go with deleting the key for this release. Lets open a Issue to make sure we make changes at the core and thought through different scenarios where this approach might provide issues.

I am +1 on merging this for the current release.

streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/spout/AvroStreamsSnapshotDeserializer.java
priyank5485: @arunmahadevan @satishd @harshach I spoke too soon. Having thought more about it, i feel that Optional.empty might introduce some regression. Regression would be from other processors like Aggregate, Join, Rule, etc. Some of the them use storm sql. Behaviour of storm sql or other processors is more likely to break anything if we use Optional.empty rather than what we currently have. Moreover, the current PR has also been tested in a SAM application that first brought this issue to light. I think for this release, I prefer the way it is in this PR - i.e. no entry for nullable fields in StreamlineEvent. Not disagreeing with any of the discussion here. And it could still cause some issues but better than Optional.empty. We need more thinking through and testing to make figure out the contract for nullable fields from avro schema flowing in SAM application

priyank5485:
Merged this in to HDF-3.0-maint. Not merging in to master. Will raise a new issue to track avro null fields in StreamlineEvent

priyank5485:
Raised this https://github.com/hortonworks/streamline/issues/821

PR #784, opened 2017-07-04
ISSUE-783: Putting a special character like question mark in any processor name causes exception in App Deployment

### LINKED ISSUE #783: Putting a special character like question mark in any processor name causes exception in App Deployment

arunmahadevan:
Modified StormRestAPIClient to encode parts of the url.

Fixes #783

### DIFF
--- a/streams/runners/storm/common/src/main/java/com/hortonworks/streamline/streams/storm/common/StormRestAPIClient.java
+++ b/streams/runners/storm/common/src/main/java/com/hortonworks/streamline/streams/storm/common/StormRestAPIClient.java
@@ -30,6 +30,8 @@ import java.io.IOException;
 import java.security.PrivilegedAction;
 import java.util.Map;
 
+import static com.hortonworks.streamline.common.util.WSUtils.encode;
+
 public class StormRestAPIClient {
     private static final Logger LOG = LoggerFactory.getLogger(StormRestAPIClient.class);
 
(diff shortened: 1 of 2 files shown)


HeartSaVioR:
+1

streams/runners/storm/common/src/main/java/com/hortonworks/streamline/streams/storm/common/StormRestAPIClient.java
```diff
@@ -45,29 +47,29 @@ public StormRestAPIClient(Client client, String stormApiRootUrl, Subject subject
     }
 
     public Map getTopologySummary(String asUser) {
-        return doGetRequest(getTopologySummaryUrl(asUser));
+        return doGetRequest(getTopologySummaryUrl(encode(asUser)));
     }
 
     public Map getTopology(String topologyId, String asUser) {
-        return doGetRequest(getTopologyUrl(topologyId, asUser));
+        return doGetRequest(getTopologyUrl(encode(topologyId), encode(asUser)));
```
satishd: nit: May want to push encode into respective methods instead of passing encoded strings to that.
Actually most of these methods generate string of the form like below. This can be extracted to one method which does encoding of topology-id and as-user by passing the given operation.

`stormApiRootUrl + "/topology/" + topologyId + <operation> + <as-user>`

New method can be 
```
    private String generateUrl(String topologyId, String asUser, String operation) {
        String baseUrl = stormApiRootUrl + "/topology/" + encode(topologyId) + "/" + operation;
        if (StringUtils.isNotEmpty(asUser)) {
            baseUrl += "?doAsUser=" + encode(asUser);
        }
        return baseUrl;
    }
```
activation url can be ` generateUrl(topologyId, asUser, "activate")`  which can be used inside activateUrl by removing getTOpologyxxxxUrl methods.

arunmahadevan: it will involve refactoring the most of the code since there are other methods like `getComponentUrl` `getTopologySummaryUrl` etc which doesnt follow the above pattern and all parts of the url need not be encoded. For now my preference is to encode in the public methods.

I also moved the encode to WsUtils so it can be used in other places if needed.

satishd: OK, this can be taken in another PR.

PR #789, opened 2017-07-06
ISSUE-745: Added logout REST endpoint

### LINKED ISSUE #745: Add logout link to invalidate the kerberos auth cookie

arunmahadevan:
Fixes #745 

@shahsank3t, sending a POST to "api/v1/catalog/users/current/logout" will invalidate the cookie. Checked in safari and via cUrl.

@harshach, we need to configure ambari to set `cookie.path: "/"` under servletFilters config in the generated streamline.yaml to always set cookie for "/".

### DIFF
--- a/conf/streamline-dev.yaml
+++ b/conf/streamline-dev.yaml
@@ -98,3 +98,5 @@ logging:
 #     kerberos.principal: "HTTP/streamline-ui-host.com"
 #     kerberos.keytab: "/vagrant/keytabs/http.keytab"
 #     kerberos.name.rules: "RULE:[2:$1@$0]([jt]t@.*EXAMPLE.COM)s/.*/$MAPRED_USER/ RULE:[2:$1@$0]([nd]n@.*EXAMPLE.COM)s/.*/$HDFS_USER/DEFAULT"
+#     token.validity: "36000"
+#     cookie.path: "/"
--- a/conf/streamline.mysql.yaml
+++ b/conf/streamline.mysql.yaml
@@ -97,3 +97,5 @@ logging:
 #     kerberos.principal: "HTTP/streamline-ui-host.com"
 #     kerberos.keytab: "/vagrant/keytabs/http.keytab"
 #     kerberos.name.rules: "RULE:[2:$1@$0]([jt]t@.*EXAMPLE.COM)s/.*/$MAPRED_USER/ RULE:[2:$1@$0]([nd]n@.*EXAMPLE.COM)s/.*/$HDFS_USER/DEFAULT"
+#     token.validity: "36000"
+#     cookie.path: "/"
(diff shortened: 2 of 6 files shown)


priyank5485:
+1

priyank5485:
@arunmahadevan Having thought more on this, do we want the requirement of / as cookie.path? Most user setups wont have permissions on that directory. I think we should avoid making that parameter mandatory and let client(browser, curl, etc) decide the cookie path. If cookie.path config is used indeed then we can have a reference to the filter and use it to getCookiePath instead of hardcoding. I prefer not using cookie.path at all.

arunmahadevan:
@priyank5485 , I am not sure if I understand what you meant by user not having permission on the path, but `cookie.path` is the URI stored within the cookie and not the file system path. Without this browser uses the URI visited by the user as the path and stores multiple cookie instances (one per URI) which is undesirable since this is a login cookie and it cannot be easily cleared with a Set-Cookie.

priyank5485:
@arunmahadevan I misunderstood the use of `cookie.path` When you say browser stores multiple cookies, are you saying that each time user clicks on a link in UI that goes to a different UI the entire SPNEGO sequence happens again creating a new cookie for that URI? From what i remember cookie is set by server only the first time, whatever URI that may be. By default `cookie.path` is null so there is no path set in the first cookie by server. And the browser keeps sending the same cookie back in any subsequent requests. Let me know if i am mis-understanding anything here.

arunmahadevan:
@priyank5485 , while browsing the UI the path mostly doesn't change. However while testing by directly hitting the API endpoints like "/api/v1/catalog/users/current" etc via browser, I observed multiple instances of cookies stored in the browser. It may depend but its better to set the path explicitly.

shahsank3t:
+1 
Tested API on a secure cluster and works as expected.

priyank5485:
@arunmahadevan Just to close the loop here. Last time we discussed i think you were not seeing multiple instances of cookies. Can you double check that? If its not the case and if we can delete the cookies then we can do away with an extra ambari config. Otherwise, we can take this approach

arunmahadevan:
@priyank5485 , I am seeing multiple instances of the cookies. See attached screenshot.

The cookie [specification](https://tools.ietf.org/html/rfc6265#section-4.1.2.4) clearly says "If the server omits the Path attribute, the user
   agent will use the "directory" of the request-uri's path component as the default value.", so its required to set the "path=/" attribute for the authentication cookies.

@harshach we need to configure ambari to set `cookie.path: "/"` under servletFilters config in the generated streamline.yaml to always set cookie for "/".

screen shot 2017-07-11 at 1 03 25 pm

priyank5485:
@arunmahadevan In that case, it seems that we dont want the path to be configurable and always want it to be "/". Then we can just hard code it in streamline.yaml instead of adding it to ambari?

arunmahadevan:
@priyank5485 , yes it can be hardcoded in the streamline.yaml generated by ambari and need not be configurable. So I presume we still need to update the config/scripts that ambari uses to generate streamline.yaml.

priyank5485:
@arunmahadevan Sounds good. I will take care of ambari config. I will merge this in.

PR #792, opened 2017-07-06
ISSUE-785 Revisit manual service registerer

### LINKED ISSUE #785: Revisit manual service registerer
This issue addresses #780 and #781 

Below content is copied from #780 

This is a bit complicated because of Ambari import. Ambari provides <protocol>://localhost:<port> so that was fairly easy, but not similar as above configuration.

Thought about manual registration more, and I also found what I was missing: configuration file doesn't have default value hence and we can't get the default value anyway.

We need to define required fields for service and let user also input on registering. Since we add resource file for Hadoop / HBase / Hive in topology, we may not need to define required fields for those services, but need to define for others.

Another issue is current implementation of Component. It contains host information as hosts, port, protocol, which can't handle above case. (different protocol and/or different port)

There're two ways of handle this case:
* allow multiple Components for same component (for example, NIMBUS)
  * hosts could be changed to host
* introduce a new entity for storing ComponentHost
  * move hosts, port, protocol to this entity
  * hosts could be changed to host

> Storm

* required
  * nimbus.seeds / nimbus.port => Nimbus component
  * ui.host / ui.port => UI component
* optional
  * nimbus.thrift.max_buffer_size
  * storm.thrift.transport
  * storm.principal.tolocal

> Kafka

* required
  * zookeeper.connect
  * listeners -> Broker component
* optional
  * security.inter.broker.protocol

> Druid

* required
  * druid.zk.service.host
  * druid.selectors.indexing.serviceName
  * druid.discovery.curator.path

HeartSaVioR:
This closes #785 

* Define required and optional fields instead of uploading config file
  * we missed that config file doesn't have all fields and streamline can't infer default value
* Introduce ComponentProcess to break down Component and Process
  * hosts / port / protocol are moved from Component to ComponentProcess
  * WARN: table schema changed

@hmcl 
I modified lots of metadata codes because of ComponentProcess. 
Could you help me reviewing the code and see if I'm breaking anything? Thanks in advance.

---
This change is [Reviewable](https://reviewable.io/reviews/hortonworks/streamline/792)


HeartSaVioR:
Manually tested with unsecured cluster: both manual / imported cluster with Kafka only topology.

HeartSaVioR:
This patch also fixes test failing on RestIntegrationTest.testAllResources. I disabled topology REST API test because it now depends on corresponding complicated entities (Namespace, Service, Component, ComponentProcess, ServiceConfiguration, Mapping).

HeartSaVioR:
This is quite a huge patch but I would be really happy if someone can take a stab with reviewing. I'll address conflicts after passing review process.

harshach:
@hmcl can you review this.

streams/cluster/src/test/java/com/hortonworks/streamline/streams/cluster/register/impl/KafkaServiceRegistrarTest.java
```diff
@@ -51,47 +54,70 @@ public void testRegister() throws Exception {
...
+        registrar.register(cluster, config, Collections.emptyList());
 
         Service kafkaService = environmentService.getServiceByName(cluster.getId(), Constants.Kafka.SERVICE_NAME);
         assertNotNull(kafkaService);
+
+        Component broker = environmentService.getComponentByName(kafkaService.getId(), ComponentPropertyPattern.KAFKA_BROKER.name());
+        assertNotNull(broker);
+
+        Collection<ComponentProcess> brokerProcesses = environmentService.listComponentProcessesInComponent(broker.getId());
+        assertEquals(4, brokerProcesses.size());
```
hmcl: why 4? Is it NUM_LISTENERS?  can you put a constant describing what this number is?

streams/cluster/src/test/java/com/hortonworks/streamline/streams/cluster/register/impl/StormServiceRegistrarTest.java
```diff
@@ -57,38 +61,62 @@ public void testRegister() throws Exception {
...
 
         Service stormService = environmentService.getServiceByName(cluster.getId(), Constants.Storm.SERVICE_NAME);
         assertNotNull(stormService);
+
+        Component nimbus = environmentService.getComponentByName(stormService.getId(), ComponentPropertyPattern.NIMBUS.name());
+        assertNotNull(nimbus);
+
+        Collection<ComponentProcess> nimbusProcesses = environmentService.listComponentProcessesInComponent(nimbus.getId());
+        List<String> hosts = nimbusProcesses.stream().map(ComponentProcess::getHost).collect(Collectors.toList());
+        assertEquals(Sets.newHashSet("storm-1", "storm-2"), new HashSet<>(hosts));
```
hmcl: Is this always true, or does it depend on the order of the sets? If so it could cause sporadic test failures.

streams/cluster/src/test/java/com/hortonworks/streamline/streams/cluster/register/impl/StormServiceRegistrarTest.java
```diff
@@ -57,38 +61,62 @@ public void testRegister() throws Exception {
...
+        List<Integer> ports = nimbusProcesses.stream().map(ComponentProcess::getPort).collect(Collectors.toList());
+        assertEquals(Sets.newHashSet(6627, 6627), new HashSet<>(ports));
```
hmcl: Is this always true, or does it depend on the order of the sets? If so it could cause sporadic test failures.

streams/cluster/src/test/java/com/hortonworks/streamline/streams/cluster/register/impl/KafkaServiceRegistrarTest.java
```diff
@@ -101,15 +127,17 @@ public void testRegister_component_kafka_broker_notPresent() throws Exception {
     }
 
     @Test
```
hmcl: NIT: you could use an annotation saying that this test is expected to throw IAE

streams/cluster/src/test/java/com/hortonworks/streamline/streams/cluster/register/impl/KafkaServiceRegistrarTest.java
```diff
@@ -51,47 +54,70 @@ public void testRegister() throws Exception {
...
+        assertEquals(1, brokerProcesses.size());
+        assertTrue(brokerProcesses.stream().anyMatch(p -> p.getHost().equals("kafka-1") && p.getPort().equals(9092) && p.getProtocol().equals("PLAINTEXT")));
+
+        ServiceConfiguration serverPropertiesConf = environmentService.getServiceConfigurationByName(kafkaService.getId(), CONFIGURATION_NAME_SERVER_PROPERTIES);
+        assertNotNull(serverPropertiesConf);
+        Map<String, String> serverPropertiesConfMap = serverPropertiesConf.getConfigurationMap();
+        assertFalse(serverPropertiesConfMap.containsKey(KafkaServiceRegistrar.PARAM_SECURITY_INTER_BROKER_PROTOCOL));
     }
 
     @Test
```
hmcl: NIT: you could use an annotation saying that this test is expected to throw IAE

streams/cluster/src/test/java/com/hortonworks/streamline/streams/cluster/register/impl/StormServiceRegistrarTest.java
```diff
@@ -97,27 +125,10 @@ public void testRegisterWithoutNimbusPrincipal() throws Exception {
...
-            Service stormService = environmentService.getServiceByName(cluster.getId(), Constants.Storm.SERVICE_NAME);
-            assertNull(stormService);
-        }
+        assertFalse(confMap.containsKey(StormServiceRegistrar.PARAM_NIMBUS_THRIFT_MAX_BUFFER_SIZE));
+        assertFalse(confMap.containsKey(StormServiceRegistrar.PARAM_PRINCIPAL_TO_LOCAL));
+        assertFalse(confMap.containsKey(StormServiceRegistrar.PARAM_THRIFT_TRANSPORT));
+        assertFalse(confMap.containsKey(StormServiceRegistrar.PARAM_NIMBUS_PRINCIPAL_NAME));
     }
 
     @Test
```
hmcl: NIT: you could use an annotation saying that this test is expected to throw IAE

streams/cluster/src/test/java/com/hortonworks/streamline/streams/cluster/register/impl/StormServiceRegistrarTest.java
```diff
@@ -145,11 +157,12 @@ public void testRegister_component_nimbus_notPresent() throws Exception {
 
         StormServiceRegistrar registrar = initializeServiceRegistrar();
 
-        try (InputStream is = getClass().getClassLoader().getResourceAsStream(STORM_YAML_FILE_PATH)) {
+        try {
```
hmcl: NIT: you could use an annotation saying that this test is expected to throw IAE

streams/cluster/src/test/java/com/hortonworks/streamline/streams/cluster/register/impl/ZookeeperServiceRegistrarTest.java
```diff
@@ -50,71 +55,40 @@ public void testRegister() throws Exception {
...
-            registrar.register(cluster, config, Lists.newArrayList(zooCfg));
-            fail("Should throw IllegalArgumentException");
-        } catch (IllegalArgumentException e) {
-            // OK
-            Service zkService = environmentService.getServiceByName(cluster.getId(), Constants.Zookeeper.SERVICE_NAME);
-            assertNull(zkService);
-        }
-    }
+        Collection<ComponentProcess> zkServerProcesses = environmentService.listComponentProcessesInComponent(zkServer.getId());
+        assertEquals(Sets.newHashSet("zookeeper-1", "zookeeper-2"),
```
hmcl: Is this always true, or does it depend on the order of the sets? If so it could cause sporadic test failures.

streams/cluster/src/test/java/com/hortonworks/streamline/streams/cluster/service/metadata/KafkaMetadataServiceTest.java
```diff
@@ -118,30 +129,28 @@ public void test_KafkaZkConnection_createPath() throws Exception {
...
+
+        List<ComponentProcess> brokerProcesses = new ArrayList<>();
+        brokerProcesses.add(broker1);
+        brokerProcesses.add(broker2);
+        return brokerProcesses;
     }
 
     @Test
     public void test_getBrokerInfoFromZk() throws Exception {
+        List<ComponentProcess> brokerProcesses = getBrokerComponentProcesses();
```
hmcl: redundant var

streams/cluster/src/test/java/com/hortonworks/streamline/streams/cluster/service/metadata/StormMetadataServiceTest.java
```diff
@@ -85,6 +92,12 @@ public void getTopologies() throws Exception {
             securityContext.getAuthenticationScheme(); result = AUTHENTICATION_SCHEME_NOT_KERBEROS;
         }};
 
+        List<ComponentProcess> nimbusProcesses = buildNimbusComponentProcesses();
+        List<ComponentProcess> uiProcesses = buildUiComponentProcesses();
```
hmcl: redundant vars

bootstrap/services/storm-bundle.json
```diff
@@ -4,33 +4,62 @@
   "serviceUISpecification": {
     "fields": [
       {
+        "uiName": "Nimbuses hostnames",
```
HeartSaVioR: Storm uses `Nimbuses` and `Nimbodes` to represent plural form of Nimbus.
I'm not clear which word Nathan referred, but at least one of word `nimbus` has `nimbuses` as plural form.
https://www.merriam-webster.com/dictionary/nimbus

bootstrap/services/storm-bundle.json
```diff
@@ -4,33 +4,62 @@
...
-        "uiName": "Storm config file (storm.yaml)",
-        "fieldName": "storm.yaml",
-        "hint" : "yaml",
-        "isOptional": false,
-        "tooltip": "Upload storm.yaml file used by Storm cluster",
-        "type": "file"
+        "uiName": "Principal to local plug-in",
+        "fieldName": "storm.principal.tolocal",
+        "isOptional": true,
+        "tooltip": "The plugin that will convert a principal to a local user",
```
HeartSaVioR: I just copied explanation from current Storm Config doc.
https://github.com/apache/storm/blob/master/storm-client/src/jvm/org/apache/storm/Config.java#L1267-L1271

If we can find better description, we can apply it to both side.

bootstrap/services/storm-bundle.json
```diff
@@ -4,33 +4,62 @@
...
         "type": "string"
       },
       {
-        "uiName": "Storm config file (storm.yaml)",
-        "fieldName": "storm.yaml",
-        "hint" : "yaml",
-        "isOptional": false,
-        "tooltip": "Upload storm.yaml file used by Storm cluster",
-        "type": "file"
+        "uiName": "Principal to local plug-in",
```
HeartSaVioR: Storm clearly states that it is a plug-in to convert principal to local user.
https://github.com/apache/storm/blob/master/storm-client/src/jvm/org/apache/storm/Config.java#L1267-L1271

If we can find better description, we can apply it to both side.

bootstrap/services/storm-bundle.json
```diff
@@ -4,33 +4,62 @@
...
+        "fieldName": "storm.principal.tolocal",
+        "isOptional": true,
+        "tooltip": "The plugin that will convert a principal to a local user",
+        "type": "string"
+      },
+      {
+        "uiName": "Nimbus Principal",
```
HeartSaVioR: Kerberos looks better.

streams/actions/src/main/java/com/hortonworks/streamline/streams/actions/container/TopologyActionsContainer.java
```diff
@@ -101,15 +104,30 @@ private TopologyActions initTopologyActions(Map<String, Object> conf, String cla
 
         Component uiServer = getComponent(streamingEngineService, COMPONENT_NAME_STORM_UI_SERVER)
                 .orElseThrow(() -> new RuntimeException(streamingEngine + " doesn't have " + COMPONENT_NAME_STORM_UI_SERVER + " as component"));
-        String uiHost = uiServer.getHosts().get(0);
-        Integer uiPort = uiServer.getPort();
+
+        Collection<ComponentProcess> uiServerProcesses = environmentService.listComponentProcessesInComponent(uiServer.getId());
```
HeartSaVioR: We extract the first one and name `uiServerProcess` so it will conflict.

streams/actions/src/main/java/com/hortonworks/streamline/streams/actions/container/TopologyActionsContainer.java
```diff
@@ -101,15 +104,30 @@ private TopologyActions initTopologyActions(Map<String, Object> conf, String cla
...
+        if (uiServerProcesses.isEmpty()) {
+            throw new RuntimeException(streamingEngine + " doesn't have any process for " + COMPONENT_NAME_STORM_UI_SERVER + " as component");
+        }
+
+        ComponentProcess uiServerProcess = uiServerProcesses.iterator().next();
+        String uiHost = uiServerProcess.getHost();
```
HeartSaVioR: Lots of local var can be final. Let's adopt style guide and integrate checkstyle to the build phase if we really want to handle style issue.

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/AbstractServiceRegistrar.java
```diff
@@ -117,8 +126,26 @@ public Service register(Cluster cluster, Config config, List<ConfigFileInfo> con
         return service;
     }
 
+    protected boolean validateComponentProcesses(List<ComponentProcess> componentProcesses) {
```
HeartSaVioR: plural form and List look like duplicated. `isComponentProcessesValid` looks OK.

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/AbstractServiceRegistrar.java
```diff
@@ -117,8 +126,26 @@ public Service register(Cluster cluster, Config config, List<ConfigFileInfo> con
...
+        if (componentProcesses.size() <= 0) {
+            return false;
+        }
 
+        return componentProcesses.stream().allMatch(componentProcess ->
+                StringUtils.isNotEmpty(componentProcess.getHost()) &&
+                        componentProcess.getPort() != null);
+    }
 
+    protected boolean validateComponentProcessesWithProtocolRequired(List<ComponentProcess> componentProcesses) {
```
HeartSaVioR: Same here: `isComponentProcessesWithProtocolRequiredValid` looks OK.

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/KafkaServiceRegistrar.java
```diff
@@ -15,67 +15,96 @@
...
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.math3.util.Pair;
 
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
```
HeartSaVioR: There's no HDFS_SITE, and no CORE_SITE.

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/KafkaServiceRegistrar.java
```diff
@@ -86,23 +115,68 @@ protected boolean validateServiceConfiguationsAsFlattenedMap(Map<String, String>
...
+    }
+
+    private ServiceConfiguration buildKafkaEnvServiceConfiguration(Config config) {
+        ServiceConfiguration serverProperties = new ServiceConfiguration();
+        serverProperties.setName(CONFIG_KAFKA_ENV);
+
+        Map<String, String> confMap = new HashMap<>();
+
+        try {
+            String json = objectMapper.writeValueAsString(confMap);
```
HeartSaVioR: Yes some configurations will be filled while importing from Ambari, but these are actually not needed so just putting empty JSON.

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/StormServiceRegistrar.java
```diff
@@ -37,69 +39,65 @@
 public class StormServiceRegistrar extends AbstractServiceRegistrar {
     public static final String COMPONENT_STORM_UI_SERVER = ComponentPropertyPattern.STORM_UI_SERVER.name();
     public static final String COMPONENT_NIMBUS = ComponentPropertyPattern.NIMBUS.name();
-    public static final String PARAM_STORM_UI_SERVER_HOSTNAME = "uiServerHostname";
-    public static final String PARAM_NIMBUS_HOSTNAMES = "nimbusesHostnames";
-    public static final String PARAM_NIMBUS_PRINCIPAL_NAME = "nimbusPrincipalName";
-    public static final String STORM_ENV = ServiceConfigurations.STORM.getConfNames()[1];
+
+    public static final String PARAM_NIMBUS_SEEDS = "nimbus.seeds";
```
HeartSaVioR: That represents config key in UI. If you think config key is better than param, we can change it.

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/StormServiceRegistrar.java
```diff
@@ -117,42 +115,124 @@ protected boolean validateServiceConfiguationsAsFlattenedMap(Map<String, String>
...
+        uiProcess.setHost(stormUiServerHost);
+        uiProcess.setPort(stormUiServerPort.intValue());
+
+        return new Pair<>(stormUiServer, Collections.singletonList(uiProcess));
+    }
+
+    private Number readNumberFromConfig(Config config, String parameterName) {
+        Number paramValue;
+        try {
+            paramValue = config.getAny(parameterName);
```
HeartSaVioR: As you are seeing that compiler doesn't complain, no need to cast.

HeartSaVioR: Ah OK you meant to reduce lines.

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/service/EnvironmentService.java
```diff
@@ -104,16 +106,24 @@ public Service initializeService(Cluster cluster, String serviceName) {
...
         Component component = new Component();
         component.setId(this.dao.nextId(COMPONENT_NAMESPACE));
         component.setName(componentName);
         component.setServiceId(service.getId());
         component.setTimestamp(System.currentTimeMillis());
-        component.setHosts(hosts);
         return component;
     }
 
+    public ComponentProcess initializeComponentProcess(Component component, String host) {
```
HeartSaVioR: Either is fine so just fixed.

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/service/metadata/json/KafkaBrokerListeners.java
```diff
@@ -55,7 +65,19 @@ public ListenersPropParsed(List<ListenersPropEntry> parsedProps) {
 
         public ListenersPropParsed(ServiceConfiguration config) {
             try {
-                final Map<String, String> configMap = config.getConfigurationMap();
+                Map<String, String> confMap = config.getConfigurationMap();
```
HeartSaVioR: config.getConfigurationMap() throws IOException so we should add throw to method signature (bad to do that since only this constructor will throw IOE). Let's keep it as it is.

streams/cluster/src/test/java/com/hortonworks/streamline/streams/cluster/register/impl/KafkaServiceRegistrarTest.java
HeartSaVioR: Yes number of component processes for broker. I'll put a comment.

streams/cluster/src/test/java/com/hortonworks/streamline/streams/cluster/register/impl/KafkaServiceRegistrarTest.java
HeartSaVioR: I intentionally catch the exception to check service is not added.

streams/cluster/src/test/java/com/hortonworks/streamline/streams/cluster/register/impl/KafkaServiceRegistrarTest.java
HeartSaVioR: Same here.

streams/cluster/src/test/java/com/hortonworks/streamline/streams/cluster/register/impl/StormServiceRegistrarTest.java
HeartSaVioR: Java `Set` instance declares this clearly. No need to worry about order.

```
    /**
     * Compares the specified object with this set for equality.  Returns
     * <tt>true</tt> if the specified object is also a set, the two sets
     * have the same size, and every member of the specified set is
     * contained in this set (or equivalently, every member of this set is
     * contained in the specified set).  This definition ensures that the
     * equals method works properly across different implementations of the
     * set interface.
     *
     * @param o object to be compared for equality with this set
     * @return <tt>true</tt> if the specified object is equal to this set
     */
    boolean equals(Object o);
```

streams/cluster/src/test/java/com/hortonworks/streamline/streams/cluster/register/impl/StormServiceRegistrarTest.java
HeartSaVioR: Same here.

streams/cluster/src/test/java/com/hortonworks/streamline/streams/cluster/register/impl/StormServiceRegistrarTest.java
HeartSaVioR: Same here.

streams/cluster/src/test/java/com/hortonworks/streamline/streams/cluster/register/impl/StormServiceRegistrarTest.java
HeartSaVioR: Same here.

streams/cluster/src/test/java/com/hortonworks/streamline/streams/cluster/register/impl/ZookeeperServiceRegistrarTest.java
HeartSaVioR: Same here.

HeartSaVioR:
@hmcl 
Thanks for the detailed review. I addressed review comments. 
Btw, I guess you're still in reviewing. If then please let me know if you finish your review. Thanks again!

streams/cluster/src/test/java/com/hortonworks/streamline/streams/cluster/register/impl/StormServiceRegistrarTest.java
hmcl: Sure. I was just making sure :)

bootstrap/services/storm-bundle.json
hmcl: I learned something :). Thanks.

bootstrap/services/storm-bundle.json
hmcl: OK. I think that this will not be an easy to process info for the user, but we can work on that later. Perhaps fix it in both places as you say if we come up with something better.

bootstrap/services/storm-bundle.json
hmcl: same comment

streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/StormServiceRegistrar.java
hmcl: Oh, I see. Would UI_PARAM? UI_SET_PARAM?  UI_DEFINED_PARAM be more suiting?

streams/cluster/src/test/java/com/hortonworks/streamline/streams/cluster/register/impl/KafkaServiceRegistrarTest.java
hmcl: Makes sense. Thanks.

HeartSaVioR:
Filed #875 . Thanks for the review @hmcl. I'll merge.

HeartSaVioR:
Merged via 0b9279e873c6d6d21e20121613959d2799be2d8d

PR #797, opened 2017-07-08
ISSUE-796: HBase Client Retries Must be Configurable

hmcl:
- Set hbase.client.retries.number to max value of 3
 - Refactor Hadoop configuration override to support prop values computed by functions
 - Enhance unit tests

Change-Id: I5628a37813b96634cbcdd66798d8afb988757f25


harshach:
+1

PR #799, opened 2017-07-10
ISSUE-740 Streamline initial screen in secure mode

### LINKED ISSUE #740: Streamline initial screen in secure mode
1. Install streamline in secure mode (build 448)
2. Hit streamline URL E.g. http://arun-hdf-secure-448-1.openstacklocal:7777/
3. It gives a page with hints (see screenshot)

Issues:
1. For firefox, the message mentions "sam-ui-hostname:7799" but I am hitting streamline at port 7777, it can be confusing for users. So the message can be made generic "sam-ui-hostname:PORT" or can include the actual port.

2. For chrome, it mentions only registry-ui-hostname. Streamline hostname missing here ?

Maybe we can just display the message relevant to the browser the user is using, and show the instructions for other browsers as some expandable option.

screen shot 2017-06-06 at 2 08 27 pm

sameer79:
This close #740 
currently, we are showing ref screenshot below:-

screenshot


shahsank3t:
+1

PR #807, opened 2017-07-12
ISSUE-774: Make reader and writer schema version mandatory for kafka

priyank5485:
Change-Id: Ieb8b2195f51018ed97a39993c8719da07faac982

### DIFF
--- a/bootstrap/components/sinks/kafka-sink-topology-component.json
+++ b/bootstrap/components/sinks/kafka-sink-topology-component.json
@@ -35,6 +35,15 @@
         "hint": "schema,override"
       },
       {
+        "uiName": "Writer schema version",
+        "fieldName": "writerSchemaVersion",
+        "isOptional": false,
+        "tooltip": "Version of schema for topic to use for serializing the message. Default is the latest version for the schema",
+        "type": "enumstring",
+        "options": [],
+        "hint": "dependsOn-topic,schemaVersion"
+      },
+      {
         "uiName": "Security Protocol",
         "fieldName": "securityProtocol",
         "isOptional": false,
@@ -68,15 +77,6 @@
         "options": [],
         "defaultValue": null,
         "hint": "inputFields"
-      },
-      {
-        "uiName": "Writer schema version",
-        "fieldName": "writerSchemaVersion",
-        "isOptional": false,
-        "tooltip": "Version of schema for topic to use for serializing the message. Default is the latest version for the schema",
-        "type": "enumstring",
-        "options": [],
-        "hint": "dependsOn-topic,schemaVersion"
       },
       {
         "uiName": "Ack Mode",
--- a/bootstrap/components/sinks/kafka-sink-topology-component.json
+++ b/bootstrap/components/sinks/kafka-sink-topology-component.json
@@ -72,10 +72,11 @@
       {
         "uiName": "Writer schema version",
         "fieldName": "writerSchemaVersion",
-        "isOptional": true,
+        "isOptional": false,
         "tooltip": "Version of schema for topic to use for serializing the message. Default is the latest version for the schema",
-        "type": "string",
-        "defaultValue": null
+        "type": "enumstring",
+        "options": [],
+        "hint": "dependsOn-topic,schemaVersion"
       },
       {
         "uiName": "Ack Mode",
--- a/bootstrap/components/sources/kafka-source-topology-component.json
+++ b/bootstrap/components/sources/kafka-source-topology-component.json
@@ -37,10 +37,11 @@
       {
         "uiName": "Reader schema version",
         "fieldName": "readerSchemaVersion",
-        "isOptional": true,
+        "isOptional": false,
         "tooltip": "Version of schema for topic to use for deserializing the message. Default is version used by kafka producer to write the message",
-        "type": "string",
-        "defaultValue": null
+        "type": "enumstring",
+        "options": [],
+        "hint": "dependsOn-topic,schemaVersion"
       },
       {
         "uiName": "Consumer Group Id",
--- a/streams/service/src/main/java/com/hortonworks/streamline/streams/service/SchemaResource.java
+++ b/streams/service/src/main/java/com/hortonworks/streamline/streams/service/SchemaResource.java
@@ -155,7 +155,7 @@ public class SchemaResource {
             }
         } catch (SchemaNotFoundException e) {
             LOG.error("Schema not found: [{}]", schemaName, e);
-            throw EntityNotFoundException.byId(schemaName);
+            throw EntityNotFoundException.byName(schemaName);
         } catch (Exception e) {
             throw new RuntimeException(e);
         }
@@ -173,18 +173,18 @@ public class SchemaResource {
                     (version)));
             if (schemaVersionInfo != null) {
                 LOG.debug("######### Received schema version from schema registry: {}", schemaVersionInfo);
-                String schema = schemaVersionInfo != null ? schemaVersionInfo.getSchemaText() : null;
+                String schema = schemaVersionInfo.getSchemaText();
                 if (schema != null && !schema.isEmpty()) {
                     schema = AvroStreamlineSchemaConverter.convertAvroSchemaToStreamlineSchema(schema);
                 }
                 LOG.debug("######### Converted schema: {}", schema);
                 return WSUtils.respondEntity(schema, OK);
             } else {
-                return WSUtils.respondEntity("", NOT_FOUND);
+                throw new SchemaNotFoundException("Version " + version + " of schema " + schemaName + " not found ");
             }
         } catch (SchemaNotFoundException e) {
             LOG.error("Schema not found: [{}]", schemaName, e);
-            throw EntityNotFoundException.byId(schemaName);
+            throw EntityNotFoundException.byName(schemaName);
         } catch (Exception e) {
             throw new RuntimeException(e);
         }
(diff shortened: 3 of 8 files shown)


priyank5485:
@shahsank3t @satishd @harshach Please review when you get a chance.
@shahsank3t We will need changes on UI side as well corresponding to the changes in the component spec json file for kafka source and sink.

bootstrap/components/sinks/kafka-sink-topology-component.json
```diff
@@ -72,10 +72,11 @@
       {
```
shahsank3t: We might want to keep this `writerSchemaVersion` below `topic` field to make a better user experience.

streams/service/src/main/java/com/hortonworks/streamline/streams/service/SchemaResource.java
```diff
@@ -132,4 +133,61 @@ public Response getKafkaSourceSchema(@PathParam("topicName") String topicName,
...
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response getSchemaForVersion(@PathParam("schemaName") String schemaName, @PathParam("version") String version,
+                                        @Context SecurityContext securityContext) {
+        try {
+            LOG.info("Get schema:version {}", schemaName + ":" + version);
+            SchemaVersionInfo schemaVersionInfo = schemaRegistryClient.getSchemaVersionInfo(new SchemaVersionKey(schemaName, Integer.parseInt
+                    (version)));
+            if (schemaVersionInfo != null) {
+                LOG.debug("######### Received schema version from schema registry: {}", schemaVersionInfo);
+                String schema = schemaVersionInfo != null ? schemaVersionInfo.getSchemaText() : null;
```
satishd: no need to have this check as it is already done in the enclosing if condition.

streams/service/src/main/java/com/hortonworks/streamline/streams/service/SchemaResource.java
```diff
@@ -132,4 +133,61 @@ public Response getKafkaSourceSchema(@PathParam("topicName") String topicName,
...
+                    schemaVersions.add(schemaVersionInfo.getVersion().toString());
+                }
+                return WSUtils.respondEntities(schemaVersions, OK);
+            } else {
+                LOG.debug("######### Received null response for schema versions from schema registry for schema: {}", schemaName);
+                return WSUtils.respondEntity(schemaVersions, NOT_FOUND);
+            }
+        } catch (SchemaNotFoundException e) {
+            LOG.error("Schema not found: [{}]", schemaName, e);
+            throw EntityNotFoundException.byId(schemaName);
```
satishd: This should be `EntityNotFoundException#byName` instead of `EntityNotFoundException#byId`

streams/service/src/main/java/com/hortonworks/streamline/streams/service/SchemaResource.java
```diff
@@ -132,4 +133,61 @@ public Response getKafkaSourceSchema(@PathParam("topicName") String topicName,
...
+    @GET
+    @Path("/{schemaName}/versions")
+    @Timed
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response getAllSchemaVersions(@PathParam("schemaName") String schemaName,
+                                         @Context SecurityContext securityContext) {
+        try {
+            LOG.info("Get all versions for schema : {}", schemaName);
+            Collection<SchemaVersionInfo> schemaVersionInfos = schemaRegistryClient.getAllVersions(schemaName);
+            Collection<String> schemaVersions = new ArrayList<>();
```
satishd: Is there any reason for using `Collection<String>` instead of `Collection<Integer>` ?

streams/service/src/main/java/com/hortonworks/streamline/streams/service/SchemaResource.java
```diff
@@ -132,4 +133,61 @@ public Response getKafkaSourceSchema(@PathParam("topicName") String topicName,
...
+            if (schemaVersionInfo != null) {
+                LOG.debug("######### Received schema version from schema registry: {}", schemaVersionInfo);
+                String schema = schemaVersionInfo != null ? schemaVersionInfo.getSchemaText() : null;
+                if (schema != null && !schema.isEmpty()) {
+                    schema = AvroStreamlineSchemaConverter.convertAvroSchemaToStreamlineSchema(schema);
+                }
+                LOG.debug("######### Converted schema: {}", schema);
+                return WSUtils.respondEntity(schema, OK);
+            } else {
+                return WSUtils.respondEntity("", NOT_FOUND);
```
satishd: may want to throw `SchemaNotFoundException ` here so that it throws right error to the client.

bootstrap/components/sinks/kafka-sink-topology-component.json
```diff
@@ -72,10 +72,11 @@
...
         "uiName": "Writer schema version",
         "fieldName": "writerSchemaVersion",
-        "isOptional": true,
+        "isOptional": false,
```
satishd: Do not we have any issue while importing earlier streamline topologies as they may not have this configuration?

bootstrap/components/sources/kafka-source-topology-component.json
```diff
@@ -37,10 +37,11 @@
       {
         "uiName": "Reader schema version",
         "fieldName": "readerSchemaVersion",
-        "isOptional": true,
+        "isOptional": false,
```
satishd: same comment as earlier about importing earlier version topologies.

streams/service/src/main/java/com/hortonworks/streamline/streams/service/SchemaResource.java
priyank5485: @satishd Because of the flux translation layer its easiest to make everything a string. Thats the approach I have taken for all kafka consumer and producer related properties - to keep them as string and handle the conversion where needed.

bootstrap/components/sinks/kafka-sink-topology-component.json
priyank5485: @shahsank3t Will do

streams/service/src/main/java/com/hortonworks/streamline/streams/service/SchemaResource.java
priyank5485: Will change

streams/service/src/main/java/com/hortonworks/streamline/streams/service/SchemaResource.java
priyank5485: Will fix it

streams/service/src/main/java/com/hortonworks/streamline/streams/service/SchemaResource.java
priyank5485: Will change

bootstrap/components/sources/kafka-source-topology-component.json
priyank5485: @satishd I dont think so. We are not exporting any component spec jsons. We should be okay since the fieldName has not changed.

priyank5485:
@satishd @shahsank3t I have addressed your review comments. Please review again. @shahsank3t I have included your UI patch as a commit in this PR, however webservice module is failing compilation due to a javascript error in Fields.jsx Can you fix it?

priyank5485:
@shahsank3t Fixed the UI build issue

shahsank3t:
+1

bootstrap/components/sources/kafka-source-topology-component.json
satishd: @priyank5485 I understand component spec jsons are not exported but the fields mentioned in uispec are represented by respective components which are stored in the backend storage. Earlier topology config's  kafka source will not have `readerSchemaVersion` in their config and when UI tries to load that in editor it may recognize that a mandatory field is missing and it may be throwing validation error. 

@shahsank3t @priyank5485 Pl let me know if I am missing anything here. I am just cautious whether we are breaking any topologies saved in earlier versions before this PR change.

shahsank3t: @satishd 
Two scenarios:
1. If user imports the topology and directly tries to deploy, that will still work since all the streams/edges and everything was properly linked (unless on runtime backend does a validation on every component field whether a mandatory one is missing or not)

2. If user imports and opens the component (in this case Kafka source), the missing field will be blank and if the user doesn't add value for that field, UI won't allow to save & keeps showing `Required Field` error.

priyank5485: @satishd Never hurts to be cautious. Glad we discussed it. Like @shahsank3t confirmed, this wont be an issue. Merging this in

satishd: @priyank5485 @shahsank3t I do not think current experience is seamless, it is still kind of broken as user needs to update with reader schema version when topology is edited. 
Ideally, user should not get any validation error when importing a topology created in earlier streamline versions. It should be upgraded seamlessly by showing respective info messages. In this case, this should have been upgraded by setting the right reader version(fetch from SR by sending the current schema text and its schema name) without asking user to configure it. 
This is not really a UI issue but a general import issue which should be handled by respective sources/sinks/processors, filed issue #822 on the same.

priyank5485: @satishd This change is kind of a backwards incompatible so its not possible to have a seamless experience. Moreover, I think UI forcing the user to pick a reader schema version in newer version is not a big inconvenience. I agree with you on showing an info message. However, setting the right reader version approach you suggested is not clear and seems like an overkill.  What is the right reader version here? You mentioned you get that by sending current schema text and name and get it from SR. Where do we get schema text from? And which version? By default in earlier versions if reader schema version is not specified the writer schema version from payload is used to deserialize.

PR #812, opened 2017-07-13
ISSUE-802: Streaming Analytics Manager Application status issue

### LINKED ISSUE #802: Streaming Analytics Manager Application status issue
When an application deployment fails due to some errors (in this case, atlas hook missing jars for example), the application shows in running state in SAM console. I can also see the storm topology running from the Storm UI.
When trying to kill that app from the Streaming Analytics Manager UI, it fails to kill it with the following message:
An exception with message [Invalid action for current state: TOPOLOGY_STATE_DEPLOYMENT_FAILED] was thrown while processing request.
The only way to get out of this state is to kill the Storm topology on the Storm UI.
There are two issues with this:
- What is the actual state of the app? For Storm, it is running. For SAM, it is not and its status is TOPOLOGY_STATE_DEPLOYMENT_FAILED. This needs to be clarified.
- The ability to kill the app from the SAM UI without having to go to the Storm UI.

arunmahadevan:
Kill any running instance of topology in case the deployment throws Exceptions but topology starts running in the cluster

Fixes #802

### DIFF
--- a/streams/actions/src/main/java/com/hortonworks/streamline/streams/actions/StatusImpl.java
+++ b/streams/actions/src/main/java/com/hortonworks/streamline/streams/actions/StatusImpl.java
@@ -19,8 +19,8 @@ import java.util.HashMap;
 import java.util.Map;
 
 public class StatusImpl implements TopologyActions.Status {
-    String status = "Unknown"; // default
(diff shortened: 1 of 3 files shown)


harshach:
+1

Issue #801, opened 2017-07-11
Handle errors during custom processor upload


arunmahadevan:
Upload a custom processor after entering an empty JSON `{}` in the input and output schema. The upload wheel icon keeps spinning and nothing happens.

The HTTP status code is 400 and the response content is 
`Can not deserialize instance of java.util.ArrayList out of START_OBJECT token
 at [Source: {"streamingEngine":"STORM","name":"test","description":"test","customProcessorImpl":"com.hortonworks.streamline.examples.processors.ConsoleCustomProcessor","inputSchema":{"fields":{}},"outputSchema":{"fields":{}},"topologyComponentUISpecification":{"fields":[{"fieldName":"a","uiName":"a","isOptional":false,"type":"string","defaultValue":"a","isUserInput":true,"tooltip":"a"}]},"jarFileName":"streamline-examples-processors-0.1.0-SNAPSHOT.jar"}; line: 1, column: 181] (through reference chain: com.hortonworks.streamline.streams.catalog.processor.CustomProcessorInfo["inputSchema"]->com.hortonworks.registries.common.Schema["fields"])
`
However UI seems to be parsing the response text as a JSON and failing. I see the below error in the developer console. `Uncaught (in promise) SyntaxError: Unexpected token C in JSON at position 0`

UI should also take the HTTP response code into account before trying to parse the response text as a JSON. You may also want to check/fix the other pages where we let user upload jars.

shahsank3t:
https://github.com/hortonworks/streamline/pull/814 PR has been raised to fix this issue

shahsank3t:
PR has been merged

PR #814, opened 2017-07-13
ISSUE-801 Handle errors during custom processor upload

joylyn:
This PR contains UI side changes for #801


shahsank3t:
+1 LGTM

@arunmahadevan  can you please test the scenario once from your end and verify if the error message is proper or needs to be changed. Thanks in advance.

Issue #826, opened 2017-07-18
Sanitize mysql create_tables scripts


arunmahadevan:
In many columns we have defined column types as `varchar(256)`. This can potentially throw error during table creation when using utf-8 charset. Revisit the column definitions and use sane limits with max = 255 instead of 256.

Exception in thread "main" java.sql.SQLSyntaxErrorException: (conn:12499) Specified key was too long; max key length is 767 bytes
--

HeartSaVioR:
It will just works (though with utf8mb4 it should have same issue), but ideally revisiting the column definitions and setting index as proper length would be better. In many case I didn't see the benefit of indexing full text.

HeartSaVioR:
The issue is not automatically closed. Resolved via #827 .

PR #827, opened 2017-07-18
ISSUE-826 Sanitize mysql create_tables scripts

satishd:


satishd:
>It still have potential issue with utf8mb4 so this might not an ideal, but this is better anyway.

@HeartSaVioR Already mentioned that in git commit message and user can always change the script as utf8mb4 is not a common case.

HeartSaVioR:
@satishd 
Yeah great. I missed to see commit message.

satishd:
Closes #826

PR #828, opened 2017-07-18
ISSUE-774: reader and writer schema value needs to passed as string

shahsank3t:
Reader/writer schema version is being passed as int value which should be passed as a string from UI. This change resolves for both source & sink.


priyank5485:
+1

PR #830, opened 2017-07-18
ISSUE-829: Fix for Concat function

arunmahadevan:
`CONCAT` was assumed to be a calcite built-in function, however the calcite parser
returns it as a unresolved user defined function.
Due to this, to topology deployment code tries to download and package the jar which fails.

1. Implemented `CONCAT` as a streamline function.
2. Commenting the `SUBSTRING` function for now since theres no way from UI to pass the start index field as parameter for the Function.

Tested projection processor with following functions (`UPPER`, `LOWER`, `CONCAT`, `INITCAP`, `CHAR_LENGTH`) and it works.


arunmahadevan:
@harshach , @priyank5485 please take a look.

HeartSaVioR:
In calcite, CONCAT is implemented as `JDBC function` (I interpreted this expression as non-ANSI standard) so need to wrap it with `{fn }`.
`||` is used for concatenating two strings, and this is ANSI standard, despite some "famous" databases ignore to support the standard.
https://calcite.apache.org/docs/reference.html#jdbc-function-escape
https://stackoverflow.com/questions/11403931/what-is-the-difference-between-operator-and-concat-function-in-oracle

I'm fine with any way (wrap it, convert to `||`, overriding function as this patch does) so please treat my comment as information. If we don't like wrapping with fn, +1 for the patch.

Btw, this also means we need to have at least SQL function reference page if we would want to support handling raw query.

bootstrap/bootstrap-udf.sh
```diff
@@ -64,13 +64,15 @@ curl -i --negotiate -u:anyUser  -b /tmp/cookiejar.txt -c /tmp/cookiejar.txt -s -
 echo "  - identity"
 curl -i --negotiate -u:anyUser  -b /tmp/cookiejar.txt -c /tmp/cookiejar.txt -s -X POST "${CATALOG_ROOT_URL}/streams/udfs" -F udfJarFile=@${jarFile} -F udfConfig='{"name":"IDENTITY", "displayName": "Identity", "description": "Identity function", "type":"FUNCTION", "className":"com.hortonworks.streamline.streams.udf.Identity", "builtin":true};type=application/json'
 
+echo "  - concat"
+curl -i --negotiate -u:anyUser  -b /tmp/cookiejar.txt -c /tmp/cookiejar.txt -s -X POST "${CATALOG_ROOT_URL}/streams/udfs" -F udfJarFile=@${jarFile} -F udfConfig='{"name":"CONCAT", "displayName": "CONCAT", "description": "Concatenate", "type":"FUNCTION", "className":"com.hortonworks.streamline.streams.udf.Concat", "builtin":true};type=application/json'
```
HeartSaVioR: Just wondering: is it working without defining argTypes and returnType? If then what's the usage of two fields?

HeartSaVioR:
Same as SUBSTRING: as you may already know, the syntax for ANSI standard SUBSTRING is `SUBSTRING (char_exp FROM start_position [ FOR length ])`, which is fairly different from UDF. There're also many other functions in Calcite reference page which have their own syntax.

So may need to decide on this: ignoring ANSI standard and convert functions as receiving comma-separated parameters, or don't modify its syntax (and semantic) and let UI/users refer the reference page or expect ANSI standard by default.

priyank5485:
@arunmahadevan Tested the patch. No more deployment error. +1 for this change

bootstrap/bootstrap-udf.sh
arunmahadevan: argTypes and returnTypes are figured out from the implementation. For calcite provided functions we pass it in the POST request. These are used by UI to validate if the user is selecting compatible arguments for a function.

arunmahadevan:
@HeartSaVioR , we can support ANSI standard functions and syntax.  We need not support calicte syntax and for such functions we can expose our own UDFS than wrapping the functions in `{fn }`. `SUBSTRING` is disabled for now since the UI does not support passing literal arguments for functions.

PR #832, opened 2017-07-19
ISSUE-765: DB based file storage

### LINKED ISSUE #765: Implement FileSync to synchornize the jars uploaded from one server onto another
Currently, we expect users to use HDFS as the FileStorage incase of HA, so that the users uploading jar from one server connection can be seen on another server. This is a heavy dependency on HA and we should look into implementing a FileSync and default LocalFileStorage and remove HDFS as required FS incase of HA.
cc @arunmahadevan and @satishd please add if you have any thoughts on this.
This will be needed for registry as well.

arunmahadevan:
Currently, we expect users to use HDFS as the FileStorage incase of HA, so that
users uploading jar from one server connection can be seen on another server.

Storing the jars as a blob into DB as a workaround lets us do away with HDFS dependency for HA. This will delegate the replication to the db and if users need HA for stored files, they can have a DB cluster with some kind of a master-master or a master-slave replication.

Fixes #765


webservice/src/main/java/com/hortonworks/streamline/webservice/StreamlineApplication.java
```diff
@@ -217,11 +214,14 @@ private CacheBuilder getGuavaCacheBuilder() {
         return CacheBuilder.newBuilder().maximumSize(maxSize);
     }
 
-    private FileStorage getJarStorage (StreamlineConfiguration configuration) {
+    private FileStorage getJarStorage (StreamlineConfiguration configuration, StorageManager storageManager) {
         FileStorage fileStorage = null;
         try {
             fileStorage = ReflectionHelper.newInstance(configuration.getFileStorageConfiguration().getClassName());
             fileStorage.init(configuration.getFileStorageConfiguration().getProperties());
+            if (fileStorage instanceof DbFileStorage) {
```
HeartSaVioR: If we're a fan of interface, we even could have an interface `StorageManagerAware` and let DbFileStorage implement it. Maybe more flexible given that FileStorage is a kind of plug-in.

bootstrap/sql/mysql/create_tables.sql
```diff
@@ -69,6 +69,14 @@ CREATE TABLE IF NOT EXISTS file (
     UNIQUE KEY `jars_UK_name_version` (name, version)
 );
 
+CREATE TABLE IF NOT EXISTS fileblob (
+    name VARCHAR(255) NOT NULL,
+    version BIGINT NOT NULL,
+    data LONGBLOB NOT NULL,
+    timestamp BIGINT,
```
omkreddy: AFAIK, "timestamp" is reserved word in mysql/oracle. It is better to rename  this field.

storage/core/src/main/java/com/hortonworks/streamline/storage/CacheBackedStorageManager.java
```diff
@@ -52,32 +52,50 @@ public void init(Map<String, Object> properties) {
...
+        if (storable.isCacheable()) {
+            cache.put(storable.getStorableKey(), storable);
+        }
     }
 
     @Override
     public <T extends Storable> T get(StorableKey key) throws StorageException {
-        return (T) cache.get(key);
+        Storable storable = cache.get(key);
+        if (storable != null && !storable.isCacheable()) {
```
omkreddy: Do we need this check? we are not adding non-cacheable's to cache.

common/src/main/java/com/hortonworks/streamline/common/util/ReflectionHelper.java
```diff
@@ -47,7 +47,24 @@
...
-        Method method = object.getClass().getMethod(methodName, valueToSet.getClass());
+        Method method = null;
+        try {
+            method = object.getClass().getMethod(methodName, valueToSet.getClass());
+        } catch (NoSuchMethodException ex) {
+            // try setters that accept super types
+            Method[] methods = object.getClass().getMethods();
+            for (int i = 0; i < methods.length; i++) {
+                if (methods[i].getParameterCount() == 1) {
+                    if (methods[i].getParameters()[0].getType().isAssignableFrom(valueToSet.getClass())) {
```
arunmahadevan: Missed checking method name, will add.

storage/core/src/main/java/com/hortonworks/streamline/storage/filestorage/DbFileStorage.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    @Override
+    public String uploadFile(InputStream inputStream, String name) throws IOException {
+        LOG.debug("Upload '{}'", name);
+        Optional<FileBlob> existing = get(name);
+        if (existing.isPresent()) {
+            LOG.debug("Trying to update existing file '{}'", name);
+            FileBlob updated = new FileBlob(existing.get());
+            updated.setData(inputStream);
```
arunmahadevan: I had initially wrapped the inputStream in `BufferedInputStream` but later decided to remove it and leave it to the caller. I missed removing it from the add part. Will remove it.

storage/core/src/main/java/com/hortonworks/streamline/storage/impl/jdbc/provider/mysql/query/MysqlUpdateQuery.java
```diff
@@ -0,0 +1,40 @@
...
+    public MysqlUpdateQuery(Storable storable) {
+        super(storable);
+        setParameterizedSqlReal();
+    }
+
+    @Override
+    protected void setParameterizedSql() {
+    }
+
+    // TODO: Refactor setParameterizedSql above
```
arunmahadevan: will address.

bootstrap/sql/mysql/create_tables.sql
arunmahadevan: "timestamp" is a keyword but not a reserved keyword, so it can be used as column name without quoting. Actually `timestamp` is used in most of the streamline tables so if we want to rename we can take it up in a separate issue, but its safe to leave it as is.
https://dev.mysql.com/doc/refman/5.7/en/keywords.html

storage/core/src/main/java/com/hortonworks/streamline/storage/CacheBackedStorageManager.java
arunmahadevan: In `GuavaCache` implementation we load the values via a Loading cache so by invoking `cache.get(key)` it automatically loads it into cache. So we need to remove it. Will add a comment to make it clear.

arunmahadevan:
@HeartSaVioR , @omkreddy , addressed review comments.

HeartSaVioR:
Confirmed the following commit addresses all review comments from me.
Still not manually tested though since for now I'm debugging on my own work.

Please take my comment as +1 when anyone verifies the patch via manual test.

common/src/main/java/com/hortonworks/streamline/common/util/ReflectionHelper.java
```diff
@@ -47,7 +47,24 @@
...
-        Method method = object.getClass().getMethod(methodName, valueToSet.getClass());
+        Method method = null;
+        try {
+            method = object.getClass().getMethod(methodName, valueToSet.getClass());
+        } catch (NoSuchMethodException ex) {
+            // try setters that accept super types
+            Method[] methods = object.getClass().getMethods();
+            for (int i = 0; i < methods.length; i++) {
+                if (methods[i].getName().equals(methodName) && methods[i].getParameterCount() == 1) {
+                    if (methods[i].getParameters()[0].getType().isAssignableFrom(valueToSet.getClass())) {
```
hmcl: what if there are multiple methods that verify isAssignableFrom?

storage/core/src/main/java/com/hortonworks/streamline/storage/CacheBackedStorageManager.java
```diff
@@ -52,32 +52,51 @@ public void init(Map<String, Object> properties) {
...
 
     @Override
     public <T extends Storable> T remove(StorableKey key) throws StorageException {
-        writer.remove(key);
-        final T oldVal = (T) cache.get(key);
-        cache.remove(key);
-        return oldVal;
+        Storable storable = (Storable) writer.remove(key);
+        if (storable != null && storable.isCacheable()) {
+            storable = cache.get(key);
```
hmcl: what's the reasoning behind overriding the storable variable retrieved from the DB with the value in cache?

storage/core/src/main/java/com/hortonworks/streamline/storage/CacheBackedStorageManager.java
```diff
@@ -52,32 +52,51 @@ public void init(Map<String, Object> properties) {
...
+            cache.put(storable.getStorableKey(), storable);
+        }
     }
 
     @Override
     public <T extends Storable> T get(StorableKey key) throws StorageException {
-        return (T) cache.get(key);
+        Storable storable = cache.get(key);
+        // we could load the entries via a LoadingCache so need to explicitly remove it
+        if (storable != null && !storable.isCacheable()) {
```
hmcl: I am not following this if statement. If the storable is the result of `cache.get(key)` and it is not null, how can it be not cacheable? If it is not cacheable, the loading cache should not cache it.

storage/core/src/main/java/com/hortonworks/streamline/storage/CacheBackedStorageManager.java
```diff
@@ -52,32 +52,51 @@ public void init(Map<String, Object> properties) {
...
+            cache.remove(key);
+        }
+        return (T) storable;
     }
 
     @Override   //TODO:
```
hmcl: please remove the TODO

storage/core/src/main/java/com/hortonworks/streamline/storage/annotation/VersionField.java
```diff
@@ -0,0 +1,29 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation for field that can be used for optimistic locking
+ */
+@Target(ElementType.FIELD)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface VersionField {
```
hmcl: Is `FieldVersion` a more suiting name?

storage/core/src/main/java/com/hortonworks/streamline/storage/filestorage/DbFileStorage.java
```diff
@@ -0,0 +1,91 @@
...
+    @Override
+    public void init(Map<String, String> config) {
+    }
+
+    @Override
+    public String uploadFile(InputStream inputStream, String name) throws IOException {
+        LOG.debug("Upload '{}'", name);
+        Optional<FileBlob> existing = get(name);
+        if (existing.isPresent()) {
+            LOG.debug("Trying to update existing file '{}'", name);
```
hmcl: Updating existing...

storage/core/src/main/java/com/hortonworks/streamline/storage/filestorage/DbFileStorage.java
```diff
@@ -0,0 +1,91 @@
...
+            FileBlob updated = new FileBlob(existing.get());
+            updated.setData(inputStream);
+            updated.setVersion(existing.get().getVersion() + 1);
+            updated.setTimestamp(System.currentTimeMillis());
+            dao.update(updated);
+        } else {
+            LOG.debug("Trying to add a new file '{}'", name);
```
hmcl: Adding new file ...

storage/core/src/main/java/com/hortonworks/streamline/storage/filestorage/DbFileStorage.java
```diff
@@ -0,0 +1,91 @@
...
+public class DbFileStorage implements FileStorage, StorageManagerAware {
+    private static final Logger LOG = LoggerFactory.getLogger(DbFileStorage.class);
+    private StorageManager dao;
+
+    @Override
+    public void init(Map<String, String> config) {
+    }
+
+    @Override
+    public String uploadFile(InputStream inputStream, String name) throws IOException {
```
hmcl: NIT: The if/else blocks have very similar code. It might be possible to create a helper method to reuse the code.

storage/core/src/main/java/com/hortonworks/streamline/storage/filestorage/DbFileStorage.java
```diff
@@ -0,0 +1,91 @@
...
+    public InputStream downloadFile(String name) throws IOException {
+        LOG.debug("Download file '{}'", name);
+        FileBlob res = get(name)
+                .orElseThrow(() -> new IOException("Not able to get file blob with name : " + name));
+        return res.getData();
+    }
+
+    @Override
+    public boolean deleteFile(String name) throws IOException {
+        LOG.debug("Delete file '{}'", name);
```
hmcl: Deleting

storage/core/src/main/java/com/hortonworks/streamline/storage/filestorage/DbFileStorage.java
```diff
@@ -0,0 +1,91 @@
...
+            fileBlob.setVersion(0L);
+            fileBlob.setData(inputStream);
+            dao.add(fileBlob);
+        }
+        return name;
+    }
+
+    @Override
+    public InputStream downloadFile(String name) throws IOException {
+        LOG.debug("Download file '{}'", name);
```
hmcl: Downloading

storage/core/src/main/java/com/hortonworks/streamline/storage/filestorage/FileBlob.java
```diff
@@ -0,0 +1,110 @@
...
+import com.hortonworks.streamline.storage.annotation.StorableEntity;
+import com.hortonworks.streamline.storage.annotation.VersionField;
+import com.hortonworks.streamline.storage.catalog.AbstractStorable;
+
+import java.io.ByteArrayInputStream;
+import java.io.InputStream;
+import java.util.Collections;
+
+@StorableEntity
+public class FileBlob extends AbstractStorable {
```
hmcl: Is FileBlob a binary file ?

webservice/src/main/java/com/hortonworks/streamline/webservice/StreamlineApplication.java
```diff
@@ -217,11 +214,14 @@ private CacheBuilder getGuavaCacheBuilder() {
...
     }
 
-    private FileStorage getJarStorage (StreamlineConfiguration configuration) {
+    private FileStorage getJarStorage (StreamlineConfiguration configuration, StorageManager storageManager) {
         FileStorage fileStorage = null;
         try {
             fileStorage = ReflectionHelper.newInstance(configuration.getFileStorageConfiguration().getClassName());
             fileStorage.init(configuration.getFileStorageConfiguration().getProperties());
+            if (fileStorage instanceof StorageManagerAware) {
+                ((DbFileStorage) fileStorage).setStorageManager(storageManager);
```
hmcl: can't this produce ClassCastException?

storage/core/src/main/java/com/hortonworks/streamline/storage/impl/jdbc/JdbcStorageManager.java
```diff
@@ -81,6 +84,14 @@ public void addOrUpdate(Storable storable) throws StorageException {
     }
 
     @Override
+    public void update(Storable storable) {
+        if (queryExecutor.update(storable) == 0) {
```
hmcl: I am not quite following the relation between updating 0 rows and concurrent update? Shouldn't the ConcurrentUpdateException be thrown in queryExecutor.update(storable) method instead?

storage/core/src/main/java/com/hortonworks/streamline/storage/Storable.java
```diff
@@ -85,4 +85,11 @@
      */
     void setId(Long id);
 
+    /**
+     * Returns if the storable should be cached or not
+     */
+    default boolean isCacheable() {
```
arunmahadevan: FileBlob contains an `InputStream` which is basically invalid after the content is sent out, so should not be cached. Other cases could be things like large `byte[]` which we dont want to be cached.

storage/core/src/main/java/com/hortonworks/streamline/storage/StorageManager.java
```diff
@@ -68,6 +68,14 @@
     void addOrUpdate(Storable storable) throws StorageException;
 
     /**
+     * Updates an existing {@link Storable}. Throws {@link StorageException} if the
+     * storable could not be updated.
+     *
+     * @param storable the storable
+     */
+    void update(Storable storable);
```
arunmahadevan: It updates existing rows. It will not attempt to add a row automatically. I am not sure why we have `addOrUpdate`, it has confusing semantics and the caller cant know whether the row existed and got updated or just got added. We should have had only `add` and `update` since we have separate POST and PUT calls for adding and updating.

storage/core/src/main/java/com/hortonworks/streamline/storage/filestorage/DbFileStorage.java
arunmahadevan: Its not much, so would like to leave it as is.

storage/core/src/main/java/com/hortonworks/streamline/storage/filestorage/DbFileStorage.java
```diff
@@ -0,0 +1,91 @@
...
+
+    @Override
+    public String uploadFile(InputStream inputStream, String name) throws IOException {
+        LOG.debug("Upload '{}'", name);
+        Optional<FileBlob> existing = get(name);
+        if (existing.isPresent()) {
+            LOG.debug("Trying to update existing file '{}'", name);
+            FileBlob updated = new FileBlob(existing.get());
+            updated.setData(inputStream);
+            updated.setVersion(existing.get().getVersion() + 1);
```
arunmahadevan: Not required. The update will fail if we are not updating the previous version. (i.e if there was a concurrent update)

storage/core/src/main/java/com/hortonworks/streamline/storage/filestorage/DbFileStorage.java
```diff
@@ -0,0 +1,91 @@
...
+        return dao.exists(FileBlob.getStorableKey(name));
+    }
+
+    private Optional<FileBlob> get(String name) {
+        return Optional.ofNullable(dao.get(FileBlob.getStorableKey(name)));
+    }
+
+    @Override
+    public void setStorageManager(StorageManager storageManager) {
+        this.dao = storageManager;
```
arunmahadevan: Since these are constructed via reflection and the dao is not applicable to other `FileStorage` implementations, added the setter.

common/src/main/java/com/hortonworks/streamline/common/util/ReflectionHelper.java
arunmahadevan: It will apply the first match. Typically there should be only one match. Overloading methods with subtypes is **bad** and should not be done since Java resolves method overload based on the declared type and not the actual type of the argument.

storage/core/src/main/java/com/hortonworks/streamline/storage/CacheBackedStorageManager.java
arunmahadevan: I refactored the existing code which was like below, so I am not sure why was it that way  (may be we assumed the cache has the latest value which could be out of sync with the db?)
```java
writer.remove(key);
final T oldVal = (T) cache.get(key);
cache.remove(key);
return oldVal;
```

storage/core/src/main/java/com/hortonworks/streamline/storage/CacheBackedStorageManager.java
arunmahadevan: I think we assumed a loading cache here so we don't explicitly retrieve the value from db and put it into the cache. With the existing loading cache implementation its not straightforward to instruct loading cache to load but not cache and still return the value so I would like to defer this to later.

storage/core/src/main/java/com/hortonworks/streamline/storage/annotation/VersionField.java
arunmahadevan: Its a field that can be used for versioning, not the version of a field, so would like to retain VersionField .

storage/core/src/main/java/com/hortonworks/streamline/storage/filestorage/FileBlob.java
arunmahadevan: it can be any file, stored into a Blob field in DB

webservice/src/main/java/com/hortonworks/streamline/webservice/StreamlineApplication.java
arunmahadevan: will address

storage/core/src/main/java/com/hortonworks/streamline/storage/impl/jdbc/JdbcStorageManager.java
arunmahadevan: The query executor is low-level and just returns the number of rows updated. This logic is common for mysql/postgres so would like to retain it here.

So assuming we passed a Storable with proper values (including PK and optional Version field), only way update would return 0, is if that row (with that PK plus optional version field) was not found which means there was a concurrent update or delete for that same record. If there was some other DB issues, it would manifest as some JDBCException which will propagate up.

arunmahadevan:
@hmcl , thanks for the review, answered the comments and also pushed some changes. If I missed something let me know.

arunmahadevan:
@HeartSaVioR , @omkreddy , @hmcl let me know if you are fine with this patch or you still see any pending issues. 

@harshach , QE keeps hitting the FileNotFound exception, so if we can get this merged and set the default to DbFileStorage for HA mode deployments it will help.

storage/core/src/main/java/com/hortonworks/streamline/storage/StorageManager.java
hmcl: I agree. So I suggest that we create an issue to address that.

hmcl:
@arunmahadevan I am doing one last pass. I should be done soon. Thanks for the patch & for addressing the review comments.

storage/core/src/main/java/com/hortonworks/streamline/storage/StorageManager.java
arunmahadevan: Raised: https://github.com/hortonworks/streamline/issues/846

>What happens if you try updating a storable that does not exist?

It will throw an exception.

storage/core/src/main/java/com/hortonworks/streamline/storage/CacheBackedStorageManager.java
arunmahadevan: Ok

storage/core/src/main/java/com/hortonworks/streamline/storage/CacheBackedStorageManager.java
arunmahadevan: Raised https://github.com/hortonworks/streamline/issues/847

storage/core/src/main/java/com/hortonworks/streamline/storage/impl/jdbc/JdbcStorageManager.java
arunmahadevan: It could be,

1. The row existed during the query, but got deleted before it could be updated.
 In this case its a ConcurrentModification.
2. Wrong PK. Ideally this should not happen we expect only valid rows to be updated. i.e. the REST should have returned the row to the client which updated some non PK columns.

Its not really possible to differentiate b/w 1 and 2 from the storage manager and since we dont expect (2) in most of the cases, it should be ok to throw ConcurrentUpdateException. 

I will reword the exception message to include (2) as a possibility.

storage/core/src/main/java/com/hortonworks/streamline/storage/filestorage/DbFileStorage.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    @Override
+    public boolean deleteFile(String name) throws IOException {
+        LOG.debug("Deleting file '{}'", name);
+        return dao.remove(FileBlob.getStorableKey(name)) != null;
+    }
+
+    @Override
+    public boolean exists(String name) {
```
arunmahadevan: I think ideally the other methods should be renamed e.g. `upload` instead of `uploadFile` (since its already defined in the context of FileStorage and so implies that we are dealing with files). I will rename the other methods.

storage/core/src/main/java/com/hortonworks/streamline/storage/impl/jdbc/provider/mysql/query/MySqlQuery.java
```diff
@@ -20,13 +20,15 @@
 import com.hortonworks.streamline.storage.impl.jdbc.provider.sql.query.AbstractSqlQuery;
 
 public class MySqlQuery extends AbstractSqlQuery {
+    private final String theSql;
```
arunmahadevan: right, i had renamed sql->theSql to identify the places where the variable was used earlier and forgot to rename.

storage/core/src/main/java/com/hortonworks/streamline/storage/impl/jdbc/provider/sql/factory/AbstractQueryExecutor.java
```diff
@@ -176,8 +173,8 @@ public void setStorableFactory(StorableFactory storableFactory) {
 
     // =============== Private helper Methods ===============
 
-    protected void executeUpdate(SqlQuery sqlBuilder) {
-        getQueryExecution(sqlBuilder).executeUpdate();
+    protected int executeUpdate(SqlQuery sqlBuilder) {
```
arunmahadevan: ok

arunmahadevan:
@hmcl , addressed latest comments in ea41943. Will squash after the +1.

storage/core/src/main/java/com/hortonworks/streamline/storage/filestorage/DbFileStorage.java
hmcl: OK. I believe you forgot to rename the methods. You can do it later if you want. Let's just file a ticket such that we don't forget.

arunmahadevan:
@hmcl , renamed the methods and rebased into 3 logical commits.

PR #834, opened 2017-07-20
ISSUE-833: Fix failing AvroStreamsSnapshotDeserializerTest

priyank5485:
Change-Id: Ic782797a78708d65a0a1c7c820e0ae7a3e0cf750

### DIFF
--- a/streams/runners/storm/runtime/src/test/java/com/hortonworks/streamline/streams/runtime/AvroStreamsSnapshotDeserializerTest.java
+++ b/streams/runners/storm/runtime/src/test/java/com/hortonworks/streamline/streams/runtime/AvroStreamsSnapshotDeserializerTest.java
@@ -20,6 +20,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.hortonworks.registries.schemaregistry.SchemaIdVersion;
 import com.hortonworks.registries.schemaregistry.SchemaMetadata;
 import com.hortonworks.registries.schemaregistry.SchemaMetadataInfo;
+import com.hortonworks.registries.schemaregistry.SchemaVersionInfo;
 import com.hortonworks.registries.schemaregistry.SchemaVersionKey;
 import com.hortonworks.registries.schemaregistry.client.SchemaRegistryClient;
 import com.hortonworks.registries.schemaregistry.serdes.avro.AvroSnapshotSerializer;
@@ -44,7 +45,7 @@ import java.util.Map;
  */
 public class AvroStreamsSnapshotDeserializerTest {
     private static final Logger LOG = LoggerFactory.getLogger(AvroStreamsSnapshotDeserializerTest.class);
-
+    private static final SchemaIdVersion SCHEMA_ID_VERSION = new SchemaIdVersion(1L, 1);
 
     @Mocked
     SchemaRegistryClient mockSchemaRegistryClient;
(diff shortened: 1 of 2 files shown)


priyank5485:
@satishd I have already pushed this fix to HDF-3.0-maint since it was breaking the build and was a blocker. Can you review? If you want to make any changes then we will need to make it in HDF-3.0-maint as well

streams/runners/storm/runtime/src/test/java/com/hortonworks/streamline/streams/runtime/AvroStreamsSnapshotDeserializerTest.java
```diff
@@ -66,7 +68,10 @@ public void testAvroPayloadConversions() throws Exception {
 
             new Expectations() {
                 {
-                    mockSchemaRegistryClient.getSchemaMetadataInfo(anyLong);
+                    mockSchemaRegistryClient.getSchemaVersionInfo(withEqual(SCHEMA_ID_VERSION));
```
hmcl: What's the purpose of withEqual in here ?

streams/runners/storm/runtime/src/test/java/com/hortonworks/streamline/streams/runtime/AvroStreamsSnapshotDeserializerTest.java
```diff
@@ -66,7 +68,10 @@ public void testAvroPayloadConversions() throws Exception {
...
+                    //only versions of schema matters for this mock, as getVersion is only used during de-serialization
+                    result = new SchemaVersionInfo(1l, schemaMetadata.getName(), SCHEMA_ID_VERSION.getVersion(), "doesNotMatter",1l, "doesNotMatter");
+                    mockSchemaRegistryClient.getSchemaMetadataInfo(withEqual(schemaMetadata.getName()));
```
hmcl: What's the purpose of withEqual in here ?

streams/runners/storm/runtime/src/test/java/com/hortonworks/streamline/streams/runtime/AvroStreamsSnapshotDeserializerTest.java
priyank5485: Not that important really. Its just a little stricter than before. It just makes sure that its called with same SchemaIdVersion as the one used for serialization.

streams/runners/storm/runtime/src/test/java/com/hortonworks/streamline/streams/runtime/AvroStreamsSnapshotDeserializerTest.java
priyank5485: Not that important really. Its just a little stricter than before. It just makes sure that its called with same schema name as the one used for serialization.

streams/runners/storm/runtime/src/test/java/com/hortonworks/streamline/streams/runtime/AvroStreamsSnapshotDeserializerTest.java
HeartSaVioR: http://jmockit.org/api1x/mockit/Expectations.html#withEqual-T-

According to API spec, seems like just putting desired value just works, which is same behavior for Mockito.when(). (`all actual values` or `all matchers`)

satishd: `withEqual` May not be needed here, actual object should be sufficient. Javadoc says

`Usually, this particular method should not be used. Instead, simply pass the desired argument value directly, without any matcher. Only when specifying values for a varargs method it's useful, and even then only when some other argument matcher is also used.`

satishd:
@priyank5485 I am fine with these changes except using withEqual as mentioned in the comment.

HeartSaVioR:
I also don't have strong opinion on that. Please treat it as one of opinions. +1 on the change.

hmcl:
I also don't want to be too picky about it, but withEqual is just adding some extra, unnecessary, complexity to the code. As the method's javadoc states, it's better and advisable to simply pass the actual object. I would favor that to keep it simple.

PR #837, opened 2017-07-24
ISSUE-836: Fix identity function

arunmahadevan:
Renamed to `IDENTITY_FN` so that calcite does not treat it as a special keyword.

@harshach , pinging to see if this can be included in the maint release.

@shahsank3t , After applying this patch, the UI displays `IDENTITY_FN` in projection processor dropdown. UI should display the value from "displayName"


priyank5485:
+1 Will get it in to maint release @arunmahadevan

PR #845, opened 2017-07-25
ISSUE-844: Custom UDFs implemented as inner classes can't be registered/uploaded

mattf-horton:
This PR makes the following changes:

1. Fix the bug in ProxyUtil.java#L75 to use FqdnName instead of CanonicalName.
2. Add unit test capability for Custom UDF upload:
    - new file UDFCatalogResourceTest.java and new module custom-udf-microtest (provides the jar)
    - modified root pom.xml and streams/service/pom.xml to add test dependency (and fixed junit and jmockit dependencies)
    - UDFCatalogResource.java#processUdf made package-private for unit test access
    - StreamCatalogService.java#loadUdfsFromJar made static for unit test access when StreamCatalogService is mocked.
3. UDF.java: auto-convert to Canonical if the user inputs Fqdn form of classname (with '$').
4. Early check for concreteness: A lot of time is wasted and a lot of noise is generated in the logs processing interfaces from the SDK, included in the CustomUDF uber-jar, that meet the UDFsuperClasses.assignableFrom() test, but aren't of interest.  We shut these out by only fetching concrete (instantiable) classes.  Changed name of JarReader#findSubtypeOfClasses to findConcreteSubtypesOfClass.
5. A minor nuisance: streamline-dist/VERSION is a generated file not intended to be committed, but it is not currently in .gitignore .  Added it.


mattf-horton:
Note: in a couple files, IntelliJ simplified the java include lists automatically.  I reviewed the changes and left them in.

streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/service/StreamCatalogService.java
```diff
@@ -36,36 +32,12 @@
...
-import com.hortonworks.streamline.streams.catalog.TopologySource;
-import com.hortonworks.streamline.streams.catalog.TopologySourceStreamMapping;
-import com.hortonworks.streamline.streams.catalog.TopologyStream;
-import com.hortonworks.streamline.streams.catalog.TopologyTestRunCase;
-import com.hortonworks.streamline.streams.catalog.TopologyTestRunCaseSink;
-import com.hortonworks.streamline.streams.catalog.TopologyTestRunCaseSource;
-import com.hortonworks.streamline.streams.catalog.TopologyTestRunHistory;
-import com.hortonworks.streamline.streams.catalog.TopologyVersion;
-import com.hortonworks.streamline.streams.catalog.TopologyWindow;
+import com.hortonworks.streamline.streams.catalog.*;
```
HeartSaVioR: Same here.

streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/service/StreamCatalogService.java
```diff
@@ -78,19 +50,11 @@
...
 import com.hortonworks.streamline.streams.layout.storm.FluxComponent;
-import com.hortonworks.streamline.streams.rule.UDAF;
-import com.hortonworks.streamline.streams.rule.UDAF2;
-import com.hortonworks.streamline.streams.rule.UDF2;
-import com.hortonworks.streamline.streams.rule.UDF3;
-import com.hortonworks.streamline.streams.rule.UDF4;
-import com.hortonworks.streamline.streams.rule.UDF5;
-import com.hortonworks.streamline.streams.rule.UDF6;
-import com.hortonworks.streamline.streams.rule.UDF7;
+import com.hortonworks.streamline.streams.rule.*;
```
HeartSaVioR: Same here.

streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/service/StreamCatalogService.java
```diff
@@ -100,27 +64,13 @@
...
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.Set;
-import java.util.UUID;
+import java.util.*;
```
HeartSaVioR: Same here.

streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/service/StreamCatalogService.java
```diff
@@ -100,27 +64,13 @@
...
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import static com.hortonworks.streamline.common.ComponentTypes.NOTIFICATION;
-import static com.hortonworks.streamline.common.util.WSUtils.CURRENT_VERSION;
-import static com.hortonworks.streamline.common.util.WSUtils.buildEdgesFromQueryParam;
-import static com.hortonworks.streamline.common.util.WSUtils.buildEdgesToQueryParam;
-import static com.hortonworks.streamline.common.util.WSUtils.currentVersionQueryParam;
-import static com.hortonworks.streamline.common.util.WSUtils.versionIdQueryParam;
+import static com.hortonworks.streamline.common.util.WSUtils.*;
```
HeartSaVioR: Same here.

streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/service/StreamCatalogService.java
```diff
@@ -2411,7 +2361,7 @@ public UDF addUDF(UDF udf) {
         return udf;
     }
 
-    public Map<String, Class<?>> loadUdfsFromJar(java.io.File jarFile) throws IOException {
+    public static Map<String, Class<?>> loadUdfsFromJar(java.io.File jarFile) throws IOException {
```
HeartSaVioR: I know this doesn't require any fields in StreamCatalogService, but if we would want to make it static, maybe better to have utility class instead. Just a 2 cents.

streams/service/src/main/java/com/hortonworks/streamline/streams/service/UDFCatalogResource.java
```diff
@@ -318,7 +293,8 @@ public Response downloadUdf(@PathParam("udfId") Long udfId, @Context SecurityCon
         throw EntityNotFoundException.byId(udfId.toString());
     }
 
-    private void processUdf(InputStream inputStream,
+    /* package-private (for unit-test) */
+    void processUdf(InputStream inputStream,
```
HeartSaVioR: Let's add annotation @VisibleForTesting instead if Guava is available for this module.

streams/service/src/test/resources/custom-udf-microtest/pom.xml
```diff
@@ -0,0 +1,66 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <!-- Does not need a <parent> declaration -->
+
+    <groupId>com.hortonworks.streamline.test</groupId>
+    <artifactId>custom-udf-microtest</artifactId>
+    <version>0.1.0</version> <!--fixed version - update manually -->
```
HeartSaVioR: Do we have specific reason to avoid coupling version to the project version?

streams/service/src/test/resources/custom-udf-microtest/pom.xml
```diff
@@ -0,0 +1,66 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <!-- Does not need a <parent> declaration -->
```
HeartSaVioR: Wondering there's a reason to avoid coupling with other modules.

streams/service/src/test/resources/custom-udf-microtest/pom.xml
```diff
@@ -0,0 +1,66 @@
...
+
+    <properties>
+        <maven-shade-plugin.version>2.4.1</maven-shade-plugin.version>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.hortonworks.streamline</groupId>
+            <artifactId>streamline-sdk</artifactId>
+            <version>0.1.0-SNAPSHOT</version>
```
HeartSaVioR: Similar issue here. I think this is easy to be broken.

HeartSaVioR:
FYI: I pulled the change and ran the unit tests via `mvn clean install`, it succeed.

streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/service/StreamCatalogService.java
```diff
@@ -22,11 +22,7 @@
 import com.google.common.base.Function;
 import com.google.common.base.Joiner;
 import com.google.common.base.Preconditions;
-import com.google.common.collect.Collections2;
-import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Lists;
-import com.google.common.collect.Sets;
+import com.google.common.collect.*;
```
mattf-horton: I will revert the import wildcards, since they are contrary to Streamline team coding standards.  Thanks for informing me.

pom.xml
```diff
@@ -19,6 +19,11 @@
         <!--examples-->
         <module>examples/processors</module>
         <module>docker</module>
+        <!-- test jar builders -->
+        <module>streams/service/src/test/resources/custom-udf-microtest</module>
```
mattf-horton: @HeartSaVioR , thanks for the review.

Regarding the custom-udf-microtest module:  I agree the location of the jar-creation module seems hackey.  I was trying to achieve:
- Keep the source code for the test jar closely associated with the test jar itself (unlike the pre-built objects typically used for such purposes).
- But have the built jar end up under streams/service/src/test/resources/ without needing a copy phase.

The approach I took demonstrably works, and leaves the jar-creating source code closely tied to the jar itself, which I think is a big plus.  But it looks weird.

I'll re-organize it to a top-level module and see if you like it better.  We'll then be able to move it wherever you think it fits best.

streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/service/StreamCatalogService.java
mattf-horton: Hm.  It fits very well with the other methods in StreamCatalogService, it would seem artificial to remove it to a utility class.  Perhaps it is merely fortunate that it is possible to change it to static, and it is the only method I needed to access from the mocked class.  But in the interests of making only minimal changes to enable unit testing, I think it is best to do it this way.

streams/service/src/main/java/com/hortonworks/streamline/streams/service/UDFCatalogResource.java
mattf-horton: Good, thanks.

streams/service/src/test/resources/custom-udf-microtest/pom.xml
mattf-horton: Definitely yes.  If you give it a parent relationship, it tends to pull all the parent's dependencies into the jar (altho of course this can be worked around).  Also, as a test jar, it is intended to be as stand-alone as possible.

Finally, there's no benefit in declaring a parent, because it doesn't _need_ to inherit anything from streamline pom.  It's fine being just a sub-module for dependency-tree purposes.

streams/service/src/test/resources/custom-udf-microtest/pom.xml
mattf-horton: Okay, that would be one thing it could usefully inherit from a parent :-)
I'll see if the <minimizeJar> shading command prevents it from pulling in all the parent's dependencies, and if so go back to letting it have streamline pom as parent.

common/src/main/java/com/hortonworks/streamline/common/util/ProxyUtil.java
```diff
@@ -66,16 +66,16 @@ public O loadClassFromJar(String jarPath, String classFqdn)
...
         return Collections2.filter(classes, new Predicate<Class<?>>() {
             @Override
             public boolean apply(@Nullable Class<?> s) {
                 try {
-                    proxyUtil.loadClassFromJar(jarFile.getAbsolutePath(), s.getCanonicalName());
+                    proxyUtil.loadClassFromJar(jarFile.getAbsolutePath(), s.getName());
                     return true;
                 } catch (Throwable ex) {
-                    LOG.warn("class {} is subtype of {}, but it can't be initialized.", s, superTypeClass);
+                    LOG.warn("class {} is concrete subtype of {}, but can't be initialized due to exception:", s, superTypeClass, ex);
```
mattf-horton: No, the slf4j apis (since version 1.6.0) treat the last argument specially if it is a Throwable.  Please see https://www.slf4j.org/faq.html#paramException
Indeed, I first used a '{}' with the ex, and it didn't print the stack trace.  This does.

streams/service/src/main/java/com/hortonworks/streamline/streams/service/UDFCatalogResource.java
```diff
@@ -42,44 +42,19 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import javax.ws.rs.Consumes;
+import javax.ws.rs.*;
```
mattf-horton: Hi @arunmahadevan , thanks for the review.

As noted with @HeartSaVioR 's comments, I will remove all wildcard imports.

pom.xml
mattf-horton: Okay, that seems a good place to put them, I'll try that.  If I can't prevent it from including all the super-module dependencies, I'll have to put it at the top level.

streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/UDF.java
```diff
@@ -112,12 +112,20 @@ public void setType(Type type) {
...
         return className;
     }
 
+    /**
+     * Set inner-class names to canonical form if given fqdn form (with '$')
+     * @param className
+     */
     public void setClassName(String className) {
-        this.className = className;
+        this.className = className.replace('$', '.');
```
mattf-horton: Thanks @arunmahadevan .  I will add those tests, and instead of forcing to canonical form in the setClassName call, I'll force it to fqdn name in ProcessUdf():L313, right before updateTypeInfo().

streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/service/StreamCatalogService.java
HeartSaVioR: OK. Make sense.

streams/service/src/test/resources/custom-udf-microtest/pom.xml
mattf-horton: It turns out the <minimizeJar> configuration successfully prevents the test jar from inheriting the parent's dependencies, so I put test-support/custom-udf-microtest as submodules under streamline, and let it inherit the product version.
If folks want to similarly move source code for other test jars, like those in streams/service/src/test/resources/customprocessorupload, into test-support/, it's a good place that avoids dependency problems.

pom.xml
mattf-horton: Streamline-functions builds its own target jar, so it can't have a subsidiary module as a jar. 
 So I put the test jar sources under streamline/test-support/custom-udf-microtest.  

If folks want to similarly move source code for other test jars, like those in streams/service/src/test/resources/customprocessorupload, into test-support/, it's a good place that avoids dependency ordering problems.

streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/UDF.java
```diff
@@ -112,13 +112,16 @@ public void setType(Type type) {
...
+     * @return class name
+     */
     public String getClassName() {
         return className;
     }
 
-    public void setClassName(String className) {
-        this.className = className;
-    }
+    public void setClassName(String className) { this.className = className; }
```
HeartSaVioR: nit: we don't use one liner method definition.

HeartSaVioR:
+1 given that my review comments are all addressed.

arunmahadevan:
@mattf-horton , I am not sure why you are not able to build a test jar and add it as a dependency. Just trying to see if we can avoid adding new top level modules for testing.

For e.g.
1. In streams/functions/pom.xml
```
    <build>
        <plugins>
            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-jar-plugin</artifactId>
                <version>3.0.2</version>
                <executions>
                    <execution>
                        <goals>
                            <goal>test-jar</goal>
                        </goals>
                    </execution>
                </executions>
            </plugin>
        </plugins>
    </build>
```

2. Add a new UDF/UDAF that you intend to use in other modules under
streams/functions/src/test/java/com/hortonworks/streamline/streams/udaf/test/Foo.java

3. Now add the test jar as dependency elsewhere if you want to use those classes, e.g. in streams/service/pom.xml
```
        <dependency>
            <groupId>com.hortonworks.streamline</groupId>
            <artifactId>streamline-functions</artifactId>
            <type>test-jar</type>
            <version>0.1.0-SNAPSHOT</version>
            <scope>test</scope>
        </dependency>
```
4. If you plan to just refer to the jar it will be available under `streams/functions/target/streams/functions/target/streamline-functions-0.1.0-SNAPSHOT-tests.jar`

The `streamline-functions` module should be built before the module where you plan to use it.

streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/UDF.java
mattf-horton: Ah yes, fooled again by IntelliJ.  Fixed.  Thanks for pointing it out.

mattf-horton:
Hi @arunmahadevan , I'm open to doing it that way if you think it is best; however, I've worked on this and your suggestion to add testing to streams/runtime/storm/bolt/rules/FunctionsTest.java, and have the following observations:

1. In the spirit of unit testing, I wanted the CustomUDF upload test jar to ONLY have the contents of a few custom UDF implementations, and the needed streamline-sdk classes.  This means it should be in its own module. (It is probably possible to manipulate the maven assembly plugin to make it happen inside another module, but that is strongly dis-recommended by online sources.)  But if I add to streamline-functions test sources and pom in the way you suggest, the jar will contain all streamline-function test classes, which are unrelated.

    a. The difficulty I was referring to wasn't the mechanics of building a test jar, but the need to keep the CustomUDF jar sources in their own module while having them "under" another module.

2. Conversely, I'm uncomfortable putting CustomUDF implementations in the streamline-functions test sources, when they don't have anything to do with the other content in streamline-functions (which are all about the built-ins), and can't be used for unit testing in the streamline-functions module because even mocked uploading of a CustomUDF requires data structures from much farther downstream in the dependency chain.

3. Understand that you want to minimize the number of top-level modules, but Streamline really needs a place for the source code of test jars and other resources to go.  For example, all of these jars are checked into git as part of Streamline, but I couldn't find any associated sources:
```
./streams/runtime/src/test/resources/custom-split-join-lib.jar
./streams/service/src/test/resources/customprocessorupload/iotas-core.jar
./webservice/src/main/resources/node_modules/JSV/jsdoc-toolkit/java/classes/js.jar
./webservice/src/main/resources/node_modules/JSV/jsdoc-toolkit/jsdebug.jar
./webservice/src/main/resources/node_modules/JSV/jsdoc-toolkit/jsrun.jar
./webservice/src/test/resources/parser.jar
./webservice/src/test/resources/testnotifier.jar
```
The way I've done it provides a single top-level home for modules for all these jars to go, as well as the source code for the CustomUDF test jar.

4. It didn't work out to test CustomUDFs from streams/runtime/storm/bolt/rules/FunctionsTest.java.  While this is a very cool example of how to bring up a runnable RulesBolt with mocked dependencies, it appears it can only run with function classes already known in the JVM.  To upload a CustomUDF and make it referenceable, requires a StreamCatalogService, and I couldn't figure out how to make the RulesBolt reference one.  Harsha agreed it would be difficult to combine them.

5. Using my CustomUDF test jar, I do have a thorough unit test of the UDFCatalogResource#processUdf method, with mocked dependencies, to upload CustomUDFs.

I will do a manual end-to-end test, and fix or report back here any problems I find.

Hopefully I've responded adequately to your other suggestions.  Thanks.

arunmahadevan:
@mattf-horton , understand your concerns. I dont see other test classes being part of the uploaded Jar a major issue since the UDF's are loaded by class name. 

The concern here is adding a top level module for test-support, haven't seen such top level modules in other projects. If it was for something like integration tests it would make sense. Anyways I am ok if @HeartSaVioR, @harshach and others think its reasonable.

HeartSaVioR:
Streamline has a bit complicated module structure making contributors not easy to understand, but a one of implicit rule is that the modules which only bind to streamline are better to be placed under streams module. Hence streams/test-support/custom-udf-microtest feels better to me.

mattf-horton:
@HeartSaVioR , happy to make that change.  @arunmahadevan , is that satisfactory?

Regarding auto-test status, I merged latest master to pick up the addition of streamline-sql module.  That module has a compile error in PreparedStatementBuilder.java.  I'm not proposing the fix as part of my PR since I don't want to step on others' work.  But until that is fixed, master fails compilation (with or without my PR).  Unit tests pass on all other modules (by-passing storage-core), except for a dependency on storage in webservice.

mattf-horton:
By substituting jars into a real-world working HDF deployment, I've established that without this fix UDFs implemented as inner classes are not loaded, and with the fix they are successfully loaded.

@arunmahadevan , okay with you if I commit this?  Thanks.

arunmahadevan:
@mattf-horton , looks good to merge. Please squash the commits before merging.

arunmahadevan:
@mattf-horton, you want to squash and merge the changes to master ? let me know if you would like me to merge. Thanks.

mattf-horton:
Thanks, @arunmahadevan .

PR #850, opened 2017-07-26
ISSUE-840: Add standard SQL arithmetic functions in streamline

arunmahadevan:
Enabling the arithmetic functions supported by Caclite 1.11

Fixes #840

PR #852, opened 2017-07-28
ISSUE-851: Add missing string UDFs in streamline

### LINKED ISSUE #851: Add missing string UDFs in streamline
Expose builtin / add custom string functions similar to the ones supported in calcite.
https://calcite.apache.org/docs/reference.html#character-string-operators-and-functions

arunmahadevan:
1. Added Substring, Trim, Ltrim, Rtrim, Position, Overlay.
2. Made some backend changes to support overloaded UDFs

With the proposed changes there can be multiple entries (one per overload) in the UDF table (and the UDF api) for a single function. Some changes are required in the UI side to handle this and let the user select the function with the right number of arguments. cc @shahsank3t 

The function names of the streamline provided udfs are appendend with `_FN` to be consistent and avoid conflicts with the calcite defaults. That way when we expose the SQL editor, user can choose to use the streamline functions or the SQL standard ones. E.g. `SELECT SUBSTRING(stringField FROM 10) FROM inputStream` vs `SELECT SUBSTRING_FN(stringField, 10) FROM inputStream`. We need to provide some migration sql for the users who would have created rules with the older function names.

This includes changes from https://github.com/hortonworks/streamline/pull/850

Fixes #851


streams/functions/src/main/java/com/hortonworks/streamline/streams/udf/Rtrim.java
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * Trims trailing white spaces from the input.
+ * See {@link String#trim()}
+ */
+public class Rtrim implements UDF<String, String> {
+    @Override
+    public String evaluate(String input) {
+        int len = input.length();
+        while ((len > 0) && (input.charAt(len - 1) <= ' ')) {
```
HeartSaVioR: Same here for `input.charAt(len - 1) <= ' '`

streams/functions/src/main/java/com/hortonworks/streamline/streams/udf/Position2.java
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * Returns the position of the first occurrence of sub string in  a string starting the search from an index.
+ * <p>
+ * Note: The first position in string is 1 and if substring is not found within string, the function will return 0
+ * </p>
+ */
+public class Position2 implements UDF3<Integer, String, String, Integer> {
+    @Override
+    public Integer evaluate(String string, String sub, Integer fromIndex) {
+        return string.indexOf(sub, fromIndex) + 1;
```
HeartSaVioR: `fromIndex - 1` given that fromIndex also starts from 1.

streams/runners/storm/runtime/src/test/java/com/hortonworks/streamline/streams/runtime/storm/bolt/rules/FunctionsTest.java
```diff
@@ -63,6 +65,22 @@ public void testFunctions() throws Exception {
...
+    public void testStringFunctions2() throws Exception {
+        doTest(readFile("/streamline-udf2.json"), getTuple());
+        new Verifications() {
+            {
+                String streamId;
+                Tuple anchor;
+                List<List<Object>> tuples = new ArrayList<>();
+                mockCollector.emit(streamId = withCapture(), anchor = withCapture(), withCapture(tuples));
+                Assert.assertEquals(1, tuples.size());
+                Assert.assertEquals("SUBSTRING ello SUBSTRING2 hell POSITION 4 POSITION2 0 TRIM space LTRIM space  RTRIM  space OVERLAY abba OVERLAY2 abbaa",
```
HeartSaVioR: The result of POSITION2 should be 4 after addressing bug I commented. If you would want to test about 'none matched', set parameter to 5.

streams/service/src/main/java/com/hortonworks/streamline/streams/service/UDFCatalogResource.java
```diff
@@ -432,10 +432,11 @@ private void validateUDF(Set<String> udfs, UDF udf, boolean checkDuplicate) {
...
 
     private void checkDuplicate(UDF udf) {
-        Collection<UDF> existing = catalogService.listUDFs(
-                Collections.singletonList(new QueryParam(UDF.NAME, udf.getName())));
+        List<QueryParam> qps = QueryParam.params(UDF.NAME, udf.getName(), UDF.CLASSNAME, udf.getClassName(),
+                UDF.TYPE, udf.getType().toString());
+        Collection<UDF> existing = catalogService.listUDFs(qps);
         if (!existing.isEmpty()) {
-            LOG.warn("UDF with name {} already exists", udf.getName());
+            LOG.warn("UDF with same (name, classname, type) already exists", udf);
```
HeartSaVioR: Maybe missing `{}`? Regardless of that let's add space between  `exists` and udf.

streams/functions/src/main/java/com/hortonworks/streamline/streams/udf/Ltrim.java
```diff
@@ -0,0 +1,19 @@
...
+/**
+ * Trims leading white spaces from the input.
+ * See {@link String#trim()}
+ */
+public class Ltrim implements UDF<String, String> {
+    @Override
+    public String evaluate(String input) {
+        int len = input.length();
+        int st = 0;
+        while ((st < len) && (input.charAt(st) <= ' ')) {
```
arunmahadevan: For `TRIM` I return the value of `java.lang.String.trim` which trims `<= ' '`. So I wanted to be consistent for ltrim and rtrim. This has the advantage that tabs, newline, carriage return etc also gets trimmed.

Maybe we can keep this as the default and add overloaded methods that also accepts the character to trim.

streams/functions/src/main/java/com/hortonworks/streamline/streams/udf/Rtrim.java
arunmahadevan: For `TRIM` I return the value of `java.lang.String.trim` which trims `<= ' '`. So I wanted to be consistent for ltrim and rtrim. This has the advantage that tabs, newline, carriage return etc also gets trimmed.

Maybe we can keep this as the default and add overloaded methods that also accepts the character to trim.

streams/functions/src/main/java/com/hortonworks/streamline/streams/udf/Ltrim.java
arunmahadevan: Added TRIM2, LTRIM2, RTRIM2 that removes specified character.

streams/functions/src/main/java/com/hortonworks/streamline/streams/udf/Rtrim.java
arunmahadevan: Added TRIM2, LTRIM2, RTRIM2 that removes specified character.

streams/functions/src/main/java/com/hortonworks/streamline/streams/udf/Ltrim.java
HeartSaVioR: OK. I just found how String.trim is implemented.  http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/lang/String.java#l2852
Thanks for the pointer.

HeartSaVioR:
+1

Issue #853, opened 2017-08-01
Enable Application Test mode from UI


shahsank3t:
A couple of things need to be taken care of while enabling the test mode.

- [x] Test mode on saving the source data displays saved successfully multiple times

- [x] Test mode output displays nested data incorrectly

shahsank3t:
Raised PR https://github.com/hortonworks/streamline/pull/854

PR #854, opened 2017-08-01
Issue-853: Enable Application Test mode from UI

shahsank3t:
This fixes https://github.com/hortonworks/streamline/issues/853 and both the sub tasks mention in it.


harshach:
+1

PR #857, opened 2017-08-02
ISSUE-856 Event logging for test runner should log output events for …

### LINKED ISSUE #856: Event logging for test runner should log output events for processor
Current event logger for test runner logs events as below:
1. Source -> output events
2. Processor -> input events
3. Sink -> input events

which is based on separation of spout and bolt, but turns out that is not what we want.

We need to log events as below:
1. Source -> output events
2. Processor -> output events, there is no point in logging input events here as the previous components output events are input events for a processor.
3. Sink -> input events

HeartSaVioR:
…processor

* also add missing license doc to some classes

This closes #856

---
This change is [Reviewable](https://reviewable.io/reviews/hortonworks/streamline/857)

### DIFF
--- a/streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/testing/EventLoggingOutputCollector.java
+++ b/streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/testing/EventLoggingOutputCollector.java
@@ -1,0 +1,117 @@
+/**
+ * Copyright 2017 Hortonworks.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+
+ *   http://www.apache.org/licenses/LICENSE-2.0
+
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
+
+package com.hortonworks.streamline.streams.runtime.storm.testing;
+
+import org.apache.storm.task.OutputCollector;
+import org.apache.storm.tuple.Tuple;
+import org.apache.storm.utils.Utils;
+
+import java.util.Collection;
+import java.util.List;
+
+public class EventLoggingOutputCollector extends OutputCollector {
+    private final OutputCollector delegate;
+    private final OutputCollectorStreamlineEventLogger outputCollectorEventLogger;
+
+    public EventLoggingOutputCollector(OutputCollector delegate, String componentName, TestRunEventLogger eventLogger) {
+        // we simply ignore the _delegate in OutputCollector and override all of the methods
+        // this will work with subclass of OutputCollector since we only expose methods what we know about
+        super(null);
+        this.delegate = delegate;
+        this.outputCollectorEventLogger = new OutputCollectorStreamlineEventLogger(eventLogger, componentName);
+    }
+
+    @Override
+    public List<Integer> emit(String streamId, Tuple anchor, List<Object> tuple) {
+        return outputCollectorEventLogger.emitWithLoggingEvent(streamId, tuple, t -> delegate.emit(streamId, anchor, t));
+    }
+
+    @Override
+    public List<Integer> emit(String streamId, List<Object> tuple) {
+        return outputCollectorEventLogger.emitWithLoggingEvent(streamId, tuple, t -> delegate.emit(streamId, t));
+    }
+
+    @Override
+    public List<Integer> emit(Collection<Tuple> anchors, List<Object> tuple) {
+        return outputCollectorEventLogger.emitWithLoggingEvent(Utils.DEFAULT_STREAM_ID, tuple, t -> delegate.emit(anchors, t));
+    }
+
+    @Override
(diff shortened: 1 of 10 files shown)


HeartSaVioR:
@arunmahadevan 
1.
From the issue we clearly stated that input events should be recorded for sink, so I just follow the spec. You're also right about that we don't actually need to log events on sinks since it should be same as output events on upstream. Looks like just a matter of decision, and either is fine to me.
cc. @harshach @shahsank3t 

2. Yes right. That would be better to be consistent with both Spout and Bolt implementations. I'll create EventLoggingSpoutOutputCollector and apply similar approach.

HeartSaVioR:
Addressed 2, and also addressed missing spot (windowing bolt).

HeartSaVioR:
@arunmahadevan 
Reading your comment again, and being curious: we just replace Sources and Sinks with Test Sources and Test Sinks, hence we don't use actual Spouts and Bolts at all. We should leave the output result to history anyway, so can't remove that feature. Again it's just a matter of leaving event log or not.

streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/testing/EventLoggableOutputCollector.java
```diff
@@ -0,0 +1,41 @@
...
+
+package com.hortonworks.streamline.streams.runtime.storm.testing;
+
+import com.hortonworks.streamline.streams.StreamlineEvent;
+
+import java.util.List;
+import java.util.function.Consumer;
+import java.util.function.Function;
+
+public interface EventLoggableOutputCollector {
```
arunmahadevan: nit: this can be a changed to concrete class initialized with event logger, componentName and be a member inside the EventLoggingOutputCollector and EventLoggingSpoutOutputCollector. It looks like this interface is mostly to share a few methods internal to the two implementations.

arunmahadevan:
> Reading your comment again, and being curious: we just replace Sources and Sinks with Test Sources and Test Sinks, hence we don't use actual Spouts and Bolts at all. We should leave the output result to history anyway, so can't remove that feature. Again it's just a matter of leaving event log or not.

So now it would log the events from the Test sources and the actual processors right? Regarding sink I think we can remove the logging. Lets wait for inputs from @harshach and @shahsank3t

HeartSaVioR:
Yes right. Let's wait for inputs.

streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/testing/EventLoggableOutputCollector.java
HeartSaVioR: Yes the interface is intended to be used like a mix-up / trait. Two methods are needed from both EventLoggingOutputCollector and EventLoggingSpoutOutputCollector, and they're having different class hierarchy so I thought easiest way to add methods to both of them is using trait-like interface.

I don't have strong opinion of keeping this as it is. Just changed to what you suggested. That looks like better that they no longer expose component name and event logger to the public.

shahsank3t:
@HeartSaVioR @arunmahadevan @harshach 
Thinking out loud a scenario: 
When we have multiple rules in a rule processor and each rule associated with a different sink, the user might want to know which data was passed to which sink and thereby validate if he configured the rules properly or not.

I think it will be better if we log the incoming events of sinks as well to cover the above scenario as well as give consistent experience throughout source/processor/sink.

PS: We can indicate on UI that the events are output/input(or being dumped or saved in sink, using icons) so to clear confusion user has if any.

HeartSaVioR:
@shahsank3t 
Makes sense. In that case we should log the incoming events in sinks. 
Another approach may be adding target stream to the output event. We can add input stream to the input event, too. But this adds the thing to show, so it's just a 2 cents, and I'm OK with current.

shahsank3t:
@HeartSaVioR 
Do you mean to just add the input/output stream's name or the whole schema coming from it?

If it's the former, it's ok to show the stream name from where schema is coming but if it's latter, we need to revisit how to show data for processors (input schema & output schema) on UI.

HeartSaVioR:
@shahsank3t I mean former, but not just showing stream's name with value, but showing output value on edge instead of component (processor). Just 2 cents if we really want to show stream-aware output events.

shahsank3t:
@HeartSaVioR ah ok. It makes sense to show data on edges

@harshach What do you think about it?

harshach:
@HeartSaVioR @shahsank3t sorry I didn't get notification and didn't see the comments earlier.
I think showing at the edges in the DAG vs nodes makes sense.

HeartSaVioR:
@shahsank3t 
I've pushed the change. 

Now `a event` consists of (timestamp, event type, component name, stream ID, streamline event).
`event type` is either INPUT or OUTPUT, and it also represents what `stream ID` means. `input stream ID` for INPUT type, `output stream ID` for OUTPUT type.
In UI you may just show OUTPUT events on edges, or even show INPUT events on sinks too.

@arunmahadevan Could you review the last change? Thanks in advance.

HeartSaVioR:
@shahsank3t Kindly reminder.

shahsank3t:
+1 API response are as decided.
Working on the UI changes and will raise another PR for it so you can go ahead and merge this one.

HeartSaVioR:
Merged via 3452db146c6018f9d694dcb42e85c3ba91734ad9
(Github didn't recognize the merge and I have no idea why it did.)

PR #861, opened 2017-08-02
ISSUE-859: Migration script for UDF/UDAF

### LINKED ISSUE #859: Provide a migration script for UDF/UDAFs
The internal name for streamline provided UDF/UDAF are now appended with `_FN`. This is to migrate the data in the DB for any UDFs users may have used earlier with the older name.

arunmahadevan:
The internal name for streamline provided UDF/UDAF are now appended with _FN. This is to migrate the data in the DB and support UDFs users may have saved earlier with the older name until they save it again with the new logic.

Fixes #859


arunmahadevan:
The changes are addressed with https://github.com/hortonworks/streamline/pull/861.

PR #862, opened 2017-08-02
ISSUE-860 Add timeout option to topology test run

HeartSaVioR:
* while requesting topology test run, "durationSecs" is added to config json
  * the config is optional, and default value is 30 secs (up to runner implementation)

@shahsank3t 
This needs UI change that provides feature to let users input test run duration. 
UI can include the KV of config to the JSON: key name is "durationSecs", and unit is second.
(For type of value, you can just pick the same type as what you use to provide `testCaseId`.)
If the config doesn't exist in json, it will be run with default value (same as it is).


arunmahadevan:
+1

shahsank3t:
@HeartSaVioR 
Submitted PR: https://github.com/hortonworks/streamline/pull/865 for the UI side changes required for this PR.

PR #868, opened 2017-08-07
Integrating findbugs and fixing some of the reported issues.

arunmahadevan:
There are way too many issues, so for now I have fixed whatever I found to be real issues or easy to fix.  Some packages and issue types are excluded, but this can be taken up in future and fixed.


streams/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/rule/sql/SqlScript.java
```diff
@@ -142,7 +154,7 @@ public StreamlineEvent apply(Values values) {
                         return valuesConverter.convert(values, inputEvent);
                     }
                 });
-            } else {
+            } else if (result instanceof StreamlineEvent){
```
HeartSaVioR: nit: space between `)` and `{`

HeartSaVioR: And is it valid to return empty list? If we expect either valuesConverter is provided or result is a StreamlineEvent instance, let's throw exception for other cases.

streams/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/transform/MergeTransformRuntime.java
```diff
@@ -66,7 +66,12 @@ public String toString() {
 
         @Override
         public TransformRuntime create(Transform transform) {
-            return new MergeTransformRuntime((MergeTransform) transform);
+            if (transform instanceof MergeTransform) {
```
HeartSaVioR: This pattern is used from so many places. Could we extract a method for this pattern?

streams/service/src/main/java/com/hortonworks/streamline/streams/service/CustomProcessorUploadHandler.java
```diff
@@ -153,36 +154,20 @@ private InputStream getJarFile (CustomProcessorInfo customProcessorInfo, File ta
...
 
     private byte[] getFileAsByteArray (File tarFile, String fileName) {
-        BufferedInputStream bis = null;
-        TarArchiveInputStream tarArchiveInputStream = null;
         byte[] data = null;
-        try {
-            LOG.info("Getting file " + fileName + " from " + tarFile);
-            bis = new BufferedInputStream(new FileInputStream(tarFile));
-            tarArchiveInputStream = new TarArchiveInputStream(bis);
+        LOG.info("Getting file " + fileName + " from " + tarFile);
```
HeartSaVioR: `LOG.info("Getting file {} from {}", fileName, tarFile);`

HeartSaVioR:
And if findbugs also have max violation count, let's just set it and not exclude any packages. We may not have chance to include already excluded packages.

common/src/main/java/com/hortonworks/streamline/common/FileWatcher.java
```diff
@@ -79,7 +79,7 @@ public boolean processEvents() {
             WatchEvent<Path> ev = (WatchEvent<Path>) event;
             Path name = ev.context();
             Path child = dir.resolve(name);
-            LOG.info(String.format("%s: %s\n", event.kind().name(), child));
+            LOG.info(String.format("%s: %s%n", event.kind().name(), child));
```
arunmahadevan: 👍 Thanks. Since there were too many issues I was blindly following the recommendations and didn't dig deeper.

examples/processors/pom.xml
```diff
@@ -12,6 +12,17 @@
 
     <artifactId>streamline-examples-processors</artifactId>
     <packaging>jar</packaging>
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>findbugs-maven-plugin</artifactId>
+                <configuration>
+                    <excludeFilterFile>${project.basedir}/../../buildtools/findbugs-exclude.xml</excludeFilterFile>
```
arunmahadevan: Let me try this approach.

streams/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/rule/sql/SqlScript.java
arunmahadevan: if result is null we return an empty list. Inside `if (result != null)` can throw an exception in a final `else`

arunmahadevan:
@HeartSaVioR addressed review comments and changed the module structure as suggested.

>Btw, does findbugs cover checkstyle too? Some style issues seem to be also caught here.
Show outdated

It does not check for style issues. Only bug patterns.

>And if findbugs also have max violation count, let's just set it and not exclude any packages. We may not have chance to include already excluded packages.

I could not find an option specify the number of allowed bugs. Yes the package level excluding needs to be removed. Maybe we can exclude specific classes. This can be taken up in future.

PR #874, opened 2017-08-09
ISSUE-872 Fix bugs on feature to 'Test topology run' for integration test

### LINKED ISSUE #870: Removing topology should also remove topology test run histories
This triggers FK reference issue and makes topology deletion failing.

This also applies to HDF-3.0-maint, though we don't expose the feature "test topology run" in HDF-3.0.x version.

### LINKED ISSUE #871: TestRunProcessorBoltFluxComponent should respect underlying component's own config
In TestRunProcessorBoltFluxComponent, we also initialize underlying component so that it can be included to TestRunProcessorBolt. While initializing we use underlying component's own transformation class, but we miss to pass its own processor's config to initialized transformation instance.

### LINKED ISSUE #873: Change mysql schema for test case source, sink, histories: TEXT to MEDIUMTEXT
I found TEXT on MySQL only allow 64k. Note that it's not 64k characters and it will depend on database encoding.

MEDIUMTEXT will be OK for most of cases. PostgreSQL (and ANSI-SQL) only uses TEXT.

HeartSaVioR:
* ISSUE-870: Removing topology should also remove topology test run histories
* ISSUE-871: TestRunProcessorBoltFluxComponent should respect underlying component's own config
* ISSUE-873: Change mysql schema for test case source, sink, histories: TEXT to MEDIUMTEXT
* This closes #870
* This closes #871 
* This closes #873

---
This change is [Reviewable](https://reviewable.io/reviews/hortonworks/streamline/874)


harshach:
+1

Issue #849, opened 2017-07-26
Schema versioning and programmatic upgrade with a framework like liquibase


raju-saravanan:
Currently, all the necessary schema migrations are kept in one file. Instead of this approach, we should consider breaking down each SQL statements into their respective files and use a framework like liquibase to do SQL migration. With this approach, any new changes that have to be introduced to the SQL entities could be versioned and could be applied on top of an existing database depending upon how far behind the existing database from the latest version.

arunmahadevan:
Thanks @raju-saravanan , merged https://github.com/hortonworks/streamline/pull/861 to master

PR #882, opened 2017-08-21
ISSUE:849 Support schema versioning and migration using Flyway

raju-saravanan:
This pull request provides support for schema versioning and migration #849  and piggybacks @arunmahadevan  's pull request (https://github.com/hortonworks/streamline/pull/861) which provides migration scripts for UDF and UAF.


bootstrap/bootstrap-storage.sh
```diff
@@ -41,60 +40,57 @@ else
   JAVA="${JAVA_HOME}/bin/java"
 fi
 
-TABLE_INITIALIZER_MAIN_CLASS=com.hortonworks.streamline.storage.tool.TablesInitializer
```
arunmahadevan: Why dont we use the TabesInitializer itself to download the library ? Looking for patterns like below can be error prone.

bootstrap/bootstrap-storage.sh
```diff
@@ -41,60 +40,57 @@ else
...
-for file in "${BOOTSTRAP_DIR}"/lib/*.jar;
+if grep -E "^(\s)*db.type:(\s)*\"mysql\"(\s)*$" $CONFIG_FILE_PATH > /dev/null; then
```
arunmahadevan: This only works if theres a single line with `db.type: ` and no comments following mysql in the line. Better use existing logic to download. Also you don't need to capture 

`grep -E "^\s*db.type:\s*\"mysql\"" $CONFIG_FILE_PATH > /dev/null`

bootstrap/bootstrap-storage.sh
```diff
@@ -41,60 +40,57 @@ else
...
+  if [ ! -f ../libs/$MYSQL_JAR_PATTERN ] ; then
```
arunmahadevan: 1. This assumes you are inside bootstrap dir which may not be the case. 

Use `${BOOTSTRAP_DIR}/../libs/` 

2. This will fail if there are multiple files matching the glob

bootstrap/bootstrap-storage.sh
```diff
@@ -41,60 +40,57 @@ else
...
-
-function checkStorageConnection {
-    ${JAVA} -Dbootstrap.dir=$BOOTSTRAP_DIR  -cp ${CLASSPATH} ${TABLE_INITIALIZER_MAIN_CLASS} -m ${MYSQL_JAR_URL_PATH} -c ${CONFIG_FILE_PATH} -s ${SCRIPT_ROOT_DIR} --check-connection
-}
-
 function printUsage {
-    echo "USAGE: $0 [create|drop|check-connection|drop-create]"
+    cat <<-EOF
+USAGE: $0 [baseline|migrate|info|validate|clean]
+   baseline : Use this option, if the target database doesn't have DATABASE_CHANGE_LOG table and it already has some migrations.
```
arunmahadevan: Is this the option to be used during a new setup? If so name it "install" and start from "v0". I don't think we need to give option to install from some intermediate baseline option.

bootstrap/bootstrap-storage.sh
```diff
@@ -41,60 +40,57 @@ else
...
+              Ex:- Lets say we have a total of 3 migrations with the latest version of schema registry, the target database already has version 1 and 2 and we only need to apply migration 3.
+              We need to create a new DATABASE_CHANGE_LOG and mark migration 1 and 2 as applied, for that add a parameter "baselineVersion" under "db.migration" in streamline.yaml and set it to 2,
+              i.e that latest version of the target database. Invoke bootstrap-storage.sh script with baseline option. This will create DATABASE_CHANGE_LOG table and mark migration 2 and below are applied.
+              After this invoke bootstrap-storage.sh with migration option, this run migration 3 while ignoring migration 2 and 1.
+   migrate  : Runs the migration on the target database from the last checkpoint, if DATABASE_CHANGE_LOG is absent then creates the DATABASE_CHANGE_LOG table and runs all the migrations
```
arunmahadevan: maybe we should fail here if the db is clean (first install).

i.e. if db is clean, users should run "install"
if db already has some data users should invoke "migrate"

bootstrap/bootstrap-storage.sh
```diff
@@ -41,60 +40,57 @@ else
...
+   info     : Shows the status of all the migrations
+   validate : Validate if the all the migrations are applied on the target database
+   clean    : Drop all the tables in the target database
```
arunmahadevan: IMO, we should retain all the previous options with bootstrap-storage since users would be already familiar with those options. We only need a new option "migrate" (and maybe info/validate).

bootstrap/bootstrap-storage.sh
```diff
@@ -41,60 +40,57 @@ else
...
+    cat <<-EOF
+USAGE: $0 [baseline|migrate|info|validate|clean]
+   baseline : Use this option, if the target database doesn't have DATABASE_CHANGE_LOG table and it already has some migrations.
+              Ex:- Lets say we have a total of 3 migrations with the latest version of schema registry, the target database already has version 1 and 2 and we only need to apply migration 3.
+              We need to create a new DATABASE_CHANGE_LOG and mark migration 1 and 2 as applied, for that add a parameter "baselineVersion" under "db.migration" in streamline.yaml and set it to 2,
+              i.e that latest version of the target database. Invoke bootstrap-storage.sh script with baseline option. This will create DATABASE_CHANGE_LOG table and mark migration 2 and below are applied.
+              After this invoke bootstrap-storage.sh with migration option, this run migration 3 while ignoring migration 2 and 1.
+   migrate  : Runs the migration on the target database from the last checkpoint, if DATABASE_CHANGE_LOG is absent then creates the DATABASE_CHANGE_LOG table and runs all the migrations
+   info     : Shows the status of all the migrations
```
arunmahadevan: what would it print for a newly installed db ?

bootstrap/sql/postgresql/drop_tables.sql
```diff
@@ -1,44 +0,0 @@
-DROP TABLE IF EXISTS file CASCADE;
```
arunmahadevan: I assume flyway is going to drop the tables. Does it take care of dropping the child tables first?. I think we had some issues in the past when the order of "drop" was not correct.

conf/streamline-dev.yaml
```diff
@@ -88,6 +88,9 @@ storageProviderConfiguration:
 #    dataSource.url: "jdbc:mysql://localhost/streamline_db"
 #    dataSource.user: "streamline_user"
 #    dataSource.password: "streamline_password"
+#  db.migrate:
```
arunmahadevan: I dont think we should expose migration specific options in the config files. Users can mess up. These can be command line options to bootstrap or hard code to some reasonable values.

conf/streamline-dev.yaml
```diff
@@ -120,6 +126,7 @@ logging:
   loggers:
     # Sets the level for 'com.example.app' to DEBUG.
     com.hortonworks.streamline: DEBUG
+    org.apache.tomcat.jdbc.pool.PooledConnection: ERROR
```
arunmahadevan: is it related to the patch ? Why PooledConnection is separately handled? Root level can be set to ERROR instead.

webservice/pom.xml
```diff
@@ -173,6 +173,16 @@
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
         </dependency>
+        <dependency>
```
arunmahadevan: Why is it added in webservice module ? It should be part of the storage-tools module.

bootstrap/bootstrap-storage.sh
```diff
@@ -16,7 +16,6 @@
 # limitations under the License.
 
 
-# Resolve links - $0 may be a softlink
```
arunmahadevan: The comment should be above the `while [ -h "${PRG}" ]; do`

raju-saravanan: Restored the comment

bootstrap/bootstrap-storage.sh
raju-saravanan: Yeah, this will be error-prone. I will replace a subset of classes in tool to download MySQL dependencies. After that I can use StreamlineApplication class to trigger the migrations

bootstrap/bootstrap-storage.sh
raju-saravanan: "migrate" option takes care of fresh installation and also installing from a particular version. "baseline" option may be required in streamline considering the SQL scripts here are idempotent. I will remove "baseline"

conf/streamline-dev.yaml
raju-saravanan: flyway bundle asks for a DataSource instance where we have to mention the driver class, if the driver class is not mentioned,  driver class is inferred from URL of the database but it throws warnings to the console that the driver class is null, which could be misleading the user, so I have suppressed the warnings from PooledConnection.

webservice/pom.xml
raju-saravanan: Currently the migrations are invoked from the cli arguments passed to StreamlineApplication class, that is why dependency is added to webservice module.

bootstrap/bootstrap-storage.sh
raju-saravanan: Sure, will make the changes

bootstrap/bootstrap-storage.sh
raju-saravanan: flyway doesn't have any direct support to check if the db is empty, which would be necessary to  check if we run migration to throw an exception, I will have to find a work around for this

conf/streamline-dev.yaml
raju-saravanan: Sure, I will implement this

bootstrap/bootstrap-storage.sh
raju-saravanan: If the db is empty, then it would show the list of migrations available to run and status of all these migrations as pending (i.e, to be applied)

bootstrap/sql/postgresql/drop_tables.sql
raju-saravanan: flyway takes care of dropping dependent tables.

raju-saravanan:
@arunmahadevan I have made the following changes :
  * restore "create" option (for fresh installation) which only works on empty database
  * "migrate" works only on existing database, throws exception if the database is empty
  * moved away from flyway dropwizard plugin and restored "tool" sub module but made changes to 
     it to use flyway internally.
  * exposed few more options to bootstrap script like "info", "validate", "repair".

bootstrap/bootstrap-storage.sh
```diff
@@ -49,41 +49,42 @@ done
...
+    execute ${opt}
     ;;
-drop-create)
-    dropTables && createTables
+drop-create )
+    execute drop && execute create
     ;;
-check-connection)
-    checkStorageConnection
+check-connection )
```
arunmahadevan: why check-connection is separate? Can it be part of "execute ${opt}" above?

bootstrap/bootstrap-storage.sh
```diff
@@ -49,41 +49,42 @@ done
...
+opt="$1"
+
 case "${opt}" in
-create)
-    createTables
-    ;;
-drop)
-    dropTables
+create | drop | migrate | info | validate | repair )
+    execute ${opt}
```
arunmahadevan: you can quote the string `execute "${opt}"`

bootstrap/bootstrap-storage.sh
```diff
@@ -49,41 +49,42 @@ done
...
     ;;
-drop-create)
-    dropTables && createTables
+drop-create )
+    execute drop && execute create
```
arunmahadevan: quote "drop" and "create"

storage/tool/pom.xml
```diff
@@ -11,6 +11,10 @@
 
     <artifactId>storage-tool</artifactId>
 
+    <properties>
+        <slf4j.version>1.7.12</slf4j.version>
```
arunmahadevan: this should be in root pom.xml "dependencyManagement" . We have already defined this https://github.com/hortonworks/streamline/blob/master/pom.xml#L55

storage/tool/pom.xml
```diff
@@ -38,9 +42,25 @@
             <artifactId>phoenix-core</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>${slf4j.version}</version>
```
arunmahadevan: remove this line and add to dependency management if required.

storage/tool/pom.xml
```diff
@@ -38,9 +42,25 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-simple</artifactId>
+            <version>${slf4j.version}</version>
```
arunmahadevan: remove this line and add to dependency management if required.

storage/tool/pom.xml
```diff
@@ -38,9 +42,25 @@
...
+        </dependency>
+        <dependency>
             <groupId>org.postgresql</groupId>
             <artifactId>postgresql</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.flywaydb</groupId>
+            <artifactId>flyway-maven-plugin</artifactId>
+            <version>4.2.0</version>
```
arunmahadevan: remove this line and add to dependency management

storage/tool/src/main/java/com/hortonworks/streamline/storage/tool/SchemaMigrationHelper.java
```diff
@@ -0,0 +1,111 @@
...
+
+    private Flyway flyway;
+
+    public SchemaMigrationHelper(Flyway flyway) {
+        this.flyway = flyway;
+    }
+
+    private void create() throws SQLException {
+        if(!validateDatabase(true))
+            throw new SchemaMigrationException("Target database is not empty, please consider executing with \"migrate\" option");
```
arunmahadevan: "Please use an empty database or use "migrate" if you are already running a previous version.

storage/tool/src/main/java/com/hortonworks/streamline/storage/tool/SchemaMigrationHelper.java
```diff
@@ -0,0 +1,111 @@
...
+        flyway.migrate();
+    }
+
+    private void migrate() throws SQLException {
+        if(!validateDatabase(false))
+            throw new SchemaMigrationException("Target database is empty, please consider executing with \"create\" option");
```
arunmahadevan: Target database is empty and no previous versions found, use the "create" option.

storage/tool/src/main/java/com/hortonworks/streamline/storage/tool/SchemaMigrationHelper.java
```diff
@@ -0,0 +1,111 @@
...
+
+    private void clean() {
+        flyway.clean();
+    }
+
+    private boolean checkConnection() throws SQLException {
+        try (Connection connection = flyway.getDataSource().getConnection()) {
+            return true;
+        } catch (FlywaySqlException e) {
+            System.out.println("Connection failure: " + e.getMessage());
```
arunmahadevan: use LOG.

If you are trying to display error in the console use stderr.

storage/tool/src/main/java/com/hortonworks/streamline/storage/tool/SchemaMigrationHelper.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    private void repair() {
+        flyway.repair();
+    }
+
+    private boolean isDatabaseEmpty(Connection connection) throws SQLException {
+        DatabaseMetaData databaseMetaData = connection.getMetaData();
+        try (ResultSet resultSet = databaseMetaData.getTables(connection.getCatalog(), connection.getSchema(), "", null)) {
+            if (resultSet.next())
```
arunmahadevan: conditional could be clearer here.

storage/tool/src/main/java/com/hortonworks/streamline/storage/tool/SchemaMigrationHelper.java
```diff
@@ -0,0 +1,111 @@
...
+                return false;
+            return true;
+        } catch (SQLException e) {
+            throw new SchemaMigrationException("Unable the obtain the state of the target database", e);
+        }
+    }
+
+    private boolean validateDatabase(Boolean checkDatabaseEmpty) throws SQLException {
```
arunmahadevan: I dont see any validations happening here. If this is to have a single method to check if database is empty or not I think we can just have a `isDatabaseEmpty()` that returns true/false.

storage/tool/src/main/java/com/hortonworks/streamline/storage/tool/SchemaMigrationHelper.java
```diff
@@ -0,0 +1,111 @@
...
+                break;
+            case VALIDATE:
+                validate();
+                break;
+            case DROP:
+                clean();
+                break;
+            case CHECK_CONNECTION:
+                if(!checkConnection()) {
+                    System.exit(1);
```
arunmahadevan: why exit here ?

bootstrap/bootstrap-storage.sh
```diff
@@ -32,7 +32,7 @@ done
 BOOTSTRAP_DIR=`dirname ${PRG}`
 CONFIG_FILE_PATH=${BOOTSTRAP_DIR}/../conf/streamline.yaml
 MYSQL_JAR_URL_PATH=https://dev.mysql.com/get/Downloads/Connector-J/mysql-connector-java-5.1.40.zip
-SCRIPT_ROOT_DIR="${BOOTSTRAP_DIR}/sql"
+SCRIPT_ROOT_DIR="${BOOTSTRAP_DIR}/sql/"
```
raju-saravanan: Ok, I will remove this.

bootstrap/bootstrap-storage.sh
```diff
@@ -49,41 +49,42 @@ done
...
-    ${JAVA} -Dbootstrap.dir=$BOOTSTRAP_DIR  -cp ${CLASSPATH} ${TABLE_INITIALIZER_MAIN_CLASS} -m ${MYSQL_JAR_URL_PATH} -c ${CONFIG_FILE_PATH} -s ${SCRIPT_ROOT_DIR} --check-connection
+function execute {
+    ${JAVA} -Dbootstrap.dir=$BOOTSTRAP_DIR  -cp ${CLASSPATH} ${TABLE_INITIALIZER_MAIN_CLASS} -m ${MYSQL_JAR_URL_PATH} -c ${CONFIG_FILE_PATH} -s ${SCRIPT_ROOT_DIR} --${1}
 }
 
 function printUsage {
-    echo "USAGE: $0 [create|drop|check-connection|drop-create]"
+    cat <<-EOF
+    USAGE: $0 [create|migrate|info|validate|drop|repair|check-connection]
+       create   : Applies all the migrations from scratch, the target database has to be empty for this option
```
raju-saravanan: Done

bootstrap/bootstrap-storage.sh
```diff
@@ -49,41 +49,42 @@ done
...
+       migrate  : Applies all the migration on the target database from the last check point, the target database should not be empty for this option
```
raju-saravanan: Done

bootstrap/bootstrap-storage.sh
raju-saravanan: Done

bootstrap/bootstrap-storage.sh
raju-saravanan: Done

bootstrap/bootstrap-storage.sh
raju-saravanan: Depending upon the exist status of the execute command, we print a meaningful message to the console. We can make the exception throw by the Tableinitializer class more readable and move this option to the first case condition.

arunmahadevan:
from the git root directory,

```
$ ./bootstrap/bootstrap-storage.sh check-connection
Using Configuration file: ./bootstrap/../conf/streamline.yaml
Downloading mysql jar from url: https://dev.mysql.com/get/Downloads/Connector-J/mysql-connector-java-5.1.40.zip
Downloading file mysql-connector-java-5.1.40.zip into /var/folders/sb/v_d6sh615q18cdfw4jz4nnd40000gn/T/
Copying mysql libraries into lib dir...
Unzipping downloaded mysql driver and copying
Copying file to libs /Users/aiyer/github/streamline/./bootstrap/lib/../../libs/mysql-connector-java-5.1.40-bin.jar
Failed to copy mysql driver into ./bootstrap/lib and /Users/aiyer/github/streamline/./bootstrap/lib/../../libs/
Exception in thread "main" org.flywaydb.core.api.FlywayException: Unable to instantiate JDBC driver: com.mysql.cj.jdbc.Driver : com.mysql.cj.jdbc.Driver
	at org.flywaydb.core.internal.util.jdbc.DriverDataSource.<init>(DriverDataSource.java:133)
	at org.flywaydb.core.Flyway.setDataSource(Flyway.java:821)
	at com.hortonworks.streamline.storage.tool.FlywayFactory.get(FlywayFactory.java:44)
	at com.hortonworks.streamline.storage.tool.TablesInitializer.main(TablesInitializer.java:153)
Caused by: java.lang.ClassNotFoundException: com.mysql.cj.jdbc.Driver
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:335)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	at java.lang.Class.forName0(Native Method)
	at java.lang.Class.forName(Class.java:348)
	at org.flywaydb.core.internal.util.ClassUtils.instantiate(ClassUtils.java:59)
	at org.flywaydb.core.internal.util.jdbc.DriverDataSource.<init>(DriverDataSource.java:123)
	... 3 more
Connection check failed.
```

But it copies the downloaded file to bootstrap/lib and if I run "./bootstrap/bootstrap-storage.sh check-connection" again it succeeds.  After that the other options work. Can you take a look?

storage/tool/pom.xml
raju-saravanan: Done

storage/tool/pom.xml
raju-saravanan: Done

storage/tool/pom.xml
raju-saravanan: Done

storage/tool/pom.xml
raju-saravanan: Done

storage/tool/src/main/java/com/hortonworks/streamline/storage/tool/SchemaMigrationHelper.java
raju-saravanan: Done

storage/tool/src/main/java/com/hortonworks/streamline/storage/tool/SchemaMigrationHelper.java
raju-saravanan: Done

storage/tool/src/main/java/com/hortonworks/streamline/storage/tool/SchemaMigrationHelper.java
raju-saravanan: Changed this to stderr here

storage/tool/src/main/java/com/hortonworks/streamline/storage/tool/SchemaMigrationHelper.java
raju-saravanan: if the check connection fails, the code should exit with a failure, that why i am exiting here

storage/tool/src/main/java/com/hortonworks/streamline/storage/tool/SchemaMigrationHelper.java
raju-saravanan: I have put a comment above if statement, to say what we are checking for

raju-saravanan:
I am not able to reproduce this issue in my local

~/Desktop/hortonworks-streamline-0.1.0-SNAPSHOT ᐅ ./bootstrap/bootstrap-storage.sh check-connection
Using Configuration file: ./bootstrap/../conf/streamline.yaml
Downloading mysql jar from url: https://dev.mysql.com/get/Downloads/Connector-J/mysql-connector-java-5.1.40.zip
Downloading file mysql-connector-java-5.1.40.zip into /var/folders/dh/lwlnbwlj2cx6gr2zz0nr0__40000gn/T/
Copying mysql libraries into lib dir...
Unzipping downloaded mysql driver and copying
Copying file to libs /Users/sraju/Desktop/hortonworks-streamline-0.1.0-SNAPSHOT/./bootstrap/lib/../../libs/mysql-connector-java-5.1.40-bin.jar
mysqlJarFile ./bootstrap/lib/mysql-connector-java-5.1.40-bin.jar
Thu Aug 24 16:18:34 IST 2017 WARN: Establishing SSL connection without server's identity verification is not recommended. According to MySQL 5.5.45+, 5.6.26+ and 5.7.6+ requirements SSL connection must be established by default if explicit option isn't set. For compliance with existing applications not using SSL the verifyServerCertificate property is set to 'false'. You need either to explicitly disable SSL by setting useSSL=false, or set useSSL=true and provide truststore for server certificate verification.
Connection check successful


This seems like a one off issue, something must have gone wrong while coping the jar to lib directory. Since the stacktrace is not printed by the code when the exception was thrown, I can't further explore on this. I have made changes to the code to print the stacktrace so that in future if we run into similar issue we have a look.

arunmahadevan:
@raju-saravanan , i am running it from github parent directory of SAM, not from installed location.

raju-saravanan:
@arunmahadevan addressed all your comments, can you have another look?

storage/tool/src/main/java/com/hortonworks/streamline/storage/tool/SchemaMigrationHelper.java
arunmahadevan: It may be better to exit from one place (e.g. main) that having the helper invoking System.exit. execute can return true/false to indicate if the operation succeeded or not.

arunmahadevan:
Overall LGTM +1.

storage/tool/src/main/java/com/hortonworks/streamline/storage/tool/SchemaMigrationHelper.java
raju-saravanan: Alternatively, can I throw an exception from checkConnection() if the checkConnection fails? In that case I don't have to do System.exit(1).

arunmahadevan: in that case will it print the exception in the console? If so it may not be desirable.

raju-saravanan: Done.

PR #883, opened 2017-08-24
ISSUE-856 Event logging for test runner should log output events

sameer79:


webservice/src/main/resources/app/scripts/containers/Streams/TestRunComponents/TestSourceNodeModal.jsx
```diff
@@ -179,7 +179,7 @@ class TestSourceNodeModal extends Component{
     const {showInputError,sourceNodeArr, testName} = this.state;
     let validate = false,validationArr = [];
     _.map(sourceNodeArr, (source,i) => {
-      if(source.records === '' || source.repeatTime === '' || source.repeatTime === undefined || source.records === undefined || testName === ''){
+      if(source.records === '' || source.repeatTime === '' || (source.repeatTime === undefined || source.repeatTime === 0 )|| source.records === undefined || testName === ''){
```
shahsank3t: It's ok to have 0 as Repeat when the user just wants to run the test on the provided record.
Let's not restrict the user to enter more than 0 as value.

HeartSaVioR: Maybe restricting it and set default value to 1 would be clearer.

shahsank3t:
+1 after that nit suggestion is addressed.

sameer79:
Address the suggestion

PR #886, opened 2017-09-01
BUG-87346 Test mode: few minor UI issues

sameer79:
1] Clicking on test case itself test case dialog is shown to configure OR edit.
2] while double clicking on the source, processors, and sink the node configuration is shown in non-editable mode.


shahsank3t:
+1

PR #888, opened 2017-09-04
ISSUE-887 Recording events from sink leads the result confused

### LINKED ISSUE #887: Recording events from sink leads the result confused
Looks like recording events from sink leads the result confused.

Belows are same events recorded from both processor and sink:
```
{"timestamp":1504248617938,"eventType":"OUTPUT","componentName":"RULE","streamId":"rule_notifier_stream_3","targetComponentName":"NOTIFICATION","streamlineEvent":{"subject":"Testing","from":"aiyer@...","to":"aiyer@...","body":"userid: PUFPaY9KxDAcGqfsorJp2R stars: 1","contentType":"text/html"}}
```
```
{"timestamp":1504248617939,"eventType":"INPUT","componentName":"NOTIFICATION","streamId":"rule_notifier_stream_3","targetComponentName":"RULE","streamlineEvent":{"subject":"Testing","from":"aiyer@...","to":"aiyer@...","body":"userid: PUFPaY9KxDAcGqfsorJp1R stars: 1","contentType":"text/html"}}
```

These are INPUT type events so the event log itself is correct, but field name (targetComponentName) looks like not correct for INPUT type.
Maybe we could just get rid of events in sink, given that output events which will go to sink and input events in sink is redundant.

HeartSaVioR:
This closes #887

* just remove logging events from Sink

Change-Id: I5bfec56e4ac151e1056a59f90faca30f3b7f53d4

### DIFF
--- a/streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/testing/TestRunSinkBolt.java
+++ b/streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/testing/TestRunSinkBolt.java
@@ -42,7 +42,6 @@ public class TestRunSinkBolt extends BaseRichBolt {
     private transient ObjectMapper objectMapper;
 
     private final TestRunSink testRunSink;
-    private transient TestRunEventLogger eventLogger;
 
     public TestRunSinkBolt(String testRunSinkJson) {
         this(Utils.createObjectFromJson(testRunSinkJson, TestRunSink.class));
@@ -72,19 +71,12 @@ public class TestRunSinkBolt extends BaseRichBolt {
             LOG.error("Can't open file for preparing to write: " + outputFilePath);
             throw new RuntimeException(e);
         }
-
-        eventLogger = TestRunEventLogger.getEventLogger(testRunSink.getEventLogFilePath());
     }
 
     @Override
     public void execute(Tuple input) {
         Object value = input.getValueByField(StreamlineEvent.STREAMLINE_EVENT);
-        String sourceComponentId = input.getSourceComponent();
-        String streamId = input.getSourceStreamId();
         StreamlineEvent event = (StreamlineEvent) value;
-
-        eventLogger.writeEvent(System.currentTimeMillis(), TestRunEventLogger.EventType.INPUT, testRunSink.getName(),
-                streamId, StormTopologyUtil.extractStreamlineComponentName(sourceComponentId), event);
 
         String outputFilePath = testRunSink.getOutputFilePath();
         try (FileWriter fw = new FileWriter(outputFilePath, true)) {
--- a/streams/actions/src/main/java/com/hortonworks/streamline/streams/actions/topology/service/TopologyTestRunner.java
+++ b/streams/actions/src/main/java/com/hortonworks/streamline/streams/actions/topology/service/TopologyTestRunner.java
@@ -147,7 +147,7 @@ public class TopologyTestRunner {
         Map<String, TestRunSink> testRunSinkMap = sinks.stream()
                 .collect(toMap(s -> s.getName(), s -> {
                     String uuid = UUID.randomUUID().toString();
-                    TestRunSink testRunSink = new TestRunSink(getTopologyTestRunResult(uuid), eventLogFilePath);
+                    TestRunSink testRunSink = new TestRunSink(getTopologyTestRunResult(uuid));
                     testRunSink.setName(s.getName());
                     return testRunSink;
                 }));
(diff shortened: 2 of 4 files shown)


HeartSaVioR:
@shahsank3t @arunmahadevan 
This removes redundant events which one is from processor (output event) and another one is from sink (input event). Once we have stream and target component in event information, event logging in sink is not strictly required. 
Please review this. Thanks in advance!

shahsank3t:
@HeartSaVioR 
Assuming we have two streams coming into HBase sink, shouldn't we be showing the sink's logging events which will be an aggregate data of those two streams? It helps users to get directly complete data being stored in the sink rather than wondering and manually analyzing himself.

HeartSaVioR:
@shahsank3t 
I don't think it's required, but I see the benefit of keeping it as it is. Event log in text is still a bit confused though.

Otherwise, how about changing field name in event like below: 
* `targetComponentName` -> `outputComponentName` for output event and `inputComponentName` for input event
* `streamId` -> `outputStreamId` for output event and `inputStreamId` for input event

given that the name `targetComponentName` and `streamId` is not clear covering both input event and output event?

@arunmahadevan What do you think?

shahsank3t:
Yes, I am good with keeping the logs and just changing the field names like you suggested.

arunmahadevan:
I dont think we really need to log the events at the sink because it will just duplicate the events. Right now I see a small dot on each edges as I click the events in the event log box.

One issue I see is the events are interleaved. E.g. If Source has two outgoing edges, I see events alternating between the edges. Maybe we need to sort it by (source, target) so that we see all the events that fall on edge-1 followed by edge-2?

Heres the example I tried.

screen shot 2017-09-06 at 10 10 13 am

[The downloaded event log file](https://github.com/hortonworks/streamline/files/1279797/out.txt)

arunmahadevan:
>Assuming we have two streams coming into HBase sink, shouldn't we be showing the sink's logging events which will be an aggregate data of those two streams? 

Weren't we just logging the input to the sinks earlier which would be the output of its source component ? In that case there will be two duplicate events.

If there are two inputs to a sink, even now I can see the events on each input to the sink.

HeartSaVioR:
Suppose the topology DAG: `A -> B -> C -> D`. When I run a test with some input events to A, I don't mind which events are flowing through B and C, unless I don't see the expected result from D. For me, events in B and C are for only debugging purpose.

Btw, I don't mind removing sink events because we already stored events which sinks are receiving to test run history entity.

shahsank3t:
@arunmahadevan @HeartSaVioR 
Ah yes, I got it wrong before. I understand now logging events at sink will be duplicated, it's better to get rid of those. +1 for the same from my side :)

shahsank3t:
@HeartSaVioR 
> Suppose the topology DAG: A -> B -> C -> D. When I run a test with some input events to A, I don't mind which events are flowing through B and C, unless I don't see the expected result from D. For me, events in B and C are for only debugging purpose.

We have a small dot on the edge connecting C and D, on clicking it, will show the expected result for D (or can say the output from C), since sinks don't perform any operations on schema it will be redundantly stored if logged.
Correct if I'm wrong here @arunmahadevan

HeartSaVioR:
Suppose the case a sink receiving multiple streams:

```
A -> B -> C  -> D -> F
                    |-> E /
```

In this case, which events F receives are not aggregated, and when we want to see which events F gets, we need to check D -> F and E -> F separately.

Like I said, a history entity has the result which are events which all sinks receive, so we could show the result of history if we really want to, so I don't mind.

HeartSaVioR:
Assuming we agree the direction, could we go on reviewing the code?

streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/testing/TestRunSinkBolt.java
```diff
@@ -72,20 +71,13 @@ public void prepare(Map stormConf, TopologyContext context, OutputCollector coll
...
         Object value = input.getValueByField(StreamlineEvent.STREAMLINE_EVENT);
-        String sourceComponentId = input.getSourceComponent();
-        String streamId = input.getSourceStreamId();
         StreamlineEvent event = (StreamlineEvent) value;
 
-        eventLogger.writeEvent(System.currentTimeMillis(), TestRunEventLogger.EventType.INPUT, testRunSink.getName(),
-                streamId, StormTopologyUtil.extractStreamlineComponentName(sourceComponentId), event);
-
         String outputFilePath = testRunSink.getOutputFilePath();
         try (FileWriter fw = new FileWriter(outputFilePath, true)) {
```
arunmahadevan: Seems we are writing the events to this file in the sink. Why is it required?

HeartSaVioR: This is for aggregating the result of sinks and storing the aggregated result to test history entity. We may still want to sort out features for test run and remove unnecessary features.

PR #890, opened 2017-09-04
ISSUE-889 Retain Window configuration when running test topology with…

### LINKED ISSUE #889: Window configuration is lost when running test topology with WindowedRule
When topology has WindowedRule which proceeds event time window, WindowRuleBoltFluxComponent reads the rules and set `withWindowConfig` to WindowRulesBolt accordingly. TestRunWindowProcessorBolt simply ignores it so it is always running with processing time window.

There's a method `getTimestampExtractor` in IWindowedBolt which open the way to provide timestamp extractor, and setup of WindowRulesBolt is for that method, so we can just override the method in TestRunWindowProcessorBolt and return delegate's return value of getTimestampExtractor.

Thanks @arunmahadevan for reporting and guiding a way to fix.

HeartSaVioR:
… WindowedRule

* also loosen type checking with timestamp extractor
  * it didn't work with test topology with short value of field (not strictly long)

This closes #889


streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/bolt/StreamlineTimestampExtractor.java
```diff
@@ -18,9 +18,14 @@ public StreamlineTimestampExtractor(String fieldName) {
     public long extractTimestamp(Tuple tuple) {
         StreamlineEvent event = (StreamlineEvent) tuple.getValueByField(StreamlineEvent.STREAMLINE_EVENT);
         Object ts = event.get(fieldName);
-        if (ts == null || !(ts instanceof Long)) {
+        if (ts == null) {
+            throw new IllegalArgumentException("Streamline event does not contain field: " + fieldName);
+        }
+
+        if (!(ts instanceof Number)) {
```
arunmahadevan: The timestamp has to be "milliseconds since January 1, 1970 UTC", so by putting values other than long (byte, short, int) we cannot represent dates beyond Jan 1970. So IMO, better to retain the `instanceof Long` and may be we can explicitly state something like `Streamline event does not contain a long value in the field` in the exception. 
Recently there was a case where the customer was passing seconds (instead of millis) in a long field which was causing issues. Will be great if we can catch that also. E.g. If value is less than 2 Billion log a WARNING.

arunmahadevan: Ok missed your comment
> it didn't work with test topology with short value of field (not strictly long)

Is it due to the way the values are parsed out of the test input JSON? Still think we should restrict to "long" type for production use cases.

HeartSaVioR: Yes we just read the json via JSON parser without validating/converting to conform the schema. If we want to maintain this to only allow long (type), schema validator/converter maybe needed as well.

arunmahadevan: Ok in the real mode we allow only long types (based on schema) to be selected as timestamp fields. I am not sure if we should relax the real mode types to allow any Number types as a workaround for test mode.  I think this may be an issue in general for other processors that receives number types as well.

In the test spout is it possible to do some fix to promote/cast the byte/short/int to long (or the actual schema type) based on the output schema of the source ?

HeartSaVioR: @arunmahadevan Yes I just reverted the change from StreamlineTimestampExtractor and implemented converter which converts the value being conformed to the type of field.

HeartSaVioR:
@arunmahadevan 
Addressed review comment. You can refer the last commit to narrow the diff. I'll squash the commits once you approve the patch.

common/src/main/java/com/hortonworks/streamline/common/SchemaValueConverter.java
```diff
@@ -0,0 +1,146 @@
...
+
+            case ARRAY:
+                if (value instanceof List) {
+                    return value;
+                }
+                break;
+
+        }
+
+        throw new IllegalArgumentException("Cannot convert value " + value + " with Java type class " +
```
arunmahadevan: Can this be moved under the default case?

common/src/main/java/com/hortonworks/streamline/common/SchemaValueConverter.java
```diff
@@ -0,0 +1,146 @@
...
+package com.hortonworks.streamline.common;
+
+import com.hortonworks.registries.common.Schema;
+
+import java.util.List;
+import java.util.Map;
+
+import static java.util.stream.Collectors.toMap;
+
+public final class SchemaValueConverter {
```
arunmahadevan: I think this is a nice functionality and can be reused in other places. Can you add some java doc on what this class does?

common/src/main/java/com/hortonworks/streamline/common/SchemaValueConverter.java
```diff
@@ -0,0 +1,146 @@
...
+    private SchemaValueConverter() {
+    }
+
+    public static Map<String, Object> convertMap(Schema schema, Map<String, Object> value) {
```
arunmahadevan: I think this is a nice functionality and can be reused in other places. Can you add some java doc on what this method does?

common/src/main/java/com/hortonworks/streamline/common/SchemaValueConverter.java
```diff
@@ -0,0 +1,146 @@
...
+        Map<String, Schema.Type> fieldToType = schema.getFields().stream()
+                .collect(toMap(Schema.Field::getName, Schema.Field::getType));
+
+        boolean fieldNotFound = (schema.getFields().size() != value.keySet().size()) ||
```
arunmahadevan: A `Field` can be optional. Should we account for that here ?

common/src/main/java/com/hortonworks/streamline/common/Schema.java
```diff
@@ -1,686 +0,0 @@
...
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.stream.Collectors;
-
-//TODO Make this class Jackson Compatible.
-public class Schema implements Serializable {
```
arunmahadevan: Nice. We need to raise an issue and remove any other such classes which has already been moved to registry repo.

common/src/main/java/com/hortonworks/streamline/common/SchemaValueConverter.java
HeartSaVioR: There're some `cases` which uses the line, so can't move to the default case. Default case would reach here too.

common/src/main/java/com/hortonworks/streamline/common/SchemaValueConverter.java
HeartSaVioR: Will do.

arunmahadevan:
@HeartSaVioR , tested with storm 1.1.1. Processing and event time windows works. 

However I had some extra field in the JSON and it throws an exception in the log,

```
Caused by: java.lang.IllegalArgumentException: Field list doesn't match to output schema.
	at com.hortonworks.streamline.common.SchemaValueConverter.convertMap(SchemaValueConverter.java:37)
```

1. It will be good to put the missing or extra fields in the exception so that users can quickly identify and fix it.
2. In the UI the exception is not propagated. Will raise a separate issue for this.

I think this PR can be merged after 1 is addressed.

Anther issue:
Sometimes the storm process keeps on running and the test run never completes until one manually kills the storm local mode process. As discussed we can,

1. Provide a kill API to forcefully stop the test run
2. Show the events in the UI as it becomes available. I think the API already returns the events, UI needs to poll and show the results.

That way even if the process gets stuck, the user can stop the test run after they have seen the expected results and even if the process does not get stuck it will be useful in general. Will raise issues for the above two.

HeartSaVioR:
@arunmahadevan 
Thanks for the testing. I addressed 1. 
Please note that each value is converted and conversion gets stopped at first occurrence of error case, so the values are having different fields unfortunately, only first error case would be reported.

shahsank3t:
@arunmahadevan Filed an issue https://github.com/hortonworks/streamline/issues/894 for the second part of your previous comment.

arunmahadevan:
+1

PR #896, opened 2017-09-06
Modify "migrate" option of bootstrap-storage script to behave like "create" if the target database is empty

raju-saravanan:


bootstrap/bootstrap-storage.sh
```diff
@@ -56,7 +56,7 @@ function printUsage {
     cat <<-EOF
 USAGE: $0 [create|migrate|info|validate|drop|drop-create|repair|check-connection]
    create           : Creates the tables. The target database should be empty
-   migrate          : Migrates the database to the latest version. The target database should not be empty. Use "info" to see the current version and the pending migrations
+   migrate          : Migrates the database to the latest version, behaves like "create" if the database is empty
```
arunmahadevan: Migrates the database to the latest version or creates the tables if the database is empty.

raju-saravanan: Updated the message

arunmahadevan:
+1

PR #900, opened 2017-09-07
ISSUE-899 Test mode: Provide an option to force stop a test run

### LINKED ISSUE #899: Test mode: Provide an option to force stop a test run
Sometimes the storm process keeps on running and the test run never completes until one manually kills the storm local mode process.

That way even if the process gets stuck, the user can stop the test run after they have seen the expected results and even if the process does not get stuck it will be useful in general.

HeartSaVioR:
* add new REST API endpoint: force stop running test
  * /topologies/{topologyId}/actions/killtest/{historyId}
  * it just turns on the flag to force stop, and return immediately
  * response with OK would be {"flagged": true/false}
    * flagged will be 'true' only when process is running now and 'false' otherwise so it doesn't indicate request has been failed
* let 'test run' handles force stop flag and destroy process if flag is turned on

This closes #899 

As I also noted to https://github.com/hortonworks/streamline/issues/899#issuecomment-327700819, please keep it mind that 'force stop' will make test run marked as 'failed'.

### DIFF
--- a/streams/actions/src/main/java/com/hortonworks/streamline/streams/actions/TopologyActions.java
+++ b/streams/actions/src/main/java/com/hortonworks/streamline/streams/actions/TopologyActions.java
@@ -15,6 +15,7 @@
  **/
 package com.hortonworks.streamline.streams.actions;
 
+import com.hortonworks.streamline.streams.catalog.TopologyTestRunHistory;
 import com.hortonworks.streamline.streams.layout.component.TopologyLayout;
 import com.hortonworks.streamline.streams.layout.component.impl.testing.TestRunProcessor;
 import com.hortonworks.streamline.streams.layout.component.impl.testing.TestRunSink;
@@ -39,10 +40,19 @@ public interface TopologyActions {
     // Compose and run parameter topology as test mode using the underlying streaming engine.
     // The parameter 'topology' should contain its own topology DAG.
     // Please refer the javadoc of TestRunSource and also TestRunSink to see which information this method requires.
-    void testRun(TopologyLayout topology, String mavenArtifacts,
+    void runTest(TopologyLayout topology, TopologyTestRunHistory testRunHistory, String mavenArtifacts,
                  Map<String, TestRunSource> testRunSourcesForEachSource,
                  Map<String, TestRunProcessor> testRunProcessorsForEachProcessor,
                  Map<String, TestRunSink> testRunSinksForEachSink, Optional<Long> durationSecs) throws Exception;
+
+    // Kill topology running as test mode if exists.
+    // Minimum requirement of this interface method is ensuring current running topology to be killed eventually.
+    // (Whether kill topology immediately or not is up to the implementation detail.)
+    // Querying TopologyTestRunHistory would still give the status of test run, including it is killed or not.
+    // The return value denotes whether killing topology is at least be triggered or not:
+    // it returns true if it succeeds to trigger, false otherwise.
+    // (topology is still not launched, process for topology test completed, etc.)
+    boolean killTest(TopologyTestRunHistory testRunHistory);
 
     //Kill the artifact that was deployed using deploy
     void kill (TopologyLayout topology, String asUser) throws Exception;
--- a/streams/runners/storm/actions/src/main/java/com/hortonworks/streamline/streams/actions/storm/topology/StormTopologyActionsImpl.java
+++ b/streams/runners/storm/actions/src/main/java/com/hortonworks/streamline/streams/actions/storm/topology/StormTopologyActionsImpl.java
@@ -22,6 +22,7 @@ import com.hortonworks.streamline.common.exception.service.exception.request.Top
 import com.hortonworks.streamline.streams.actions.StatusImpl;
 import com.hortonworks.streamline.streams.actions.TopologyActionContext;
 import com.hortonworks.streamline.streams.actions.TopologyActions;
+import com.hortonworks.streamline.streams.catalog.TopologyTestRunHistory;
 import com.hortonworks.streamline.streams.cluster.Constants;
 import com.hortonworks.streamline.streams.cluster.service.EnvironmentService;
 import com.hortonworks.streamline.streams.exception.TopologyNotAliveException;
@@ -58,6 +59,7 @@ import javax.ws.rs.client.ClientBuilder;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.OutputStreamWriter;
 import java.io.StringWriter;
 import java.nio.charset.Charset;
@@ -73,6 +75,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
@@ -133,6 +137,7 @@ public class StormTopologyActionsImpl implements TopologyActions {
     private long nimbusThriftMaxBufferSize;
 
     private AutoCredsServiceConfigurationReader serviceConfigurationReader;
+    private final ConcurrentHashMap<Long, Boolean> forceKillRequests = new ConcurrentHashMap<>();
 
     public StormTopologyActionsImpl() {
     }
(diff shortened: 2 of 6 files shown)


HeartSaVioR:
I added some log line with INFO because it is hard to debug topology test run with INFO level. Please let me know if we want to change the log level for them to DEBUG.

streams/runners/storm/actions/src/main/java/com/hortonworks/streamline/streams/actions/storm/topology/StormTopologyActionsImpl.java
```diff
@@ -765,6 +785,65 @@ private ShellProcessResult executeShellProcess (List<String> commands) throws  E
...
+                process.destroyForcibly();
+            }
+
+            // leave a log for each 10 secs...
+            if ((idx + 1) % 100 == 0) {
+                LOG.info("Still waiting for test run for history {} to be finished...", topologyRunHistoryId);
+            }
+
+            try {
+                process.waitFor(100, TimeUnit.MILLISECONDS);
```
arunmahadevan: maybe exit only when waitFor returns true? not sure if wait hasn't yet returned true it will cause zombies.

streams/runners/storm/actions/src/main/java/com/hortonworks/streamline/streams/actions/storm/topology/StormTopologyActionsImpl.java
```diff
@@ -259,7 +266,7 @@ public void deploy(TopologyLayout topology, String mavenArtifacts, TopologyActio
     }
 
     @Override
-    public void testRun(TopologyLayout topology, String mavenArtifacts,
+    public void runTest(TopologyTestRunHistory testRunHistory, TopologyLayout topology, String mavenArtifacts,
```
arunmahadevan: testRunHistory can be the second argument. All other methods take TopologyLayout as the first arg.

streams/service/src/main/java/com/hortonworks/streamline/streams/service/TopologyTestRunResource.java
```diff
@@ -98,6 +99,25 @@ public Response testRunTopology (@Context UriInfo urlInfo,
...
+        TopologyTestRunHistory history = catalogService.getTopologyTestRunHistory(historyId);
+
+        if (topology == null) {
+            throw EntityNotFoundException.byName("Topology with ID " + topologyId.toString());
+        } else if (history == null) {
+            throw EntityNotFoundException.byName("TopologyTestRunHistory with ID " + historyId.toString());
+        }
+
+        boolean flagged = actionsService.killTestRunTopology(topology, history);
+        return WSUtils.respondEntity(Collections.singletonMap("flagged", flagged), OK);
```
arunmahadevan: Since this just marks for kill, when does UI consider kill as successful ? It needs to query the status?

streams/runners/storm/actions/src/main/java/com/hortonworks/streamline/streams/actions/storm/topology/StormTopologyActionsImpl.java
HeartSaVioR: Good point. Will address.

streams/runners/storm/actions/src/main/java/com/hortonworks/streamline/streams/actions/storm/topology/StormTopologyActionsImpl.java
HeartSaVioR: This loop doesn't terminate unless process has exit code: isAlive() checks whether exit code is available or not.

streams/service/src/main/java/com/hortonworks/streamline/streams/service/TopologyTestRunResource.java
HeartSaVioR: Yes this API just triggers force kill, and loop for polling test history and events should be continue until history provides 'finish' to true.

HeartSaVioR:
Btw, I found only one non-daemon thread prevents process to be not finished: `executorService` in WaterMarkEventGenerator. I have been testing several scenarios and found only topologies which does event time windowing have the issue consistently. I dumped thread via jstack but impossible to find from jstack result because it doesn't have thread factory hence thread name is pool-*.

I'll file an issue and provide a patch soon.

HeartSaVioR:
Filed issue and submitted PR: https://issues.apache.org/jira/browse/STORM-2724

shahsank3t:
@HeartSaVioR While testing this out, I ran a test case and got error:

**API:** /api/v1/catalog/topologies/1/testhistories/2/events
**Respone:** {"responseMessage":"Event log file of history 2 does not exist or is not readable."}

**Logs:**
WARN   [10:19:28.807] [ForkJoinPool-2-worker-2] c.h.s.s.a.t.s.TopologyTestRunner -  Exception thrown while running Topology as test mode. Marking as 'failed'. topology id: 1 
java.lang.Exception: Topology could not be run successfully as test mode: storm deploy command failed
	at com.hortonworks.streamline.streams.actions.storm.topology.StormTopologyActionsImpl.runTest(StormTopologyActionsImpl.java:309)
	at com.hortonworks.streamline.streams.actions.topology.service.TopologyTestRunner.runTestInBackground(TopologyTestRunner.java:200)
	at com.hortonworks.streamline.streams.actions.topology.service.TopologyTestRunner.lambda$runTest$21(TopologyTestRunner.java:177)
	at com.hortonworks.streamline.common.util.ParallelStreamUtil.lambda$runAsync$7(ParallelStreamUtil.java:56)
	at java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1590)
	at java.util.concurrent.CompletableFuture$AsyncSupply.exec(CompletableFuture.java:1582)
	at java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:289)
	at java.util.concurrent.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1056)
	at java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1692)
	at java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:157)

HeartSaVioR:
@shahsank3t 
Thanks for testing!

Please make sure that the process of test is started. The new api endpoint doesn't do anything if process is not started. 
And due to how Storm local cluster works, you need to wait a bit more than some seconds to let process initializes local cluster env. and run actual worker. 
(event file is not generated before running actual worker. might be better to create one before launching process btw.)

Providing exported test topology json, test data, and streamline log would be great to track the issue and reproduce the issue you've seen.

shahsank3t:
@HeartSaVioR 
Sorry, my bad. There was some config issue which I resolved and was able to run the testcase as well as kill it.

HeartSaVioR:
@arunmahadevan Please have a second look when you have time. Thanks in advance!

streams/actions/src/main/java/com/hortonworks/streamline/streams/actions/TopologyActions.java
```diff
@@ -39,11 +40,16 @@
...
     // The parameter 'topology' should contain its own topology DAG.
     // Please refer the javadoc of TestRunSource and also TestRunSink to see which information this method requires.
-    void testRun(TopologyLayout topology, String mavenArtifacts,
+    void runTest(TopologyLayout topology, TopologyTestRunHistory testRunHistory, String mavenArtifacts,
                  Map<String, TestRunSource> testRunSourcesForEachSource,
                  Map<String, TestRunProcessor> testRunProcessorsForEachProcessor,
                  Map<String, TestRunSink> testRunSinksForEachSink, Optional<Long> durationSecs) throws Exception;
 
+    // kill topology running as test mode if exists
+    // return true if it succeeds to turn on the flag,
```
arunmahadevan: Can you make it clear what you mean by 'flag' and that the implementations are supposed to just submit for killing? Right now can one query the status of killTest via TopologyActions interface ? Maybe this can return a `Future` which can be queried to check the status (can be done later if its more work).

HeartSaVioR: Looks like I left a comment about interface too coupled with implement details. I'll leave a requirement of the method.

At least for Storm implementation we don't know whether kill succeeds or not because it just triggers the flag and killing process occurs from another thread.
We still have a chance to determine the status (with some works), but we need to decide whether backend responds immediately and let UI continue polling, or backend waits for killing and responds afterward. If we want former, maybe better not to put effort which we don't actually use.

HeartSaVioR: @arunmahadevan 
I just changed description to make requirement of method clearer, but please file an issue if you see the benefit to have return type as `Future`.

HeartSaVioR:
@arunmahadevan Squashed the commits.

PR #904, opened 2017-09-12
ISSUE-901 Test mode: Create event file prior to launch actual test process

### LINKED ISSUE #901: Test Mode: Need a field to identify when to start polling of event logs
Currently, when we hit 'testhistories' API, in background actual process take some time before starting the actual worker but there's no way to identify when that worker starts writing the logs into the file because of which UI calls 'events' API and get error: 'Event log file of history 17 does not exist or is not readable.'

UI needs a flag/field in 'testhistories' API that when set to true, UI can start polling on 'events' API and gets proper data/error.

HeartSaVioR:
* this change ensures that empty event will be available immediately when UI gets response

This closes #901.

Please refer my comment on #901. Although this doesn't go with suggested way in #901, I guess this patch resolves the origin issue simpler.

@shahsank3t UI can call /events immediately after receiving response of request for test run. It would be empty event indeed.

@arunmahadevan @shahsank3t Please review and comment. Thanks in advance!

### DIFF
--- a/streams/actions/src/main/java/com/hortonworks/streamline/streams/actions/topology/service/TopologyTestRunner.java
+++ b/streams/actions/src/main/java/com/hortonworks/streamline/streams/actions/topology/service/TopologyTestRunner.java
@@ -169,6 +169,9 @@ public class TopologyTestRunner {
                     }
                 }));
 
+        // just create event file before running actual test run process
+        createEventLogFile(eventLogFilePath);
+
         TopologyTestRunHistory history = initializeTopologyTestRunHistory(topology, testCase,
                 expectedOutputRecordsMap, eventLogFilePath);
         catalogService.addTopologyTestRunHistory(history);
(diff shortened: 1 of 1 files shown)


streams/actions/src/main/java/com/hortonworks/streamline/streams/actions/topology/service/TopologyTestRunner.java
```diff
@@ -181,6 +184,15 @@ public TopologyTestRunHistory runTest(TopologyActions topologyActions, Topology
         return history;
     }
 
+    @edu.umd.cs.findbugs.annotations.SuppressWarnings(
+            value="RV_RETURN_VALUE_IGNORED_BAD_PRACTICE",
+            justification = "We don't have to deal with the return values from mkdirs() and createNewFile()")
```
arunmahadevan: mkdirs returns false if "the dir already exists" or "dir does not exist but could not be created". Maybe you can check if dir exists before invoking mkdirs and then check for its return code ?

arunmahadevan:
+1 LGTM. @shahsank3t can you check if it addresses #901 ?

streams/actions/src/main/java/com/hortonworks/streamline/streams/actions/topology/service/TopologyTestRunner.java
HeartSaVioR: Yes that would be better. I'll address it.

HeartSaVioR:
@arunmahadevan Squashed the commits.

Issue #913, opened 2017-09-13
Add a user friendly page for status code 401 responses


priyank5485:
This issue is corresponding to https://github.com/hortonworks/registry/issues/141
For the changes to work correctly, it depends on the other PR. However, merging it without merging the other PR will still work the way it works now.

PR #914, opened 2017-09-13
ISSUE-913: Add a user friendly error page for 401 status code response

priyank5485:
Change-Id: Ia3b2a1513cbb282a319f8bcefc5cb6a2a0bdc4f2

### DIFF
--- a/common/src/main/java/com/hortonworks/streamline/common/exception/service/exception/request/WebserviceAuthorizationException.java
+++ b/common/src/main/java/com/hortonworks/streamline/common/exception/service/exception/request/WebserviceAuthorizationException.java
@@ -21,6 +21,6 @@ import javax.ws.rs.core.Response;
 
 public class WebserviceAuthorizationException extends WebServiceException {
     public WebserviceAuthorizationException(String msg) {
-        super(Response.Status.UNAUTHORIZED, msg);
+        super(Response.Status.FORBIDDEN, msg);
     }
 }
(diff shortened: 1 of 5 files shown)


priyank5485:
Closes https://github.com/hortonworks/streamline/issues/913

priyank5485:
This change is in relation to the https://github.com/hortonworks/registry/pull/247

priyank5485:
@shahsank3t I have changed the response status code from 401 to 403 in WebserviceAuthorizationException.java Can you check if that will break anything on the UI side?

shahsank3t:
@priyank5485 This won't break anything on the UI since we have no relation between our UI and this custom error pages. It will redirect to whatever page is mentioned in ErrorPageErrorHandler.

priyank5485:
@satishd Can you please review and merge this? Its already tested in a kerberos setup. Thanks.

common/src/main/java/com/hortonworks/streamline/common/exception/service/exception/request/WebserviceAuthorizationException.java
```diff
@@ -21,6 +21,6 @@
 
 public class WebserviceAuthorizationException extends WebServiceException {
     public WebserviceAuthorizationException(String msg) {
-        super(Response.Status.UNAUTHORIZED, msg);
+        super(Response.Status.FORBIDDEN, msg);
```
satishd: @priyank5485 Why did we change this from `Response.Status.UNAUTHORIZED` to `Response.Status.FORBIDDEN`?

priyank5485: @satishd I changed it because i was reading up on status codes and FORBIDDEN seems to be the right choice. For 401 Unauthorized you are supposed to have a WWW-Authenticate header which we were not passing. I am pasting a couple of links that explain the difference between the two and based on which i felt we should change it to FORBIDDEN.

https://en.wikipedia.org/wiki/List_of_HTTP_status_codes
https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html
https://stackoverflow.com/questions/3297048/403-forbidden-vs-401-unauthorized-http-responses

satishd: @priyank5485 thanks, that makes sense to me.

Issue #917, opened 2017-09-15
Test mode: Validate test case data on clicking OK


HeartSaVioR:
When the test case data is saved, it needs to be validated to see if its compatible with the current source output schema.

The sources can be edited and schema changed later, and that looks OK because validation will be also happening at runtime.

shahsank3t:
Assigning this to myself as well since the need to work on the UI side of things.

sameer79:
Rasied PR https://github.com/hortonworks/streamline/pull/926

PR #919, opened 2017-09-15
ISSUE-917 Test mode: Validate test case data on clicking OK

HeartSaVioR:
* address backend support: introduce new API endpoint for validating topology test case source
  * /api/v1/catalog/topologies/{topologyId}/testcases/{testCaseId}/sources/validate
    * for verifying data before requesting to add test case source
    * same payload as endpoint for adding test case source
  * /topologies/{topologyId}/testcases/{testCaseId}/sources/validate
    * for verifying data before requesting topology test run
    * ideal to request for all topology test case sources

@shahsank3t 
I added description for new API endpoint. It responds with HTTP status 200 when verification passes, and 400 when verification fails.
Please let me know if you have any questions regarding new endpoint.
Two new API endpoints are having different usages. For latter thing backend may be able to just retrieve all test sources and verify just before running test. Please let me know if you think this approach is better than let UI requests for that.

I'll not set auto close to #917 since UI work is necessary.

@arunmahadevan @shahsank3t Please have a look and comment. Thanks in advance.

### DIFF
--- a/common/src/main/java/com/hortonworks/streamline/common/SchemaValueConverter.java
+++ b/common/src/main/java/com/hortonworks/streamline/common/SchemaValueConverter.java
@@ -17,6 +17,7 @@
 package com.hortonworks.streamline.common;
 
 import com.hortonworks.registries.common.Schema;
+import com.hortonworks.streamline.common.exception.SchemaValidationFailedException;
 
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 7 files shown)


shahsank3t:
@HeartSaVioR 
I don't think there's a need to have 2 different API's for validation.
IMHO, let the user save test source as it is and when he hits Run, we make the second API request and backend returns success or error stating which test source failed in validation and UI can reveal the same.

arunmahadevan:
@shahsank3t I think its better to validate as soon as the user tries to save the test case (and not let them save) so that they can correct it there itself. This was also the feedback during the demo.

shahsank3t:
@arunmahadevan oh ok. I missed that part.
First validation is when the user tries to save test case and second validation is to make sure user didn't go and change the source configuration after initially saving the record. Is that correct?

arunmahadevan:
For the second part the backend maybe able to do it during the test run without UI making a separate call. I think thats what @HeartSaVioR  is asking for suggestion from UI standpoint.

shahsank3t:
> For the second part the backend maybe able to do it during the test run without UI making a separate call. I think thats what @HeartSaVioR is asking for suggestion from UI standpoint.

Oh yes, I agree, that would be better if the backend handles without UI making a separate call.

HeartSaVioR:
@arunmahadevan @shahsank3t 
Latest commit enables validation against test case sources before running topology test. Please check it out.

arunmahadevan:
LGTM. Once Sanket confirms it works fine with UI, I will merge.

HeartSaVioR:
This looks like merged via 0955980c6808ebd81948ca2038746ed34edf1732 but commit message doesn't include auto close message. I'll close it.

PR #922, opened 2017-09-15
Add Select All option to most of processor & sink where applicable

sameer79:
Select All options functionality to other processors & sinks, which allows the user to add multiple input fields.

 @shahsank3t  please review the same and comment if needed.


shahsank3t:
+1

PR #930, opened 2017-09-19
ISSUE-923 Improve the commit process and remove the CHANGELOG

### LINKED ISSUE #923: Improve the commit process and remove the CHANGELOG
Right now its getting difficult to track what issues got fixed in different branches. Sometimes we don't put ISSUE in the commit message, sometimes its BUG- and sometimes there are multiple commits for fixing an issue.

We need to follow some conventions for the commits like 
1. Mandatorily have ISSUE- in the commit  message 
2. strictly one commit per issue
3. Preferably have some script that can squash and validate the restrictions and merge.

This will make it easy to track the issues that got fixed in different branches and compare the branches.

HeartSaVioR:
* introduce merge script which is based on Spark merge script and borrow some block from Kafka
  * removed unnecessary parts regarding porting back, handling JIRA issues including marking versions

This closes #923 but need to remove CHANGELOG.md separately since including removal of CHANGELOG.md will incur merge conflict.

Please play with merge script to determine how it works and how it configures commit title and commit message of squashed commit.
The script is not that complicated if you are familiar with git command so you may want to also read the script.

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -102,18 +102,26 @@ need to:
    account.  See [Fork a repo](https://help.github.com/articles/fork-a-repo) for detailed instructions.
 2. Commit any changes to your fork.
 3. Send a [pull request](https://help.github.com/articles/creating-a-pull-request) to the Streamline GitHub repository
-   that you forked in step 1.  If your pull request is related to an existing IoTaS JIRA ticket -- for instance, because
-   you reported a bug report via JIRA earlier -- then prefix the title of your pull request with the corresponding JIRA
-   ticket number (e.g. `IOT-123: ...`).
+   that you forked in step 1.  If your pull request is related to an existing Github issue  -- for instance, because
+   you reported a bug report via Github issue earlier -- then prefix the title of your pull request with the corresponding 
+   Github issue number (e.g. `ISSUE-123: ...`).
 
 You may want to read [Syncing a fork](https://help.github.com/articles/syncing-a-fork) for instructions on how to keep
-your fork up to date with the latest changes of the upstream  `Streams` repository.
+your fork up to date with the latest changes of the upstream  `Streamline` repository.
 
-### Git Commit Messages Format
+### The format of pull request
 
-The Git commit messages must be standardized as follows:
+The title of pull request must be standardized as follows:
 
-STREAMLINE-XXX: Title matching exactly the JIRA Summary (title)
+```
+ISSUE-XXX: Title matching exactly the title of Github issue
+```
+
+The text immediately following the Github issue number (ISSUE-XXX: ) must be an exact transcription of the title of Github issue, not the a summary of the contents of the patch.
+
+If the title of Github issue does not accurately describe what the patch is addressing, the title of Github issue must be modified, and then copied to the Git commit message.
+
+And the body of pull request is encouraged to be standardized as follows:
 
 ```
 - An optional, bulleted (+, -, ., *), summary of the contents of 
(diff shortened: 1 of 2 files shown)


HeartSaVioR:
Please note that PR's body is picked for squashed commit message instead of accumulation of commit messages in PR's commits. This is based on the decision that Spark / Kafka / etc projects define the template of pull request which guarantees the PR's body to be valuable to get into commit message.

HeartSaVioR:
* It checks the title of PR to verify it conforms to the rule: the title of PR should be started with ‘ISSUE-‘.
  * It also means that pull requests should be based on corresponding issues.
  * If the PR doesn’t respect the rule, merger can stop merging, force use the title, fix the title manually.
* It checks out target branch and PR’s branch, and squash commits in PR into one.
  * If there’re multiple authors in PR, it just stops merging instead of enumerating authors and letting merge pick primary author.
    * This was a blocker for Storm introducing similar merge script.
    * For now, we can rely on merger to handle the case manually instead of making script much complicated.
  * Commit title will be the title of PR.
  * Commit message will contain below things:
    * The body of PR
    * The information of author(s)
    * The information of merger if merger resolves merge conflict manually
    * The origin commits on PR if necessary (optional)
    * Auto close message for Github PR
* Push the squashed commit to the remote (HWX gerrit repository).

arunmahadevan:
Looks good overall. A few comments after testing the merge script.

1. Set executable mode for dev-tools/merge_streamline_pr.py

2. "The title of PR doesn't conform to the Streamline rule" Maybe mention that it should start with ISSUE-

3. The script assumes the remote is 'github' for fetching. Can we figure out the remote from the PR and use that remote to fetch? For pushing it may be ok to have gerrit-master hardcoded for now.

4. When a closed PR is given it says Pull request 1 is not mergeable in its current form. Maybe inform the user that the PR is already closed in the message and if they want to continue.

5. "List pull request commits in squashed commit message(y/n)". Maybe default can be 'n' since we want to rely on PR message?

6. Warn / Fail if the PR message is empty. Right now it goes ahead with the merge.

7. When Push to origin? (y/n) , I choose 'n', the local branch is deleted. Maybe give an option to keep the local branch so that it can be reviewed and manually pushed.

8. I had only one commit in PR, but the PR is showing as closed with unmerged commit. Is this due to the script modifying the commit message? https://github.com/arunmahadevan/streamline/pull/2

HeartSaVioR:
> 1. Set executable mode for dev-tools/merge_streamline_pr.py

OK. Will fix.

> 2. "The title of PR doesn't conform to the Streamline rule" Maybe mention that it should start with ISSUE-

I thought we may want to have more rules, but mentioning why it doesn't conform would be better. Will fix.

> 3. The script assumes the remote is 'github' for fetching. Can we figure out the remote from the PR and use that remote to fetch? For pushing it may be ok to have gerrit-master hardcoded for now.

This way works with multiple projects for a long time. Unless there's specific reason to do it, I'd like to keep it as it is. Do you have reasons to suggest?

> 4. When a closed PR is given it says Pull request 1 is not mergeable in its current form. Maybe inform the user that the PR is already closed in the message and if they want to continue.

OK. Originally script allows closed PR because of handling backport, but we don't need it for now. I think we can just stop if PR is closed.

> 5. "List pull request commits in squashed commit message(y/n)". Maybe default can be 'n' since we want to rely on PR message?

Will fix.

> 6. Warn / Fail if the PR message is empty. Right now it goes ahead with the merge.

Will fix. Btw, we may have to have template form of PR message. If not adding PR body to commit message sometimes go worse.

> 7. When Push to origin? (y/n) , I choose 'n', the local branch is deleted. Maybe give an option to keep the local branch so that it can be reviewed and manually pushed.

We could do it, but maybe we don't support cleaning up in that case. I'll work on that.

> 8. I had only one commit in PR, but the PR is showing as closed with unmerged commit. Is this due to the script modifying the commit message? arunmahadevan#2

That is expected behavior. Given that we're squashing the commits, PR will not be marked as 'merged'. 
You can see other projects like Spark, Kafka, and so on.
https://github.com/apache/spark/pulls?q=is%3Apr+is%3Aclosed

HeartSaVioR:
@arunmahadevan Addressed review comments.

arunmahadevan:
>>The script assumes the remote is 'github' for fetching. Can we figure out the remote from the PR and use that remote to fetch? For pushing it may be ok to have gerrit-master hardcoded for now.

>This way works with multiple projects for a long time. Unless there's specific reason to do it, I'd like to keep it as it is. Do you have reasons to suggest?

If I want to merge multiple PRs from different authors, each PR would need to be fetched from a different remote right? In that case I need to add all the remotes and keep switching the env var?

HeartSaVioR:
@arunmahadevan 
No it pulls from PR directly, not using remote repository of the PR. You just need to set github repo and gerrit repo, and forget about everything.

arunmahadevan:
Tested and works fine. Once the documentation DEVELOPER.md is updated it can be merged.

The script requires 'github' and 'hwx-gerrit' remotes to be available.  Either the script can auto create it and point to hortonworks and gerrit repos or we need to document it.

HeartSaVioR:
@arunmahadevan Addressed the guide for merge script, and squashed the commits into one.

dev-tools/merge_streamline_pr.py
```diff
@@ -0,0 +1,270 @@
...
+import json
+import os
+import subprocess
+import sys
+import urllib2
+
+# Location of your Streamline git development area
+STREAMLINE_HOME = os.environ.get("STREAMLINE_HOME", os.getcwd())
+# Remote name which points to pull repository
+PR_REMOTE_NAME = os.environ.get("PR_REMOTE_NAME", "pull-repo")
```
arunmahadevan: Can the script fail if the PR_REMOTE_NAME is not set and the pull-repo also does not exist ? Right now the script fails while running the git commands, but it will be better to check it and fail at the begining. Similarly for PUSH_REMOTE_NAME.

HeartSaVioR: @arunmahadevan Addressed.

DEVELOPER.md
```diff
@@ -102,18 +102,18 @@ need to:
...
-   ticket number (e.g. `IOT-123: ...`).
+   that you forked in step 1.  If your pull request is related to an existing Github issue  -- for instance, because
+   you reported a bug report via Github issue earlier -- then prefix the title of your pull request with the corresponding 
+   Github issue number (e.g. `ISSUE-123: ...`).
 
 You may want to read [Syncing a fork](https://help.github.com/articles/syncing-a-fork) for instructions on how to keep
-your fork up to date with the latest changes of the upstream  `Streams` repository.
+your fork up to date with the latest changes of the upstream  `Streamline` repository.
 
 ### Git Commit Messages Format
```
arunmahadevan: Should this be changed to PR message format since the script will put the PR message as the commit message ?

HeartSaVioR: Sure, but I also think we should apply template of pull request body. For now I will just change this to PR format.

HeartSaVioR: @arunmahadevan Also addressed.

arunmahadevan:
+1

PR #937, opened 2017-09-25
ISSUE-936: Backend changes for handling the test case data as a string to preserve formatting

arunmahadevan:
Interpret the test case values as a JSON string during the conversion.


HeartSaVioR:
+1 given that code looks good and verification from UI is done.

Issue #942, opened 2017-09-27
SAM Kafka source crashes when target component has parallelism > 1


arunmahadevan:
It depends on target task parallelism and the data.

priyank5485:
@arunmahadevan Should we put this on HDF-3.0-maint branch as well for 3.0.2 release?

PR #943, opened 2017-09-27
ISSUE-942: SAM Kafka source crashes when target component has parallelism > 1

arunmahadevan:
Mask the hashcode in FieldsGroupingAsCustomGrouping to always return positive values.


arunmahadevan:
@priyank5485 I think its a combination of data and the target task parallelism which causes the index to become negative. Can you validate using the data in the test case? Set target component parallelism = 2 and use string key with value = "PUFPaY9KxDAcGqfsorJp1R".

arunmahadevan:
Fixes #942

PR #950, opened 2017-10-04
ISSUE-949: Disallow null values in test mode data

arunmahadevan:
Null values are not supported in the runtime. Keeping the same behavior in test mode for now.

### DIFF
--- a/common/src/main/java/com/hortonworks/streamline/common/SchemaValueConverter.java
+++ b/common/src/main/java/com/hortonworks/streamline/common/SchemaValueConverter.java
@@ -19,6 +19,7 @@ package com.hortonworks.streamline.common;
 import com.hortonworks.registries.common.Schema;
 import com.hortonworks.streamline.common.exception.SchemaValidationFailedException;
 
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/hortonworks/streamline/common/SchemaValueConverter.java
```diff
@@ -70,8 +72,9 @@ private SchemaValueConverter() {
                     requiredFieldsNotFoundInValue.stream().map(Schema.Field::getName).collect(toList()));
         }
 
-        return value.entrySet().stream()
-                .collect(toMap(Map.Entry::getKey, e -> convert(fieldToType.get(e.getKey()), e.getValue())));
```
arunmahadevan: `toMap` does not allow null values. Replacing with below to support null values in future.

HeartSaVioR: I didn't notice it. Nice finding. 
Looks like related issue here: https://bugs.openjdk.java.net/browse/JDK-8148463

common/src/main/java/com/hortonworks/streamline/common/SchemaValueConverter.java
```diff
@@ -82,10 +85,7 @@ private SchemaValueConverter() {
      * @return the converted value
      */
     public static Object convert(Schema.Type type, Object value) {
-        if (value == null) {
-            return null;
-        }
-
+        Objects.requireNonNull(value, "'null' value is not allowed");
```
HeartSaVioR: We may want to handle this to IllegalArgumentException, and change above `forEach` to catch IllegalArgumentException and wrap with SchemaValidationFailedException. 
Bubbling up NullPointerException or IllegalArgumentException to the REST API response would be bad for end users.

If you just want to ensure denying null, I'm OK to leave it as it is, and I could handle it.

arunmahadevan: I reverted the changes in `convert` and added a check in the `foreach` to throw `SchemaValidationException` if value is null.

HeartSaVioR:
+1

PR #952, opened 2017-10-05
ISSUE-951: Join bolt does not work in Test mode

arunmahadevan:
Join processor should be recognized as a windowed processor and wrapped in TestRunWindowProcessorBolt.

### DIFF
--- a/streams/layout/src/main/java/com/hortonworks/streamline/streams/layout/component/impl/JoinProcessor.java
+++ b/streams/layout/src/main/java/com/hortonworks/streamline/streams/layout/component/impl/JoinProcessor.java
@@ -14,30 +14,23 @@
   * limitations under the License.
  **/
 
-package com.hortonworks.streamline.streams.layout.component.impl.splitjoin;
+package com.hortonworks.streamline.streams.layout.component.impl;
 
-import com.hortonworks.streamline.streams.layout.component.impl.RulesProcessor;
-import com.hortonworks.streamline.streams.layout.component.impl.Utils;
-
-import java.util.Collections;
+import com.hortonworks.streamline.streams.layout.component.StreamlineProcessor;
+import com.hortonworks.streamline.streams.layout.component.TopologyDagVisitor;
 
 /**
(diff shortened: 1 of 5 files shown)


streams/layout/src/main/java/com/hortonworks/streamline/streams/layout/component/impl/JoinProcessor.java
```diff
@@ -14,30 +14,23 @@
...
-public class JoinProcessor extends RulesProcessor {
-
-    public static final String CONFIG_KEY_JOIN = "join-config";
-
-    public JoinProcessor() {
-    }
-
-    public void setJoinAction(JoinAction joinAction) {
-        setRules(Collections.singletonList(Utils.createTrueRule(joinAction)));
+public class JoinProcessor extends StreamlineProcessor {
```
HeartSaVioR: I'm not familiar with JoinProcessor so not sure removing all the contents is safe.
Is Join Processor enough for just extending StreamlineProcessor? Are all the configuration needed provided on StreamlineProcessor for Join Processor?

arunmahadevan: to be clear I renamed the existing JoinProcessor to `SplitJoinProcessor` since its used in the context of split-join. This functionality is not yet integrated with SAM processors. JoinProcessor is a wrapper design time class so that we can handle it separately (in test mode).

HeartSaVioR: I know we have several types of join processors but are they more than 2? I mean is there another join processor along with split-join and realtime join? Which kind of join `JoinProcessor` represents?

arunmahadevan: @HeartSaVioR , This is the regular Join processor. Theres realtime Join but that does not have any corresponding design time entity. Split-Join is not used currently.

PR #960, opened 2017-10-10
ISSUE-959: Have a common storage module across streamline and registry

arunmahadevan:
1. Moved storage code to registry and refer to registry storage module in streamline.
2. Removed storage-core and storage-tool modules from streamline
3. Updated assembly to include storage modules from registry in streamline binaries.


arunmahadevan:
Raised https://github.com/hortonworks/hdf_ambari_mp/pull/110 for ambari mpack to reflect new package. Other than that I do not see any references to "streamline.storage" in the streamline repo.

Issue #961, opened 2017-10-10
Add an example to add a spout as a custom source


priyank5485:

arunmahadevan:
Addressed with https://github.com/hortonworks/streamline/pull/963

PR #963, opened 2017-10-10
ISSUE-961: Add example for adding spout as custom source and related fix

priyank5485:
ISSUE-961: Add example for adding spout as custom source and related fix

### DIFF
--- a/common/src/main/java/com/hortonworks/streamline/common/util/ProxyUtil.java
+++ b/common/src/main/java/com/hortonworks/streamline/common/util/ProxyUtil.java
@@ -23,17 +23,25 @@ import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nullable;
 import java.io.File;
+import java.io.FilenameFilter;
 import java.io.IOException;
 import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.Proxy;
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.nio.file.Path;
 import java.util.Collection;
 import java.util.List;
 import java.util.concurrent.ConcurrentHashMap;
 
 public class ProxyUtil<O> {
     private static final Logger LOG = LoggerFactory.getLogger(ProxyUtil.class);
+    private static final FilenameFilter JAR_FILENAME_FILTER = new FilenameFilter() {
+        @Override
+        public boolean accept(File dir, String name) {
+            return name.endsWith(".jar");
+        }
+    };
 
     private final Class<O> interfaceClazz;
     private final ClassLoader parentClassLoader;
@@ -47,6 +55,21 @@ public class ProxyUtil<O> {
         this.interfaceClazz = interfaceClazz;
         this.parentClassLoader = parentClassLoader;
         this.cachedClassLoaders = new ConcurrentHashMap<>();
+    }
+
+    public O loadClassFromLibsDirectory(Path libsDirectory, String classFqdn) throws MalformedURLException, InstantiationException, IllegalAccessException, ClassNotFoundException {
+        ClassLoader classLoader;
+        if (isClassLoadedFromParent(classFqdn)) {
+            classLoader = parentClassLoader;
+        } else {
+            File file = libsDirectory.toFile();
+            classLoader = findCachedClassLoader(file.getAbsolutePath());
+            if (classLoader == null) {
+               classLoader = getJarsAddedClassLoader(file);
+            }
+        }
+        O actualObject = initInstanceFromClassloader(classFqdn, classLoader);
+        return createClassLoaderAwareProxyInstance(classLoader, actualObject);
     }
 
     public O loadClassFromJar(String jarPath, String classFqdn)
@@ -115,6 +138,27 @@ public class ProxyUtil<O> {
         if (oldCl != null) {
             // discard and pick old thing
             classLoader = oldCl;
+        }
+        return classLoader;
+    }
+
+    private ClassLoader getJarsAddedClassLoader(File file) throws MalformedURLException {
+        ClassLoader classLoader = new MutableURLClassLoader(new URL[0], parentClassLoader);
+        File[] files;
+        if ((file != null) && ((files = file.listFiles(JAR_FILENAME_FILTER)) != null)) {
+            for (File jar : files) {
+                URL u = (jar.toURI().toURL());
+                ((MutableURLClassLoader) classLoader).addURL(u);
+            }
+            ClassLoader oldCl = cachedClassLoaders.putIfAbsent(file.getAbsolutePath(), classLoader);
+            if (oldCl != null) {
+                // discard and pick old thing
+                classLoader = oldCl;
+            }
+        } else {
+            String errorMessage = "Cannot get a class loader with all jars in directory: " + file;
+            LOG.warn(errorMessage);
+            throw new RuntimeException(errorMessage);
         }
         return classLoader;
     }
--- a/common/src/main/java/com/hortonworks/streamline/common/util/ProxyUtil.java
+++ b/common/src/main/java/com/hortonworks/streamline/common/util/ProxyUtil.java
@@ -57,12 +57,12 @@ public class ProxyUtil<O> {
         this.cachedClassLoaders = new ConcurrentHashMap<>();
     }
 
-    public O loadClassFromLibsDirectory(Path libsDirectory, String classFqdn) throws MalformedURLException, InstantiationException, IllegalAccessException, ClassNotFoundException {
+    public O loadClassFromLibDirectory(Path libDirectory, String classFqdn) throws MalformedURLException, InstantiationException, IllegalAccessException, ClassNotFoundException {
         ClassLoader classLoader;
         if (isClassLoadedFromParent(classFqdn)) {
             classLoader = parentClassLoader;
         } else {
-            File file = libsDirectory.toFile();
+            File file = libDirectory.toFile();
             classLoader = findCachedClassLoader(file.getAbsolutePath());
             if (classLoader == null) {
                classLoader = getJarsAddedClassLoader(file);
@@ -142,21 +142,21 @@ public class ProxyUtil<O> {
         return classLoader;
     }
 
-    private ClassLoader getJarsAddedClassLoader(File file) throws MalformedURLException {
+    private ClassLoader getJarsAddedClassLoader(File directory) throws MalformedURLException {
         ClassLoader classLoader = new MutableURLClassLoader(new URL[0], parentClassLoader);
         File[] files;
-        if ((file != null) && ((files = file.listFiles(JAR_FILENAME_FILTER)) != null)) {
+        if ((directory != null) && ((files = directory.listFiles(JAR_FILENAME_FILTER)) != null)) {
             for (File jar : files) {
                 URL u = (jar.toURI().toURL());
                 ((MutableURLClassLoader) classLoader).addURL(u);
             }
-            ClassLoader oldCl = cachedClassLoaders.putIfAbsent(file.getAbsolutePath(), classLoader);
+            ClassLoader oldCl = cachedClassLoaders.putIfAbsent(directory.getAbsolutePath(), classLoader);
             if (oldCl != null) {
                 // discard and pick old thing
                 classLoader = oldCl;
             }
         } else {
-            String errorMessage = "Cannot get a class loader with all jars in directory: " + file;
+            String errorMessage = "Cannot get a class loader with all jars in directory: " + directory;
             LOG.warn(errorMessage);
             throw new RuntimeException(errorMessage);
         }
--- a/examples/sources/CustomSource.md
+++ b/examples/sources/CustomSource.md
@@ -1,0 +1,24 @@
+# How to add a spout as a custom source
+
+If you already have a spout implemented and want to use it in a SAM application, SAM lets you register it once  and re-use in any SAM application. This 
+way, you can avoid rewriting code and leverage existing spout code. Provided below, is an illustration of how to build and register such a spout as a SAM source. 
+All the files needed to build, register and run are present in this module.
+
+1. Check out the module on your local directory and run `mvn clean install`
+2. Open the `file-source-topology-component.json` file under resources directory and make sure that the mavenDeps property(which is of the form 
+groupId:artifactId:version) has the correct value for version. Version should be the version defined in the pom file of this module
+3. Export the variable `data`. Its value should be the location of the json file edited in step 2 above.
+4. Run the below curl command to register it as a SAM source. Before doing that, make sure you replace sam_host and sam_port with values for your environment
+ and that bundleJar form parameter points to the jar file that you build in step 1. 
+
+`curl -sS -X POST -i -F topologyComponentBundle=@$data  -F bundleJar=@/tmp/custom_source_example/streamline-examples-sources-0.1.0-SNAPSHOT.jar http://<sam_host>:<sam_port>/api/v1/catalog/streams/componentbundles/SOURCE`
+
+5. Install the jar built in step 1 above in local m2 directory for the user running SAM server on SAM server using the command below. Make sure the version 
+used is the correct one from the pom file of this project
+
+`mvn install:install-file -DgroupId=com.hortonworks.streamline -DartifactId=streamline-examples-sources -Dpackaging=jar -Dversion=0.1.0-SNAPSHOT -Dfile=/tmp/streamline-examples-sources-0.1.0-SNAPSHOT.jar -DgeneratePom=true`
+
+6. At this point, you are ready to drag and drop this source on to a SAM app. Note that this source reads from a file on local filesystem and it expects the 
+file to be in a certain format. Sample file provided in this module names.csv under resources directory can be used. Please make sure that the value for path 
+configuration parameter for this source points to the location of names.csv or a similar file in the same format on storm supervisor nodes.
+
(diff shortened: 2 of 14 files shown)


priyank5485:
Closes https://github.com/hortonworks/streamline/issues/961

priyank5485:
@arunmahadevan I have tested this example end to end and it works. Apart from the code for the example, there is one server side fix in two files (ProxyUtil.java and FluxComponentFactory.java) in this PR that is needed to make the a custom source work. Please review when you get a chance.

shahsank3t:
@priyank5485 @arunmahadevan 
You will also need https://github.com/hortonworks/streamline/pull/955 & https://github.com/hortonworks/streamline/pull/957 as the UI side fixes for this particular thing.

examples/sources/src/main/java/com/hortonworks/streamline/examples/sources/FileReaderSpout.java
```diff
@@ -0,0 +1,223 @@
...
+            while ((line = br.readLine()) != null) {
+                String[] result = line.split(delimiter);
+                if (result.length != 2) {
+                   LOG.error("Format of input file not as expected. Expecting {} separated first name and last name", delimiter);
+                } else {
+                    Map<String, Object> event = new HashMap<>();
+                    event.put(FIRST_NAME, result[0].trim());
+                    event.put(LAST_NAME, result[1].trim());
+                    List<Object> values = new ArrayList<>();
+                    values.add(new MyStreamlineEvent(event, outputStream));
```
arunmahadevan: Why do we require user to construct a `MyStreamlineEvent` class?. They should be able to use the default implementation which we also use internally.

examples/sources/src/main/java/com/hortonworks/streamline/examples/sources/FileReaderSpout.java
```diff
@@ -0,0 +1,223 @@
...
+                br.close();
+            } catch (IOException e) {
+                LOG.error("Got exception while closing buffered reader");
+                throw new RuntimeException(e);
+            }
+        }
+
+    }
+
+    private static class MyStreamlineEvent implements StreamlineEvent {
```
arunmahadevan: Why do we require user to construct a MyStreamlineEvent class?. They should be able to use the default implementation which we also use internally. If its not exposed to user, we can do it as a part of this PR.

examples/sources/src/main/java/com/hortonworks/streamline/examples/sources/FileReaderSpout.java
```diff
@@ -0,0 +1,223 @@
...
+        LOG.info("Opening FileReaderSpout");
+        this.spoutOutputCollector = spoutOutputCollector;
+    }
+
+    @Override
+    public void nextTuple() {
+        BufferedReader br = null;
+        try {
+            String line;
+            br = new BufferedReader(Files.newBufferedReader(Paths.get(path), StandardCharsets.UTF_8));
```
arunmahadevan: use `try(BufferedRead br = new BufferedReader(Files.newBufferedReader(Paths.get(path), StandardCharsets.UTF_8)) {` and remove finally

streams/runners/storm/layout/src/main/java/com/hortonworks/streamline/streams/layout/storm/FluxComponentFactory.java
```diff
@@ -33,7 +33,7 @@
     FluxComponent getFluxComponent(StreamlineComponent streamlineComponent) {
         ProxyUtil<FluxComponent> proxyUtil = new ProxyUtil<>(FluxComponent.class);
         try {
-            FluxComponent fluxComponent = proxyUtil.loadClassFromJar(extraJarsLocation.toAbsolutePath().toString(), streamlineComponent.getTransformationClass());
+            FluxComponent fluxComponent = proxyUtil.loadClassFromLibsDirectory(extraJarsLocation, streamlineComponent.getTransformationClass());
```
arunmahadevan: Is the user expected to build and place the jar into the "/tmp/storm-artifacts/jars" directory ?

The extrasJarsLocation is already a dir. I think UrlClassLoader already handles directories. 

https://docs.oracle.com/javase/7/docs/api/java/net/URLClassLoader.html
"This class loader is used to load classes and resources from a search path of URLs referring to both JAR files and directories. Any URL that ends with a '/' is assumed to refer to a directory. Otherwise, the URL is assumed to refer to a JAR file which will be opened as needed."

arunmahadevan:
You can put the description of the changes in the PR description. This will become the commit message while running the merge script.

common/src/main/java/com/hortonworks/streamline/common/util/ProxyUtil.java
```diff
@@ -49,6 +57,21 @@ public ProxyUtil(Class<O> interfaceClazz, ClassLoader parentClassLoader) {
         this.cachedClassLoaders = new ConcurrentHashMap<>();
     }
 
+    public O loadClassFromLibsDirectory(Path libsDirectory, String classFqdn) throws MalformedURLException, InstantiationException, IllegalAccessException, ClassNotFoundException {
```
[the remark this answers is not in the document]
priyank5485: Will do

common/src/main/java/com/hortonworks/streamline/common/util/ProxyUtil.java
```diff
@@ -119,6 +142,27 @@ private ClassLoader getJarAddedClassLoader(String jarPath) throws MalformedURLEx
         return classLoader;
     }
 
+    private ClassLoader getJarsAddedClassLoader(File file) throws MalformedURLException {
```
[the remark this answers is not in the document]
priyank5485: Will do

common/src/main/java/com/hortonworks/streamline/common/util/ProxyUtil.java
```diff
@@ -119,6 +142,27 @@ private ClassLoader getJarAddedClassLoader(String jarPath) throws MalformedURLEx
...
+        ClassLoader classLoader = new MutableURLClassLoader(new URL[0], parentClassLoader);
+        File[] files;
+        if ((file != null) && ((files = file.listFiles(JAR_FILENAME_FILTER)) != null)) {
+            for (File jar : files) {
+                URL u = (jar.toURI().toURL());
+                ((MutableURLClassLoader) classLoader).addURL(u);
+            }
+            ClassLoader oldCl = cachedClassLoaders.putIfAbsent(file.getAbsolutePath(), classLoader);
```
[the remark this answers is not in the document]
priyank5485: Yes, its in line with the other method that loads all the classes from a jar file. Meaning, if the same jar file is updated with newer classes on FS then it will fail to load the newer classes.

examples/sources/src/main/java/com/hortonworks/streamline/examples/sources/FileReaderSpout.java
priyank5485: Will do

examples/sources/src/main/java/com/hortonworks/streamline/examples/sources/FileReaderSpout.java
priyank5485: StreamlineEventImpl class that we use internally is part of streamline-common module. I believe we only want the user to have dependency on streamline-sdk. Moreover, when you first created StreamlineEvent, it was an interface rather than a non final class that can be extended. I wanted to showcase how one can provide their own implementation as well. If we want to force the user to use our internal class, then we probably dont need to expose it as an interface. Do you remember why it was chosen to be an interface? Either ways, there is nothing preventing them from using the internal implementation since streamline-layout-storm dependency brings streamline-common transitively.

streams/runners/storm/layout/src/main/java/com/hortonworks/streamline/streams/layout/storm/FluxComponentFactory.java
priyank5485: The user is not expected to do that. The curl call used to register a component bundle accepts a jar and it stores it under file storage. While deployment we keep track if there is bundle that is non built in. If so, we download jar from storage and put it under /tmp/storm-artifacts/jars. 

As far as UrClassLoader is concerned, i read that javadoc as well and tried putting a / at the end of extraJarsLocation to see why existing method did not work. Digging more, I found that when URLClassLoader interprets it as a directory, it actually tries to search that directory for a class using the package hierarchy. So for a URLClassLoader to load a class a.b.c.Someclass, it expects the entire directory structure corresponding to the package hierarchy and finally a .class file for that class. Hence, I had to add the other method where it would load all the jar files in the directory. This in java world is usually referred to as a "libs" directory where you put all your jars. Which is why I had used the word "libs" instead of "lib". However, I don't have a opinion on that and have renamed it as per your suggestion

examples/sources/src/main/java/com/hortonworks/streamline/examples/sources/FileReaderSpout.java
priyank5485: Same as above

priyank5485:
@arunmahadevan I have addressed your review comments. Also added a md file as a help doc like you suggested. Please review again when you get a chance. I was not sure about one thing you mentioned regarding putting the description of the changes in PR description. Isnt PR description already correct? Let me know where and how to make that change or feel free to make it yourself if its a small thing.
@shahsank3t Yes we need those two as well. We can link them. As long as they both get merged before the release is tagged we should be good.

examples/sources/src/main/java/com/hortonworks/streamline/examples/sources/FileReaderSpout.java
arunmahadevan: I was suggesting to use the default impl mainly to simplify things for the first time user. They will copy the sample code and keep adding their logic and may violate the immutability requirement in the event which will lead to issues later. 

So instead of MyStreamlineEvent it could be something like:

```java
        StreamlineEventImpl.builder()
                .put(FIRST_NAME, result[0].trim())
                .put(LAST_NAME, result[1].trim())
                .build();
```

Regarding keeping it in the SDK, yes we should provide the default implementation in the SDK itself and rename the class to DefaultStreamlineEvent. Can you raise a followup issue for this?

Main idea to keep it as an interface was to only specify the contract and provide maximum flexibility for implementation within streamline itself (and maybe for some advanced users). E.g. Currently we have StreamlineEventImpl extending the Map interface which would not be possible if StreamlineEvent was an Abstract class.

arunmahadevan:
You can put some description of the changes (instead of Change-Id: Id0c264f2397b8e14297adf38e2df4db20f082053) in the PR description here https://github.com/hortonworks/streamline/pull/963#issue-264400351 which will become the commit message.

examples/sources/CustomSource.md
```diff
@@ -0,0 +1,24 @@
+# How to add a spout as a custom source
```
arunmahadevan: Rename the file to README.md so it will show the doc in the github itself.

examples/sources/CustomSource.md
```diff
@@ -0,0 +1,24 @@
...
+
+If you already have a spout implemented and want to use it in a SAM application, SAM lets you register it once  and re-use in any SAM application. This 
+way, you can avoid rewriting code and leverage existing spout code. Provided below, is an illustration of how to build and register such a spout as a SAM source. 
+All the files needed to build, register and run are present in this module.
+
+1. Check out the module on your local directory and run `mvn clean install`
+2. Open the `file-source-topology-component.json` file under resources directory and make sure that the mavenDeps property(which is of the form 
+groupId:artifactId:version) has the correct value for version. Version should be the version defined in the pom file of this module
+3. Export the variable `data`. Its value should be the location of the json file edited in step 2 above.
```
arunmahadevan: Can you list the command here ?

examples/sources/CustomSource.md
```diff
@@ -0,0 +1,24 @@
...
+4. Run the below curl command to register it as a SAM source. Before doing that, make sure you replace sam_host and sam_port with values for your environment
+ and that bundleJar form parameter points to the jar file that you build in step 1. 
+
+`curl -sS -X POST -i -F topologyComponentBundle=@$data  -F bundleJar=@/tmp/custom_source_example/streamline-examples-sources-0.1.0-SNAPSHOT.jar http://<sam_host>:<sam_port>/api/v1/catalog/streams/componentbundles/SOURCE`
+
+5. Install the jar built in step 1 above in local m2 directory for the user running SAM server on SAM server using the command below. Make sure the version
```
arunmahadevan: Is this required? Doesn't streamline automatically ship the jar (streamline-examples-sources.jar) with the storm jar command (--artifacts) ? Its not very practical for users to install the jar on all the servers running SAM.

examples/sources/src/main/java/com/hortonworks/streamline/examples/sources/FileReaderSpout.java
priyank5485: Will use StreamlineEventImpl in this example then. Have raised https://github.com/hortonworks/streamline/issues/965

examples/sources/CustomSource.md
priyank5485: Will do

examples/sources/CustomSource.md
priyank5485: Will do

examples/sources/CustomSource.md
priyank5485: This is not required for this example. Flux transformation is expected to be in streamline-examples-sources.jar. However, the actual storm spout or bolt does not have to be in the same jar since we are leveraging the mavenDeps string to fetch it from maven. I wanted to highlight that as well in this example.

priyank5485:
@arunmahadevan I have addressed your most recent review comments. Please review when you get a chance.

examples/sources/CustomSource.md
arunmahadevan: @priyank5485 , so its not required for this example since you are already putting FileReaderSpout as part of the module correct? If so we can remove the step 5 from the README. 

Maybe as a note somewhere at the end you can specify that if storm spout is not included as part of the module you need to install that separately. In this case can they build a fat jar by just having a dependency on their spout in the pom ? It will easier than installing the spout jar in all the server hosts.

arunmahadevan:
@priyank5485 , overall it looks good. Have a question regarding step 5 in the Readme.

examples/sources/CustomSource.md
priyank5485: @arunmahadevan I have pushed a change to readme file as per your suggestion. I did not get your question about fat jar. If the spout is in a separate jar then all they need to make sure is that its either available in maven local or at remote maven repo using the groupId, artifactId and version that they set in mavenDeps. We leverage that property of a non built-in bundle while submitting the topology to storm using --artifact argument to storm.py. That way storm will make sure that it will fetch the jar. Currently, the remote repos used are provided by mavenRepoUrl in streamline.yaml configurable via Ambari. By default maven central is always checked for. They can either have it published to maven central or change mavenRepoUrl and restart SAM or install it in maven local. Hope this answers your question.

arunmahadevan:
@priyank5485 , while trying the steps in the README, I am observing that after running the curl command to upload the file, the uploaded file in the filesystem is empty. This is causing deployment to fail since it cant generate the flux.

After debugging, the [addTopologyComponentBundle](https://github.com/hortonworks/streamline/blob/master/streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/service/StreamCatalogService.java#L905) invokes [loadTransformationClassForBundle](https://github.com/hortonworks/streamline/blob/master/streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/service/StreamCatalogService.java#L908) which is consuming the InputStream and later when `uploadFileToStorage` is invoked with the same stream, it does not have any more content.

Please take a look and fix it.

arunmahadevan:
@priyank5485 , Can you also add a pom file in examples dir to build sources, processors and mlmodel? After that change the top level pom entry to build examples [here](https://github.com/hortonworks/streamline/blob/master/pom.xml#L135-L136)

priyank5485:
@arunmahadevan I remember running in to a similar issue and I think it was due to not passing the curl argument correctly and was able to fix it by tweaking curl argument. May be i did not put the correct curl argument in README. Will double check. If you still have it, can you please paste your curl command that you used? I think we need to double check the bundleJar argument.

priyank5485:
@arunmahadevan I have pushed one more commit to address the pom changes like you suggested.

priyank5485:
@arunmahadevan Thanks for catching the issue regarding jar file being empty. I had run in to the same issue while testing it on the cluster. However, i was in middle of fixing some other issue, so as a temporary workaround, I had manually copied that jar over to the file storage location and had deployed the application successfully and saw data flowing. I forgot to trace the root cause to the size 0 file issue and thought may be it was about not passing the right argument to curl. It is fixed now.

arunmahadevan:
@priyank5485 , thanks for addressing the issues, merged to master. If you think this can go into 3.0.2 you can merge this and  the UI related fixes into HDF-3.0-maint.

priyank5485:
@arunmahadevan Thanks for merging. Will check with Sanket if UI fixes can be merged to HDF-3.0-maint. If yes, I will merge all of the to HDF-3.0-maint

@shahsank3t Is it okay to cherry-pick these two on to HDF-3.0-maint? In that case i will merge those two and this PR on HDF-3.0-maint.
https://github.com/hortonworks/streamline/pull/955 and https://github.com/hortonworks/streamline/pull/957

shahsank3t:
@priyank5485 Yes, it's okay to merge #955 and #957 onto HDF-3.0-maint branch.

PR #969, opened 2017-10-18
Preliminary changes for transaction support

raju-saravanan:
First cut for adding transaction support. This is not the complete changes for this feature, but a way for @arunmahadevan  and @satishd to review the approach.


satishd:
@raju-saravanan This PR should be raised in registry as registry/storage module is shared between registry and streamline.

arunmahadevan:
@raju-saravanan, can you move all the storage module and common code to registry and have only the streamline specific changes here ?

PR #977, opened 2017-10-24
ISSUE-972 SAM test mode doesn't adjust topology message timeout with windowed rule

HeartSaVioR:
* create TestRunRulesProcessor which extends RulesProcessor
  * it enables StormTopologyFluxGenerator to handle it as RulesProcessor
* modify handling windowed rules to call copy constructor of its own class, not always RulesProcessor
  * fail back to call copy constructor of RulesProcessor if the class doesn't have copy constructor

### DIFF
--- a/streams/runners/storm/layout/src/main/java/com/hortonworks/streamline/streams/layout/storm/StormTopologyFluxGenerator.java
+++ b/streams/runners/storm/layout/src/main/java/com/hortonworks/streamline/streams/layout/storm/StormTopologyFluxGenerator.java
@@ -35,6 +35,7 @@ import com.hortonworks.streamline.streams.layout.component.TopologyLayout;
 import com.hortonworks.streamline.streams.layout.component.impl.RulesProcessor;
 import com.hortonworks.streamline.streams.layout.component.rule.Rule;
 import com.hortonworks.streamline.streams.layout.component.rule.expression.Window;
+import com.hortonworks.streamline.streams.storm.common.StormTopologyUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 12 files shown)


streams/runners/storm/layout/src/main/java/com/hortonworks/streamline/streams/layout/storm/StormTopologyFluxGenerator.java
```diff
@@ -127,10 +128,11 @@ public void visit(final RulesProcessor rulesProcessor) {
...
             // create windowed bolt per unique window configuration
             for (Map.Entry<Window, Collection<Rule>> entry : windowedRules.asMap().entrySet()) {
-                RulesProcessor windowedRulesProcessor = new RulesProcessor(rulesProcessor);
+                RulesProcessor windowedRulesProcessor = copyRulesProcessor(rulesProcessor);
+                windowedRulesProcessorId++;
                 windowedRulesProcessor.setRules(new ArrayList<>(entry.getValue()));
-                windowedRulesProcessor.setId(rulesProcessor.getId() + "." + ++windowedRulesProcessorId);
-                windowedRulesProcessor.setName("WindowedRulesProcessor");
+                windowedRulesProcessor.setId(rulesProcessor.getId() + "." + windowedRulesProcessorId);
+                windowedRulesProcessor.setName(rulesProcessor.getName() + "$$windowed-" + windowedRulesProcessorId);
```
HeartSaVioR: @arunmahadevan @shahsank3t 
Due to avoid conflict component name in Storm topology, I added `$$windowed-<seq>` after component name.
The name will be logged to test mode event logger, and also affects Storm worker log.
(I mean executor thread name will be changed.)

So either event logger in backend or UI should remove $$ part of bolt name to match it with actual component. 
(@shahsank3t Please let me know your thought about this.)

We can't modify executor thread name hence we should take different approach for log search what we are exploring.

shahsank3t: @HeartSaVioR I think removing $$ on the backend will be better since on UI looping through all the events and removing $$ might be costlier op than doing on the backend.

HeartSaVioR: @shahsank3t Agreed, and the change is specific to Storm so better to hide the details to UI side. I’ll try to handle it from backend side. Thanks for the input.

HeartSaVioR: This also affects metrics part.

There's a bug on metrics due to previous logic: we named all the windowed bolt to `WindowedRulesProcessor` which is different from Streamline component name hence we couldn't match the metrics to component.

The change names the component name differently, but even with this patch we will not be able to figure out exact Storm component name for runtime bolt. This affects both metrics from Storm REST API and metrics from Time-series DB.

I don't see clear way to do it, so maybe we could file an issue for this for now and address from other issue.

HeartSaVioR: Filed #985

arunmahadevan:
@HeartSaVioR , will the concerns raised in the above comments get addressed with https://github.com/hortonworks/streamline/issues/985 ?

HeartSaVioR:
@arunmahadevan Yes, and for test mode this patch contains necessary change. We may need to fix it again based on how we decided from #985, but this patch will be OK for now.

PR #966, opened 2017-10-17
ISSUE-964 Inject correlation information to event

HeartSaVioR:
* wrapping collector to add event correlation information before writing event & emit
  * do some hacks around TestRunWindowProcessorBolt because of Storm's behavior
    * WindowedOutputCollector in WindowedBoltExecutor automatically adds inputTuples as anchor which can't be tracked from wrapped collectors
* provide event information via:
  * constructing event tree based on events and provide full event tree or subtree
  * providing all correlated events group by component
* provide some HTTP API regarding above features
  * full event tree associated to given root event
  * sub event tree associated to given root event and start from given sub root event
  * add another HTTP API: provide all correlated events group by component
    * /api/v1/catalog/topologies/:topologyId/testhistories/:historyId/events/correlated/:rootEventId
      * (event id -> event) map for looking up all events
      * (component -> (input event ids, output event ids)) map for grouping events by component

PR #979, opened 2017-10-25
ISSUE-964 Test mode: Add event correlation

sameer79:
Integrated new API's provided in #966 and now showing events on respective component nodes itself.
Multiple events on same component nodes have their own pagination.
Each root source events have their own pagination at the bottom.


HeartSaVioR:
@sameer79 
You may want to notice that the component name in event is where we emit the event, not where it receives the event. So in any case no event will be recorded in sinks. Maybe need to decide which is more intuitive: `show events on emitting source` or `show events on received target`.

@arunmahadevan @shahsank3t Any thoughts?

shahsank3t:
@HeartSaVioR @sameer79 @arunmahadevan 
I feel we should show events on received target but also have emitting source(just for the source component).
This will help the user to see what data is being passed from the source component to another and if there's any inconsistency then they know it's because of the previous node.
Does it make sense?

HeartSaVioR:
@sameer79 
I can give an example topology with some test records:

[Streaming-Analytics-Trucking-Ref-App.json.txt](https://github.com/hortonworks/streamline/files/1416971/Streaming-Analytics-Trucking-Ref-App.json.txt)

Please rename above file to remove '.txt' at the end of file name.

> Test data for `Test-GeoStream`:

```
[
{"eventTime": "2017-09-26 14:54:44.451", "eventSource": "truck_geo_event", "truckId": 23, "driverId": 13, "driverName": "Suresh Srinivas", "routeId": 4, "route": "Des Moines to Chicago", "eventType": "Overspeed", "latitude": 41.62, "longitude": -93.58, "correlationId": 1, "geoAddress": ""},
{"eventTime": "2017-09-26 14:54:50.451", "eventSource": "truck_geo_event", "truckId": 23, "driverId": 13, "driverName": "Suresh Srinivas", "routeId": 4, "route": "Des Moines to Chicago", "eventType": "Excessive Breaking", "latitude": 41.62, "longitude": -94.58, "correlationId": 1, "geoAddress": ""}
]
```

> Test data for `Test-SpeedStream`:

```
[
{"eventTime": "2017-09-26 14:54:45.163", "eventSource": "truck_speed_event", "truckId": 23, "driverId": 13, "driverName": "Suresh Srinivas", "routeId": 4, "route": "Des Moines to Chicago", "speed": 83},
{"eventTime": "2017-09-26 14:54:52.099", "eventSource": "truck_speed_event", "truckId": 23, "driverId": 13, "driverName": "Suresh Srinivas", "routeId": 4, "route": "Des Moines to Chicago", "speed": 96}
]
```

And also set timeout period to around 200 seconds since it has several windowed bolts which are having tumbling windows with 10 sec.

HeartSaVioR:
Maybe we can add more test events to above two sources and see how we show the parents of aggregated event.

arunmahadevan:
@shahsank3t @HeartSaVioR 

I think it makes sense to display the event at source also and then how it got transformed downstream till the sink. Placing in the receiving component looks ok.

We can try out a few different flows like below and see how it shows up.

1. Join (e.g. Kafka1, Kafka2 -> Join)
2. Branch (E.g. Rule -> Notification + Hbase)
3. Linear

sameer79:
There's a multiple event use case that needs to be fixed which I'll be trying to get it done by tomorrow. Not sure if it's related to data being sent by backend or UI issue.

@arunmahadevan @HeartSaVioR @shahsank3t could you please test out if there's anything else missing while I try to debug more on the above-mentioned use case. Thanks in advance.

HeartSaVioR:
@sameer79 It would be better you can share exported topology and test events so that I can also check it via backend side. I'll also test the patch. Thanks!

HeartSaVioR:
@sameer79 @arunmahadevan 
I found the backend issue which aggregation doesn't anchor origin tuples which makes aggregated tuple (event) becoming a new root tuple (event). Going to fix it in backend PR.

I also found UI side issue which always requests subtree based on same event (regardless of which event I'm seeing) for multiple events in component. I guess that's what @sameer79 commented.

There's another UI side Issue that event is showing in source component rather than target component. Please refer previous comments in this PR.

HeartSaVioR:
@arunmahadevan @shahsank3t 
I worked with restricting event "tree" with a root event, hence event flow is shown below:

screen shot 2017-11-01 at 6 30 40 am

Please note that parents with different root IDs are not tracked in tree to make things simpler. If we want to track all, we should find all the parents in the path 'recursively' which may end up exploring most of the events for some cases.

Consider below two cases:
```
Source1 -> Rule1 -> Agg -> Join -> Projection -> Sink
```

```
Source1 -> Rule1 -> Agg    -> Join -> Projection -> Sink
Source2 ->               Rule2  ->
```

Both case needs tracking multiple root events to track the all parents, and I'm not sure there's a clear way to represent it (in a response) at once and also represent in UI.

HeartSaVioR:
@arunmahadevan 
I realized that WindowedOutputCollector in WindowedBoltExecutor automatically adds inputTuples as anchor which can't be tracked from wrapped collectors. I can call emit with anchoring tuples in WindowRulesBolt to fix it, but not sure that's preferred. Otherwise we should hack passed output collector in WindowRulesBolt: wrap collector in delegated collector in passed output collector, and inject it.

Please let me know which way is preferred. Thanks in advance.

HeartSaVioR:
> There's another UI side Issue that event is showing in source component rather than target component. Please refer previous comments in this PR.

I found it somewhat confusing with representing multiple events. Not sure which is better. Only showing them in edge looks intuitive.

HeartSaVioR:
I just found very odd backend issue which I don't have any idea why it happened.

screen shot 2017-11-01 at 10 53 01 am

LOCATION emits events which is corresponding to definition of NOTIFICATION. I'm not aware of this.

@arunmahadevan Is it intended behavior?

HeartSaVioR:
> I can call emit with anchoring tuples in WindowRulesBolt to fix it, but not sure that's preferred. Otherwise we should hack passed output collector in TestRunWindowProcessorBolt: wrap collector in delegated collector in passed output collector, and inject it.

Confirmed that both approaches work. Latter approach needs dirty hack, but the change would be more flexible compared to former since we are fixing TestRunWindowProcessorBolt which will be applied for all processors which needs windowing. Former approach requires fixing all the runtime windowed bolts.

I'll apply the latter approach to backend PR, but we could change it.

HeartSaVioR:
@sameer79 Please rebase your branch to get changes on latest #966. This will resolve aggregation issue on backend. Please note that I'm also squashing the commits into one from #966.

arunmahadevan:
@HeartSaVioR 

> LOCATION emits events which is corresponding to definition of NOTIFICATION. I'm not aware of this.

Yes this is intended, but shouldn't the event be displayed in the target (notification processor) as we had discussed earlier ?

>I'll apply the latter approach to backend PR, but we could change it.

you mean overriding emit/emitDirect in EventCorrelatingOutputCollector before passing it to delegate as its in the PR? It should be fine.

HeartSaVioR:
@arunmahadevan 

>> I'll apply the latter approach to backend PR, but we could change it.
> you mean overriding emit/emitDirect in EventCorrelatingOutputCollector before passing it to delegate as its in the PR? It should be fine.

No decorating order should be changed. Please refer below link:

https://github.com/HeartSaVioR/streamline/blob/9c5d101157d86c911abb7bcc29b0db43ec025a1a/streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/testing/TestRunWindowProcessorBolt.java#L51-L67

HeartSaVioR:
@arunmahadevan 

>> LOCATION emits events which is corresponding to definition of NOTIFICATION. I'm not aware of this.

> Yes this is intended, but shouldn't the event be displayed in the target (notification processor) as we had discussed earlier ?

I already commented earlier:

> There's another UI side Issue that event is showing in source component rather than target component. Please refer previous comments in this PR.

but now I'm not sure it is intuitive with showing multiple events, as I also commented earlier.

sameer79:
@HeartSaVioR @arunmahadevan 
I have fixed issues and addressed your comments above. Please verify it once.

@shahsank3t  Also squashed all the commits and ready for the review.

arunmahadevan:
@HeartSaVioR , one issue I observed is with aggregate, it seems the API is returning all the events from the aggregate even if a query is made to a single root (api/v1/catalog/topologies/:id/testhistories/:id/events/tree/:uuid). Can you check? I tried with an aggregate with two groups of output.

HeartSaVioR:
@arunmahadevan 
I missed the case, and it looks like breaking my assumption for the patch. I was assuming that only contributed tuples are anchored to the new tuple, but it doesn't work like the way I think. 
I'm now checking rules processor deeply, but unfortunately I'm stuck and don't get how aggregation is implemented with rules processor.

HeartSaVioR:
@arunmahadevan 
Looks like using any of engines (groovy, SQL) acts as "blackbok" for getting tuples which contribute an aggregated event.
(Single input event is easy because input event must contribute to the new event.)
And also any other processors which doesn't emit the event with contributed event(s) are affected, CustomProcessor(s) might be another candidate(s).

HeartSaVioR:
Sorry I fixed a bug from backend PR and rebased again. Downloading event file should work now.

sameer79:
Updated the UI based on new API response.
source => shows the output Events
processors & sink => shows the input Events

@arunmahadevan @HeartSaVioR 
@shahsank3t Also squashed all the commits and ready for the review.

arunmahadevan:
Thanks @sameer79, @shahsank3t  for the work. Looks good for merging to master.

PR #982, opened 2017-10-25
ISSUE-970: Make input/output schema and config fields optional for CP

priyank5485:
ISSUE-970: Make input/output schema and config fields optional for CP

### DIFF
--- a/streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/processor/CustomProcessorInfo.java
+++ b/streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/processor/CustomProcessorInfo.java
@@ -201,9 +201,9 @@ public class CustomProcessorInfo {
         uiFields.add(this.createUIField(CUSTOM_PROCESSOR_IMPL, CUSTOM_PROCESSOR_IMPL, false, false, "Custom processor interface implementation class",
                 ComponentUISpecification.UIFieldType.STRING, this.customProcessorImpl));
         ObjectMapper objectMapper = new ObjectMapper();
-        uiFields.add(this.createUIField(INPUT_SCHEMA, INPUT_SCHEMA, false, false, "Custom processor input schema", ComponentUISpecification
+        uiFields.add(this.createUIField(INPUT_SCHEMA, INPUT_SCHEMA, true, false, "Custom processor input schema", ComponentUISpecification
                 .UIFieldType.STRING, objectMapper.writeValueAsString(this.inputSchema)));
-        uiFields.add(this.createUIField(OUTPUT_SCHEMA, OUTPUT_SCHEMA, false, false, "Custom processor output schema",
+        uiFields.add(this.createUIField(OUTPUT_SCHEMA, OUTPUT_SCHEMA, true, false, "Custom processor output schema",
                 ComponentUISpecification.UIFieldType.STRING, objectMapper.writeValueAsString(this.outputSchema)));
         ComponentUISpecification componentUISpecification = new ComponentUISpecification();
         componentUISpecification.setFields(uiFields);
@@ -222,7 +222,7 @@ public class CustomProcessorInfo {
             }
         }
         result.setFields(fields);
-        return result;
+        return (result.getFields().size() == 0 ? null : result);
     }
 
     private List<ComponentUISpecification.UIField> getCustomProcessorUIFieldsWithPrefix () {
--- a/streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/processor/CustomProcessorInfo.java
+++ b/streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/processor/CustomProcessorInfo.java
@@ -222,7 +222,7 @@ public class CustomProcessorInfo {
             }
         }
         result.setFields(fields);
-        return (result.getFields().size() == 0 ? null : result);
+        return (result.getFields().isEmpty() ? null : result);
     }
 
     private List<ComponentUISpecification.UIField> getCustomProcessorUIFieldsWithPrefix () {
--- a/streams/runners/storm/layout/src/main/java/com/hortonworks/streamline/streams/layout/storm/CustomProcessorBoltFluxComponent.java
+++ b/streams/runners/storm/layout/src/main/java/com/hortonworks/streamline/streams/layout/storm/CustomProcessorBoltFluxComponent.java
@@ -39,14 +39,13 @@ public class CustomProcessorBoltFluxComponent extends AbstractFluxComponent {
     protected void generateComponent () {
         String boltId = "customProcessorBolt" + UUID_FOR_COMPONENTS;
         String boltClassName = "com.hortonworks.streamline.streams.runtime.storm.bolt.CustomProcessorBolt";
-        String[] configMethodNames = {"customProcessorImpl", "outputSchema", "inputSchema", "config", "inputSchemaMap"};
+        String[] configMethodNames = {"customProcessorImpl", "outputSchema", "config", "inputSchemaMap"};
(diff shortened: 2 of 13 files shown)


priyank5485:
@arunmahadevan  Can you please review the back end changes when you get a chance?

streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/processor/CustomProcessorInfo.java
```diff
@@ -222,7 +222,7 @@ private ComponentUISpecification getCustomProcessorUISpecification (ComponentUIS
             }
         }
         result.setFields(fields);
-        return result;
+        return (result.getFields().size() == 0 ? null : result);
```
arunmahadevan: nit: use fields.isEmpty()
Q: will it cause NPE at the caller? maybe better to return an Optional.

streams/runners/storm/layout/src/main/java/com/hortonworks/streamline/streams/layout/storm/CustomProcessorBoltFluxComponent.java
```diff
@@ -39,14 +39,13 @@
     protected void generateComponent () {
         String boltId = "customProcessorBolt" + UUID_FOR_COMPONENTS;
         String boltClassName = "com.hortonworks.streamline.streams.runtime.storm.bolt.CustomProcessorBolt";
-        String[] configMethodNames = {"customProcessorImpl", "outputSchema", "inputSchema", "config", "inputSchemaMap"};
+        String[] configMethodNames = {"customProcessorImpl", "outputSchema", "config", "inputSchemaMap"};
         Object[] values = new Object[configMethodNames.length];
         values[0] = conf.get(TopologyLayoutConstants.JSON_KEY_CUSTOM_PROCESSOR_IMPL);
         try {
             values[1] = getOutputSchemaJson();
```
arunmahadevan: getInputSchemaJson is removed but not getOutputSchemaJson but we want to make both optional ?

streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/bolt/CustomProcessorBolt.java
```diff
@@ -203,30 +170,35 @@ public void prepare (Map stormConf, TopologyContext context, OutputCollector col
...
         try {
-            Map<String, Object> mappedEvent = new HashMap<>();
-            if (!inputSchemaMap.containsKey(input.getSourceStreamId()))
-                throw new RuntimeException("Received an event on an input stream that does not have mapping for input schema fields");
-            // Create a new mapped event based on mapping to pass it to CP implementation
-            for (Map.Entry<String, String> entry: inputSchemaMap.get(input.getSourceStreamId()).entrySet()) {
-                if (event.get(entry.getValue()) != null) {
-                    mappedEvent.put(entry.getKey(), event.get(entry.getValue()));
+            StreamlineEvent toProcess = null;
+            if (inputSchemaMap == null || inputSchemaMap.isEmpty() || !inputSchemaMap.containsKey(input.getSourceStreamId())) {
```
arunmahadevan: 1. Can it ever be empty or null? This can be a private final initialized to an empty map.
2. use get to retrieve the mapping based on the stream id before the `if` and check for null and reuse it in the else.

streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/bolt/CustomProcessorBolt.java
```diff
@@ -203,30 +170,35 @@ public void prepare (Map stormConf, TopologyContext context, OutputCollector col
...
+                    }
                 }
+                toProcess = new StreamlineEventImpl(mappedEvent, event.getDataSourceId(), event.getId(), event.getHeader(), input.getSourceStreamId(), event
+                        .getAuxiliaryFieldsAndValues());
             }
-            List<StreamlineEvent> results = customProcessorRuntime.process(new StreamlineEventImpl(mappedEvent, event.getDataSourceId(), event
-                    .getId(), event.getHeader(), input.getSourceStreamId(), event.getAuxiliaryFieldsAndValues()));
+            List<StreamlineEvent> results = customProcessorRuntime.process(toProcess);
             if (results != null) {
+                Schema schema = outputSchema.values().iterator().next();
```
arunmahadevan: 1. Can there be only one output stream and schema? Is that going to be addressed separately?
2. Isn;t the output schema going to be optional as a part of this PR? if so this may throw an Exception.

streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/processor/CustomProcessorInfo.java
priyank5485: Changed to isEmpty.
No, it wont cause a NPE. Its a private method and the only call is in this class that relies on it returning null.

streams/runners/storm/layout/src/main/java/com/hortonworks/streamline/streams/layout/storm/CustomProcessorBoltFluxComponent.java
priyank5485: Those were config methods to the bolt. InputSchema to the bolt before was never used. So it was an unused instance variable in the bolt. Moreover, it did not make sense to pass it since its provided during CP registration initially. Hence removed that variable and the corresponding method. OutputSchema here is the output schema from the bolt configured by user in SAM app using UI. We need it to declare fields from the CustomProcessorBolt.

streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/bolt/CustomProcessorBolt.java
priyank5485: Yes, it can be empty or null. In case of a CP registered with optional input schema, UI will not pass any mapping and hence it will be empty or null. In this case we pass the event received in the CustomProcessorBolt as is to the CP implementation. Your suggestion is better in general, but I think in this case it will make the code a bit complicated. If you are okay, we can leave it as is.

streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/bolt/CustomProcessorBolt.java
priyank5485: Yes, its only one output schema and stream for now. It can be addressed separately based on the need for that use case. I did not plan to address it immediately.
Output schema is optional for registering the CP. This is the output schema at the component level in a SAM app. UI will force the user to create an output schema and hence this will have exactly one entry.

priyank5485:
@arunmahadevan Thanks for the review. Sorry for delayed response on review. For some reason, I dont get notifications if people comment on my PR. And yes, we are making input schema, output schema and config fields optional while registering a CP. Config fields was already optional but it still had the red asterisk showing in UI. We will just remove that with this PR.

priyank5485:
@arunmahadevan I have pushed a commit addressing your review comments and UI changes for this. Let me know if you think we need any more changes for the back end. At this point, I just need to test this out on a cluster

priyank5485:
@arunmahadevan I have rebased and tested this on a cluster. Looks good. Please review when you get a chance.

arunmahadevan:
+1

arunmahadevan:
Thanks @priyank5485 , merged to master and HDF-3.1.0.0.  You can close the PR and the related ISSUE.

PR #983, opened 2017-10-25
ISSUE-981: Re-use custom processor jars using digest

priyank5485:
ISSUE-981: Re-use custom processor jars using digest


priyank5485:
@arunmahadevan Can you please review the back end changes when you get a chance? Will apply UI patches as a different commit to this PR when they are ready.

priyank5485:
@arunmahadevan @shahsank3t I have tested and applied UI changes for this as well.

priyank5485:
@arunmahadevan Please review when you get a chance. I have rebased, added upgrade script and tested everything using mysql implementation. 
@shahsank3t I resolved UI conflicts myself that were introduced due to merging of https://github.com/hortonworks/streamline/pull/982/files Can you please review that conflict resolution?

priyank5485:
@satishd I have rebased. Can you please review and merge if okay?

streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/service/StreamCatalogService.java
```diff
@@ -1088,7 +1107,6 @@ public CustomProcessorInfo removeCustomProcessorInfoAsBundle(String name) throws
             }
         }
         CustomProcessorInfo customProcessorInfo = new CustomProcessorInfo();
-        deleteFileFromStorage(customProcessorInfo.fromTopologyComponentBundle(customProcessorBundle).getJarFileName());
```
satishd: When are uploaded custom processor jars removed? We may want to remove them when none of the custom processors refer an uploaded jar.

streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/processor/CustomProcessorInfo.java
```diff
@@ -161,6 +174,16 @@ public void setTopologyComponentUISpecification(ComponentUISpecification compone
...
+    public String getDigest() {
+        return digest;
+    }
+
+    @JsonIgnore
+    public void setDigest (String digest) {
+        this.digest = digest;
+    }
+
     public CustomProcessorInfo fromTopologyComponentBundle (TopologyComponentBundle topologyComponentBundle) throws IOException {
```
satishd: You may want to refactor this to factory method instead of instance method and have a private no args constructor to control on how this instance is created. One public/package constructor with the required fields can be declared to create an instance from tests or other scenarios.

streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/service/StreamCatalogService.java
```diff
@@ -1088,7 +1107,6 @@ public CustomProcessorInfo removeCustomProcessorInfoAsBundle(String name) throws
...
         this.removeTopologyComponentBundle(customProcessorBundle.getId());
         return customProcessorInfo;
```
satishd: Empty `customProcessorInfo` is getting returned from here with these changes. We should not allow default no args constructor for `CustomProcessorInfo` to be accessible. It is good to have factory method to create instance from TopologyComponentBundle as I mentioned in other comment in `CustomProcessorInfo`

bootstrap/bootstrap.sh
```diff
@@ -67,4 +67,4 @@ case "${opt}" in
     printUsage
     exit 1
     ;;
-esac
\ No newline at end of file
+esac
```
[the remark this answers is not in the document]
priyank5485: I use vim which and have line separator set to LF and previously it seems that it was CR LF. I cant bring that character back to make the diff go away. This seems to be a character introduced by some other editor. I think we can leave it as is and in general we should just use LF as line separator.

bootstrap/shell/v002__update_defaults.sh
```diff
@@ -258,6 +265,10 @@ function main {
 
     update_bundles
     add_udfs
+    #For HDf-3.0.x line we have don't store digest for custom processor jars.
```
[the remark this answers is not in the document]
priyank5485: Will remove the comment.

streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/processor/CustomProcessorInfo.java
priyank5485: @satishd This object is also used in json serialization and deserialization during HTTP requests. Not sure if a private constructor will work with that or not. Also, a public constructor with required fields will make it a big constructor with lot of arguments. Unless you see a significant advantage of using a static factory method over instance method, I would like to keep  it as it is since that change involves testing everything all over again. We can revisit later if needed.

streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/service/StreamCatalogService.java
priyank5485: Previously jar file name was expected to be unique and provided by user. So i deleted them without any check. Now, as you mentioned it is shared and can be deleted only when its not referenced by any.  That would need an extra query to check. At the same time what is some other connection sends a POST or PUT request for a CP that has a jar with the same digest. Its hard to control that scenario. Need to think through. Moreover, this was modeled after UDFs and Notifiers which also dont delete the jar files if its not referenced anymore. It was discussed in design that we will leave the jar files for now. We can address deleting the jar files later for both UDFs and CPs.

streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/service/StreamCatalogService.java
priyank5485: I had removed it since i did not think it was needed anymore as jar file name is not needed for deletion anymore. For now, I have changed it back to not return an empty CustomProcessorInfo

streams/service/src/main/java/com/hortonworks/streamline/streams/service/CustomProcessorUploadHandler.java
```diff
@@ -142,17 +136,36 @@ private CustomProcessorInfo getCustomProcessorInfo (File tarFile) {
...
         if (jarFileBytes != null && jarFileBytes.length > 0) {
             is = new ByteArrayInputStream(jarFileBytes);
         } else {
-            LOG.warn(customProcessorInfo.getJarFileName() + " not present in tar file: " + tarFile);
+            LOG.warn("Custom processor jar file not present in tar file: " + tarFile);
         }
         return is;
     }
 
+    private byte[] getJarFileAsByteArray (File tarFile) {
```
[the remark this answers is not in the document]
priyank5485: Good suggestion. Will change it.

priyank5485:
@satishd I have addressed your review comments. Can you please check again?

streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/service/StreamCatalogService.java
satishd: Can we file github issue to address this?

bootstrap/bootstrap.sh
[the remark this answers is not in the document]
satishd: You can revert this file not to have any change from this commit.

streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/processor/CustomProcessorInfo.java
satishd: @priyank5485 private constructor works fine with json serialization as it is done in other places in streamline and registry. 
Having static factory method will not allow you to create an empty instance and get into issues like [here](https://github.com/hortonworks/streamline/pull/983/files/728c371e0b51c9028bf2c0f7d10646f2fcef3afc#diff-6d94aae217fe8e21348f49c18e1ba329L1091). I m fine if we want to revisit this later.

streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/service/StreamCatalogService.java
```diff
@@ -1108,7 +1108,7 @@ public CustomProcessorInfo removeCustomProcessorInfoAsBundle(String name) throws
         }
         CustomProcessorInfo customProcessorInfo = new CustomProcessorInfo();
         this.removeTopologyComponentBundle(customProcessorBundle.getId());
-        return customProcessorInfo;
+        return customProcessorInfo.fromTopologyComponentBundle(customProcessorBundle);
```
satishd: nit: Can we have this as below as created instance is not used else where?
```
new CustomProcessorInfo().fromTopologyComponentBundle(customProcessorBundle);
```

satishd:
@priyank5485 I have replied to your earlier comments and left minot nit on the later commit.

streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/processor/CustomProcessorInfo.java
priyank5485: Ok. I will made a note of it and take it up as part of other CP enhancements for next release.

streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/service/StreamCatalogService.java
priyank5485: Raised https://github.com/hortonworks/streamline/issues/1056

bootstrap/bootstrap.sh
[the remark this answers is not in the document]
priyank5485: Done

streams/catalog/src/main/java/com/hortonworks/streamline/streams/catalog/service/StreamCatalogService.java
priyank5485: Will do

priyank5485:
@satishd Thanks for your review. I have addressed all your comments at this point.

Issue #973, opened 2017-10-24
Realtime join processor UI bugs


shahsank3t:
**Bugs:**
1. In Output Fields Selection, if you select a field with the same name from both streams, it only accepts one... easily seen by doing SelectAll option. The backend allows disambiguation of names using the "streamname:fieldname" syntax.
2. Pressing OK button does not save and dismiss the dialog box (Do not select "unique" checkbox to reproduce this)
3. Do we have support or nested fields on UI? Backend allows nested fields using dot syntax. For eg: "streamname:fieldname" or ""streamname:nested.field.name"

**Minor changes:**
1. Please change all 4 occurrences of the word "Buffer" to "Window".
2. Rename "Conditional Fields" to "Join Criteria"

shahsank3t:
@satishwaghela can you look it, please?

shahsank3t:
@satishwaghela 
Adding another point in bugs:
4. Should default to fields grouping and the fields should be based on the fields used in the join condition used.

roshannaik:
@shahsank3t  there is a UI bug wrt to how the output schema is shown when selecting nested fields.  this applies to both Join and newer RTJoin. 
When selecting a leaf node, it should not display all the upper nesting levels. The leaf node should just appear as top level node in the output schema.

**More generally**, since we are allowing selection of at any level of a nested hierarchy, you should display everything from the selected level and any child levels inside it.

Assuming the input has a nested field like this:   orginfo { busines_unit { department }, location }
- If you select  orginfo ... then orginfo and everything under it should show up preserving the hierarch.
- if you select business_info, then businessinfo and everything under it will show up as a top level entity. It wont get nested under orginfo.
- if you select both leaf nodes location & department then both of them should shows up as a independent top level entities. they should not be put under their parent levels.

Currently selecting department, will show this hierarchy  orginfo -> busines_unit -> department  in the output schema.

Ping me if you need further clarification.

roshannaik:
Here are some of the UI changes i can think of that we need to fix:
- The new alias-ing ability (from Join Processor), and reflecting the alias in downstream component's input schema.
- Handling of nested fields .. in output fields selection. (same as Join Processor)
- Applying fields grouping to the input streams based on the join keys specified in UI.
- Rename the component to 'Rolling Window Join'
- Please change all 4 occurrences of the word "Buffer" to "Window".
- Rename "Conditional Fields" to "Join Criteria"

*Note*: We are not enabling event-time support for this right now.

[...]

PR #991, opened 2017-10-27
ISSUE-968 Join processor output when both streams have same field

sameer79:
Join processor support for multiple streams with same fields name by using a unique alias.
the screenshot below:

joinscreenshot

Stream legends are used to denote which field belong to which stream.
We can have in the way it’s shown in the screenshot or add an icon beside the “Output Fields*” label where a user can hover and check.
I am preferring the former approach so the user doesn’t have to hover over the icon every now and then to verify as it will always be available on the screen.

Open for a suggestion from others.

@shahsank3t  @arunmahadevan  @roshannaik please review and comment.
Thanks in advanced

### DIFF
--- a/webservice/src/main/resources/app/scripts/components/SelectInputComponent.jsx
+++ b/webservice/src/main/resources/app/scripts/components/SelectInputComponent.jsx
@@ -1,0 +1,146 @@
+/**
+  * Copyright 2017 Hortonworks.
+  *
+  * Licensed under the Apache License, Version 2.0 (the "License");
+  * you may not use this file except in compliance with the License.
+  * You may obtain a copy of the License at
+  *   http://www.apache.org/licenses/LICENSE-2.0
+  * Unless required by applicable law or agreed to in writing, software
+  * distributed under the License is distributed on an "AS IS" BASIS,
+  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  * See the License for the specific language governing permissions and
+  * limitations under the License.
+**/
+
+import React, {Component} from 'react';
+import _  from 'lodash';
+import {OverlayTrigger, Popover} from 'react-bootstrap';
+import Utils from '../utils/Utils';
+
+export default class SelectInputComponent extends Component{
+  constructor(props) {
+    super(props);
+    this.filterNode = [];
+    this.state = {
+      aliasValue : props.node.alias
+    };
+  }
+
+  componentDidMount(){
+    this.highlightNode();
+  }
+
+  componentDidUpdate(){
+    this.highlightNode();
+    this.valueCheck();
+  }
+
+  valueCheck = () => {
+    const {node} = this.props;
+    if(this.state.aliasValue === ''){
+      this.setInputBoxColor(node);
+    } else {
+      this.validate(node,this.state.aliasValue);
+    }
+  }
+
+  validate = (node,val) => {
+    if(!Utils.noSpecialCharString(val)){
+      this.setInputBoxColor(node);
+    }
+  }
+
+  highlightNode = () => {
+    const {outputKeysObjArr,node,legends} = this.props;
+    this.filterNode = _.filter(outputKeysObjArr, (o) =>  o.alias === node.alias);
+    if(this.filterNode.length > 1){
+      _.map(this.filterNode,(n) => {
+        const pEl = this.getParentElement(n);
+        let color = "#676767";
+        if(n.alias === node.alias){
+          color = "#fd5454";
+        }
+        if(pEl !== null && pEl !== undefined){
+          const gParent = this.getGrandParentElement(pEl);
+          gParent.style['background-color'] =  color;
+        }
+      });
+    } else {
+      const single =  this.filterNode[0];
+      const parent =  this.getParentElement(single);
+      const gParent = this.getGrandParentElement(parent);
+      gParent.style['background-color'] =  "#676767";
+    }
+  }
+
+  getParentElement = (n) => {
+    return  this.refs['pEl_'+n.keyPath+'_'+n.name];
+  }
+
+  getGrandParentElement = (parent) => {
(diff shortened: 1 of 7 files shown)


roshannaik:
I like the way the duplicate alias reporting is done and also the stream numbering technique .. though it took me a little time to understand how the stream numbers/legend were employed.

Few things I noticed:
Need to relocate the 'Select All' link for the output fields selection box to its bottom right or much closer to it. Given its current proximity to the 'slide interval', it does not give the impression that it is actually related to the output field selector box.
No error indication if 'timestamp field' is not filled and you click OK.
Need better tooltips for:
+ **Timestamp field** : "Field (of type long) indicating the event's time'
+ **Lag ms** : "Max time to wait after window close, for a late arriving event to be part of its window" @arunmahadevan  could please you review this tooltip hint.
What's the way to reorder the output fields easily ?

sameer79:
@roshannaik
Addressed the comment and updated the PR as well.

@shahsank3t  please review.

arunmahadevan:
- Timestamp : "A 'long' field in the event that represents the event-timestamp in milliseconds (epoch-time)'
- Lag ms : "Maximum lag of the event timestamp in milliseconds up to which a late arriving event is included in the current window"

I am not sure what you mean by output stream legends, I think you mean aliases for input streams similar to table alias in SQL join.

"Input stream aliases"
  - window_transform_stream_16 -> window1 (user provided value)
  - window_transform_stream_16 -> window2 (user provided value)

then in the select box,

window1.field1 as f1 
window2.field1 as f2

and so on. If you want to auto-generate aliases that may also be fine, but user may not be able to quickly identify which is 1, 2 and so on.

Also in the generated event I assume it will be referred as "f1" and "f2". You may also need to validate that two fields don't end up having the same alias.

Are these mandatory? By default we could assume input_stream_name.field_name and let user specify aliases only if they want.

roshannaik:
@arunmahadevan , it would be helpful to clarify the word 'lag' here. Basically brief wording to indicate ... starting from when is the lag time calculated.

arunmahadevan:
Lag ms : "Maximum lag of the event timestamp (beyond the window end time) in milliseconds for including a late arriving event in the current window"

is this better?

arunmahadevan:
@shahsank3t , if the field names are unique across streams do you still want to show aliases for them?. May be just display the white box with the field name `[field_name]`. If the user goes and edits the value display `[1 field_name as (alias)]`. 

For duplicate fields, the proposed UI and forcing the user to remove the conflict looks great.

Assume the legends would be displayed at the bottom all the time. This should be renamed to "Input stream legends" or just "Legends: 1 -> foo, 2 -> bar" should be sufficient.

roshannaik:
@arunmahadevan We should allow aliases even if there is no duplication. Users should be able to rename the output fields as they please. Can be quite useful in many cases like when dealing with nested fields.

More refinement to the wording for Lag Ms. See if it looks ok...

**Lag Ms** :  If (time of window close + Lag)  <= timestamp of a late arriving event for same window, the event will be considered for processing.

arunmahadevan:
>We should allow aliases even if there is no duplication. Users should be able to rename the output fields as they please.

Yes by default display the unique "fieldname" and if the user clicks and renames it display the "fieldname as alias". Like the SQL select query where its not mandatory to have alias if the field names are unique.

arunmahadevan:
LagMs, "event time" should be less than "window close + lag" not other way.. I think if we can put it in words rather than less/greater than it would be easier to understand. Lets choose one and go with it..

roshannaik:
Sorry, I meant >=.  Understanding this lag is a bit tricky for new folks due to the mix of wall clock time and timestamps going into it, so whatever we can do to help convey it succinctly would be nice.  Here it is in a more verbal style ...

**Lag Ms** : If timestamp of a late arriving event is less than the wall clock time at which its window closed + Lag, the event will still be considered for processing in that window.

arunmahadevan:
its not the wall clock time, event time.
"If the timestamp of a late arriving event is not "lag" milliseconds beyond the window end time (based on event time), the event will still be considered for processing in that window."

sameer79:
Address comment: https://github.com/hortonworks/streamline/issues/973#issuecomment-341848438

Change in Timestamp fields dropdown
i] mandatary to select one field for per input stream.

and updated PR @arunmahadevan @roshannaik @shahsank3t  please review and comment.

roshannaik:
@shahsank3t 
- need to add validation that window interval and slide interval are > 0
- change caption 'WINDOW INTERVAL TYPE' to 'WINDOW TYPE'

sameer79:
@roshannaik address comment and updated PR as well.

@shahsank3t squash the commit and ready for review and comment.

sameer79:
@roshannaik  Address comment https://github.com/hortonworks/streamline/pull/1004#issuecomment-346198546

@shahsank3t updated PR please review and comment.

arunmahadevan:
@roshannaik, can you comment if the issue you had observed has been addressed now ?

roshannaik:
UI changes look ok... doing some more testing.

roshannaik:
@sameer79 . One more thing. Looks like the same validation from Sliding Interval field has also been accidentally applied to Lag Milliseconds field as well (along with the same error msg).  Lag Ms should allow values >=0.  unlike the Sliding Interval which cannot have 0.

sameer79:
@roshannaik Addressed the comment and updated PR as well.

roshannaik:
latest change has broken the 'timestamp fields' 
screen shot 2017-11-29 at 6 59 29 am

roshannaik:
Tested successfully. Backend looks good. Rebased.
Minor issue in UI needs fix.

sameer79:
@roshannaik  'timestamp fields' issue is fixed. please verify it, Thanks in advance.

roshannaik:
+1  
UI has turned out pretty nice. thanks !

Issue #994, opened 2017-10-30
Topology Name should not have spaces in it.


shahsank3t:
For the new topologies to be created/imported, we shall not allow space in the name.

As per discussion with @arunmahadevan , for the old topologies that has space in it, when the user goes to edit mode, we will show some pop-up and force them to change the name.

shahsank3t:
https://github.com/hortonworks/streamline/issues/994 fixes this.

PR #995, opened 2017-10-31
ISSUE-994: Topology Name should not have spaces in it.

shahsank3t:
This fixes https://github.com/hortonworks/streamline/issues/994


arunmahadevan:
@shahsank3t , can you rebase with master and merge? If the name already had spaces assume it would force the user to rename it when they edit.

arunmahadevan:
+1

shahsank3t:
@arunmahadevan 

> If the name already had spaces assume it would force the user to rename it when they edit.

Yes, we ask the user to remove space again and only then let them proceed.

rebased it.
Once Travis build check is passed, will merge it in.

PR #998, opened 2017-11-02
ISSUE-996 Removed findbugs-annotation 1.3.2 module dependency as it has LGPL license.

satishd:


arunmahadevan:
The findbugs-annotation library is still shipped (findbugs-annotations-1.3.9-1.jar) due to hbase-client. But this library has group id com.github.stephenc.findbugs with Apache license so should be fine.

+1

PR #1004, opened 2017-11-08
ISSUE-1003 - Support for event-time windows in Join processor

roshannaik:
@shahsank3t can you test UI changes using this PR ?

- Sample windowConfig  json, passed by UI to backend for EvenTime :

```json
"window" : {"windowLength" : {"class":".Window$Duration", "durationMs":1000},
            "slidingInterval" : {"class":".Window$Duration", "durationMs":1000},
            "tsFields": ["s1:tsField1" , "s2:tsField2", "s3:field3.innerTsField"], //new
            "lagMs": 10,
            "lateStream": "optionalLateStream"  //new
 }
```
- Note: Need to ensure there is one timestamp field per input stream when using event time joins

### DIFF
--- a/streams/layout/src/main/java/com/hortonworks/streamline/streams/layout/component/rule/expression/Window.java
+++ b/streams/layout/src/main/java/com/hortonworks/streamline/streams/layout/component/rule/expression/Window.java
@@ -20,6 +20,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 
 import java.io.IOException;
 import java.io.Serializable;
+import java.util.Arrays;
 
 /**
  * Captures the windowing parameters. Internally this implies the rule has to be evaluated as an
@@ -108,7 +109,9 @@ public class Window implements Serializable {
     private WindowParam windowLength;
     private WindowParam slidingInterval;
     private String tsField;
+    private String[] tsFields;
     private int lagMs;
+    private String lateStream;
 
     // for jackson
     private Window() {
@@ -124,7 +127,9 @@ public class Window implements Serializable {
         this.windowLength = other.getWindowLength();
         this.slidingInterval = other.getSlidingInterval();
         this.tsField = other.getTsField();
+        this.tsFields = other.tsFields;
         this.lagMs = other.getLagMs();
+        this.lateStream = other.lateStream;
     }
 
     public WindowParam getWindowLength() {
@@ -151,12 +156,28 @@ public class Window implements Serializable {
         this.tsField = tsField;
     }
 
+    public String[] getTsFields() {
+        return tsFields.clone();
+    }
+
+    public void setTsFields(String[] tsFields) {
+        this.tsFields = tsFields.clone();
+    }
+
     public int getLagMs() {
         return lagMs;
     }
 
     public void setLagMs(int lagMs) {
         this.lagMs = lagMs;
+    }
+
+    public void setLateStream(String stream) {
+        lateStream = stream;
+    }
+
+    public String getLateStream() {
+        return lateStream;
     }
 
     @Override
@@ -190,7 +211,9 @@ public class Window implements Serializable {
                 "windowLength=" + windowLength +
                 ", slidingInterval=" + slidingInterval +
                 ", tsField='" + tsField + '\'' +
+                ", tsFields='[" + Arrays.toString(tsFields) + "\']" +
                 ", lagMs=" + lagMs +
+                ", lateStream=" + lateStream +
                 '}';
     }
 }
--- a/streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/bolt/query/SLMultistreamTimestampExtractor.java
+++ b/streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/bolt/query/SLMultistreamTimestampExtractor.java
@@ -1,0 +1,59 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License
+ */
+
+package com.hortonworks.streamline.streams.runtime.storm.bolt.query;
+
+import com.hortonworks.streamline.streams.runtime.storm.bolt.query.RealtimeJoinBolt.StreamKind;
+import org.apache.storm.tuple.Tuple;
+import org.apache.storm.windowing.TimestampExtractor;
+
+import java.util.HashMap;
+
+
+public class SLMultistreamTimestampExtractor implements TimestampExtractor {
+    HashMap<String,FieldSelector> tsFields = new HashMap<>(); // stream name -> field selector
+
+    /** Field selectors should be qualified with stream name and can refer to nested fields. Aliases if any will be ignored.
+     *     Ex: "stream1:tsField"  or  "stream1:outerField.TsField"
+     *  'streamline-event.' prefix will be added to the events automatically
+     */
+    public SLMultistreamTimestampExtractor(String... fieldSelectors) {
+        for (int i = 0; i < fieldSelectors.length; i++) {
+            String fieldSel = WindowedQueryBolt.convertToStreamLineKeys(fieldSelectors[i]); // Strip any alias & prefix field with 'streamline-event.'
+            FieldSelector fs = new FieldSelector(fieldSel, StreamKind.STREAM);
+            if ( tsFields.containsKey(fs.streamName) ) {
+                throw new IllegalArgumentException("Only one timestamp field allowed per stream");
+            }
+            tsFields.put(fs.streamName, fs);
+        }
+    }
+
+    @Override
+    public long extractTimestamp(Tuple tuple) {
+        FieldSelector tsFieldSel = tsFields.get( tuple.getSourceStreamId() );
+        if (tsFieldSel == null) {
+            throw new IllegalArgumentException("Unrecognized source stream in Event: '" + tuple.getSourceStreamId() + "'");
+        }
+        Object tsField = tsFieldSel.findField(tuple);
+        if (tsField == null) {
+            throw new IllegalArgumentException("Timestamp field not found in event : '" + tsFieldSel.outputName + "'");
+        }
+        return Long.parseLong(tsField.toString());
+    }
+
+}
--- a/streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/bolt/query/WindowedQueryBolt.java
+++ b/streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/bolt/query/WindowedQueryBolt.java
@@ -20,6 +20,7 @@ package com.hortonworks.streamline.streams.runtime.storm.bolt.query;
 import com.hortonworks.streamline.streams.StreamlineEvent;
 import com.hortonworks.streamline.streams.common.StreamlineEventImpl;
 import com.hortonworks.streamline.streams.layout.component.rule.expression.Window;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.storm.topology.OutputFieldsDeclarer;
 import org.apache.storm.topology.base.BaseWindowedBolt;
 import org.apache.storm.tuple.Fields;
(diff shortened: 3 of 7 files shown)


shahsank3t:
@roshannaik 
What is `lateStream` in the JSON? Currently, we have nothing like that on UI.

arunmahadevan:
@roshannaik , are these changes backward compatible? It will be better if it works seamlessly for users who already have aggregate and join processor in their SAM app and migrate to HDF 3.1.0. If you can accept the tsField as a string itself and handle both single field plus and new format it may work. Otherwise we will have to provide migration scripts which will get tricky. 

For the new field (lateStream) we should handle the cases where late stream is missing in the stored config so that existing saved configs in DB will continue to work with HDF 3.1 without any migration.

streams/layout/src/main/java/com/hortonworks/streamline/streams/layout/component/rule/expression/Window.java
```diff
@@ -108,7 +108,9 @@ public String toString() {
     private WindowParam windowLength;
     private WindowParam slidingInterval;
     private String tsField;
+    private String[] tsFields;
```
arunmahadevan: have either tsField or tsFields not both. Make sure it works for both Aggregate and Join.

roshannaik:
@arunmahadevan 
Yes these enhancements are backward compatible and limited to Join processor. Does not interfere with any other components.

@shahsank3t 
`lateStream` is a feature of Storm's windowing subsystem, specific to event-time windows. In short .... if an event arrives too late (after its window is already processed) then such late events are simply emitted to a lateStream *if* it is configured. If it is not configured, then the event is simply logged and dropped. Such late events don't go to the regular output stream.  In principle any component dealing with event-time windows should support this.

The late stream would be an "special" but optional 2nd output stream for Join processor (only available for event-time joins).

streams/layout/src/main/java/com/hortonworks/streamline/streams/layout/component/rule/expression/Window.java
roshannaik: Aggregate doesn't (and shouldnt) use tsFields.  
Join doesn't use tsField.  This doesn't break Aggregate.

roshannaik:
@shahsank3t,  you may want to tackle the other stuff before exposing `lateStream`, given that it is an optional setting, things will still work without it being exposed.

streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/bolt/query/SLMultistreamTimestampExtractor.java
```diff
@@ -0,0 +1,59 @@
...
+    public long extractTimestamp(Tuple tuple) {
+        FieldSelector tsFieldSel = tsFields.get( tuple.getSourceStreamId() );
+        if (tsFieldSel == null) {
+            throw new IllegalArgumentException("Unrecognized source stream in Event: '" + tuple.getSourceStreamId() + "'");
+        }
+        Object tsField = tsFieldSel.findField(tuple);
+        if (tsField == null) {
+            throw new IllegalArgumentException("Timestamp field not found in event : '" + tsFieldSel.outputName + "'");
+        }
+        return Long.parseLong(tsField.toString());
```
arunmahadevan: shouldn't the timestamp field be a long value. if not we should catch it above and throw exception.

streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/bolt/query/WindowedQueryBolt.java
```diff
@@ -198,8 +195,12 @@ public void withWindowConfig(Window windowConfig) throws IOException {
             withLag(new BaseWindowedBolt.Duration(windowConfig.getLagMs(), TimeUnit.MILLISECONDS));
         }
 
-        if (windowConfig.getTsField() != null) {
-            withTimestampField(windowConfig.getTsField());
+        if (windowConfig.getTsFields() != null  &&  windowConfig.getTsFields().length>0) {
```
arunmahadevan: What about the existing Join bolts having tsField in the config. How would that get handled ?

roshannaik: Existing Join processor does not have tsField in the config from UI.

streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/bolt/query/SLMultistreamTimestampExtractor.java
roshannaik: Even thought the SAM UI will say type is long, in practice timestamp data will typically be represented in Integer format. It will not have the L suffix.

This feature when moved to Storm's JoinBolt (base class), timestamp can be in any format that can be converted to Long.

SAM UI may not allow strings, but back-end will. On the UI we could open it up to integers as well if needed.

arunmahadevan: We expect the field to contain timestamp value of epoch time in milliseconds, I don't think that can be in integer format. If its not long it will be better to throw exception than silently processing it. We have had issues where some customer was passing timestamp in seconds (which could fit into integer) instead of milliseconds.

arunmahadevan:
@roshannaik , can you rebase this and update when its ready to test? Will try out the changes.

streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/bolt/query/WindowedQueryBolt.java
arunmahadevan: You mean the condition `windowConfig.getTsField()` was always null earlier? If so its ok.

arunmahadevan:
@shahsank3t , can you try it out from the UI side?

@roshannaik , can you rebase this with master ?

streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/bolt/query/SLMultistreamTimestampExtractor.java
arunmahadevan: @roshannaik , anything other than long does not seem to make sense here. Can you do an `instanceOf` check to ensure the field is `long` and avoid any conversions?

streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/bolt/query/WindowedQueryBolt.java
roshannaik: right.

roshannaik:
rebased it.

arunmahadevan:
@roshannaik  can you address https://github.com/hortonworks/streamline/pull/1004/files#r150451212 ?

arunmahadevan:
@roshannaik , can you try out if the end-end runtime flow after applying https://github.com/hortonworks/streamline/pull/991 ?

@shahsank3t has already verified the UI-backend interaction. 

Once you confirm I will merge it.

arunmahadevan:
@roshannaik , I am also seeing the following findbugs warnings when I do a "mvn clean install". Can you address that as well?

```
[INFO] BugInstance size is 3
[INFO] Error size is 0
[INFO] Total bugs: 3
[INFO] Invocation of toString on Window.tsFields in com.hortonworks.streamline.streams.layout.component.rule.expression.Window.toString() [com.hortonworks.streamline.streams.layout.component.rule.expression.Window] At Window.java:[line 209] DMI_INVOKING_TOSTRING_ON_ARRAY
[INFO] com.hortonworks.streamline.streams.layout.component.rule.expression.Window.getTsFields() may expose internal representation by returning Window.tsFields [com.hortonworks.streamline.streams.layout.component.rule.expression.Window] At Window.java:[line 159] EI_EXPOSE_REP
[INFO] com.hortonworks.streamline.streams.layout.component.rule.expression.Window.setTsFields(String[]) may expose internal representation by storing an externally mutable object into Window.tsFields [com.hortonworks.streamline.streams.layout.component.rule.expression.Window] At Window.java:[line 163] EI_EXPOSE_REP2
[INFO]


To see bug detail using the Findbugs GUI, use the following command "mvn findbugs:gui"
```

roshannaik:
@arunmahadevan ... updated the PR with the fix for that.

arunmahadevan:
@roshannaik ,
I built a sample topology and on deployment I am getting the following exception.
```
Exception in thread "main" java.lang.reflect.InvocationTargetException
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at org.apache.storm.flux.FluxBuilder.buildObject(FluxBuilder.java:310)
	at org.apache.storm.flux.FluxBuilder.buildComponents(FluxBuilder.java:358)
	at org.apache.storm.flux.FluxBuilder.buildTopology(FluxBuilder.java:74)
	at org.apache.storm.flux.Flux.runCli(Flux.java:153)
	at org.apache.storm.flux.Flux.main(Flux.java:98)
Caused by: com.fasterxml.jackson.streamline.databind.exc.UnrecognizedPropertyException: Unrecognized field "tsFields" (class com.hortonworks.streamline.streams.layout.component.rule.expression.Window), not marked as ignorable (4 known properties: "lagMs", "slidingInterval", "windowLength", "tsField"])
 at [Source: {"windowLength":{"class":".Window$Duration","durationMs":5000},"slidingInterval":{"class":".Window$Duration","durationMs":5000},"tsFields":["kafka_stream_1:ts","kafka_stream_2:ts"],"lagMs":0}; line: 1, column: 141] (through reference chain: com.hortonworks.streamline.streams.layout.component.rule.expression.Window["tsFields"])
	at com.fasterxml.jackson.streamline.databind.exc.UnrecognizedPropertyException.from(UnrecognizedPropertyException.java:62)
	at com.fasterxml.jackson.streamline.databind.DeserializationContext.reportUnknownProperty(DeserializationContext.java:855)
	at com.fasterxml.jackson.streamline.databind.deser.std.StdDeserializer.handleUnknownProperty(StdDeserializer.java:1083)
	at com.fasterxml.jackson.streamline.databind.deser.BeanDeserializerBase.handleUnknownProperty(BeanDeserializerBase.java:1389)
	at com.fasterxml.jackson.streamline.databind.deser.BeanDeserializerBase.handleUnknownVanilla(BeanDeserializerBase.java:1367)
	at com.fasterxml.jackson.streamline.databind.deser.BeanDeserializer.vanillaDeserialize(BeanDeserializer.java:266)
	at com.fasterxml.jackson.streamline.databind.deser.BeanDeserializer.deserialize(BeanDeserializer.java:125)
	at com.fasterxml.jackson.streamline.databind.ObjectMapper._readMapAndClose(ObjectMapper.java:3789)
	at com.fasterxml.jackson.streamline.databind.ObjectMapper.readValue(ObjectMapper.java:2779)
	at com.hortonworks.streamline.streams.layout.component.rule.expression.Window.<init>(Window.java:119)
	... 9 more
```

Can you fix this?

arunmahadevan:
@roshannaik , I am also seeing errors and not able to configure an "Aggregate" processor. Heres the response received -

```
(was java.lang.NullPointerException) (through reference chain: com.hortonworks.streamline.streams.layout.component.rule.Rule["window"]->com.hortonworks.streamline.streams.layout.component.rule.expression.Window["tsFields"])
```

arunmahadevan:
@roshannaik , I am able to deploy the Join topology.

Only while creating an Aggregate processor I am not able to save it. @shahsank3t , can you also take a look ? I applied this PR + PR #991 to test.

@roshannaik , confirm once you have tried sending some data to your test topology and processed it with event time.

shahsank3t:
@arunmahadevan I am not able to start streamline server, getting Exception: https://gist.github.com/shahsank3t/1b0faef3c4cdef38ffa6066dc8e3bdfa 
Once that resolves, I'll check Aggregate processor.

arunmahadevan:
@shahsank3t , can you rebuild latest registry locally and then rebuild streamline. A related issue was fixed in registry recently. You may also need to cleanup local registry maven artifacts.

shahsank3t:
@arunmahadevan @roshannaik 
Getting error:
(was java.lang.NullPointerException) (through reference chain: com.hortonworks.streamline.streams.layout.component.rule.Rule["window"]->com.hortonworks.streamline.streams.layout.component.rule.expression.Window["tsFields"])

on PUT request to API: api/v1/catalog/topologies/:topologyID/windows/:windowID

arunmahadevan:
The issue with kafka was env config issue. probably we should add some checks in Ambari to prevent invalid configs. Will raise an issue for that

I think now you should be able to see the events flowing in runtime into the Join processor. Will wait for your test results.

roshannaik:
@shahsank3t There is UI bug.  When i use aliasing to disambiguate output fields in the join processor, the downstream KafkaSink's input schema does not reflect the aliased names. it shows the unaliased names. here screen shots of the schema mismatch.
joinoutputschema
kafkasinkinputschema

shahsank3t:
@sameer79 could you please fix the bug @roshannaik mentioned above, in your existing PR https://github.com/hortonworks/streamline/pull/991

arunmahadevan:
@roshannaik , i have a patch for addressing the aggregate processor issue I will merge it along with this once you have tested the Join runtime.

sameer79:
@roshannaik address the issue and fixed in PR https://github.com/hortonworks/streamline/pull/991

arunmahadevan:
@roshannaik , can you comment if you have tested the runtime part and the UI issues you encountered in https://github.com/hortonworks/streamline/pull/1004#issuecomment-346198546 have been addressed?

roshannaik:
Was able to test both backend and front end. Backend is looking good. One minor issue with validation. Have rebased this PR.

roshannaik:
The UI issues are fixed in the PR. Verified again that both UI pr & this backend PR are now working fine.

PR #1009, opened 2017-11-11
ISSUE-1008 Highlight source which emits selected root event in test e…

HeartSaVioR:
…vent correlation result

* add 'containingSelectedEvent' to component: the value is true when selected event is emitted in that component
* ensure selected event is always placed first in source's output events
* add unit tests

Only output of HTTP API is changed. See below:

screen shot 2017-11-11 at 9 53 01 pm

### DIFF
--- a/streams/common/src/main/java/com/hortonworks/streamline/streams/common/event/correlation/GroupedCorrelationEvents.java
+++ b/streams/common/src/main/java/com/hortonworks/streamline/streams/common/event/correlation/GroupedCorrelationEvents.java
@@ -15,48 +15,56 @@
  **/
 package com.hortonworks.streamline.streams.common.event.correlation;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.hortonworks.streamline.streams.common.event.EventInformation;
 
 import java.util.*;
 
+import static java.util.stream.Collectors.toMap;
+
 public class GroupedCorrelationEvents {
     private Map<String, EventInformation> allEvents;
-    private Map<String, ComponentGroupedEvents> componentGroupedEvents;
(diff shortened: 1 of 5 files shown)


streams/common/src/main/java/com/hortonworks/streamline/streams/common/event/correlation/GroupedCorrelationEvents.java
```diff
@@ -89,4 +97,48 @@ public String getComponentName() {
...
+            this.componentName = groupedEvents.getComponentName();
+
+            this.inputEventIds = new ArrayList<>(groupedEvents.getInputEventIds());
+            this.outputEventIds = new ArrayList<>(groupedEvents.getOutputEventIds());
+
+            if (this.outputEventIds.contains(selectedEventId)) {
+                // this component contains selected event Id
+                this.containingSelectedEvent = true;
+
+                this.outputEventIds.remove(selectedEventId);
```
arunmahadevan: nit: can be LinkedList since we are adding and removing entries?

HeartSaVioR: Element count in array should be super small enough so that's trivial, but the change is also trivial I'll apply it.

HeartSaVioR: Addressed.

arunmahadevan:
Thanks @HeartSaVioR , merged to master. @shahsank3t can you take up UI changes for this in a separate PR ?

shahsank3t:
@arunmahadevan yes @sameer79 has picked up and will be doing UI changes in a separate PR.

PR #1013, opened 2017-11-15
ISSUE-999: Search for component names in application editor doesn't seem to search custom processors

sameer79:
In Application Editor:
Not able to search by component name for the custom processors, has been fixed in this PR.

@shahsank3t  please review and comment.

### DIFF
--- a/webservice/src/main/resources/app/scripts/components/SpotlightSearch.jsx
+++ b/webservice/src/main/resources/app/scripts/components/SpotlightSearch.jsx
@@ -144,7 +144,24 @@ export default class SpotlightSearch extends Component {
     update the entity list with matching components
   */
   handleValueChange = (e) => {
-    let filteredEntities = Utils.filterByName(this.state.entities, e.target.value.trim());
+    const {entities} = this.state;
+    const filterValue = e.target.value.trim();
+    let filteredEntities = Utils.filterByName(entities, filterValue);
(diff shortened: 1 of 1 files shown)


webservice/src/main/resources/app/scripts/components/SpotlightSearch.jsx
```diff
@@ -144,7 +144,23 @@ export default class SpotlightSearch extends Component {
...
-    let filteredEntities = Utils.filterByName(this.state.entities, e.target.value.trim());
+    const {entities} = this.state;
+    const filterValue = e.target.value.trim();
+    let filteredEntities = Utils.filterByName(entities, filterValue);
+    const customProcessors = _.filter(entities, {'subType' : 'CUSTOM'});
+    /* when the normal search return empty we again run search on
+      custom processors "topologyComponentUISpecification.fields" filter by name
+      to get the custom processors default name while it was created.. 
+    */
+    if(filteredEntities.length === 0 && customProcessors.length){
```
shahsank3t: we don't need `filteredEntities.length` check. Think of the scenario where the CP name starts with something similar to any other components (Eg: Rule_Custom_Processor) and if user searches for `Rule`, we need to show both Rule processor and this Rule_Custom_Processor CP. With this check that will always be avoided.

sameer79:
Address comment and update PR @shahsank3t  please review and comment.

PR #1016, opened 2017-11-16
ISSUE-1014 Dynamically changing log level at Application level

HeartSaVioR:
* add HTTP API endpoints to get/modify log level at Application level
  * GET /api/v1/catalog/topologies/:topologyId/logconfig
  * POST /api/v1/catalog/topologies/:topologyId/logconfig
    * Params: targetLogLevel, durationSecs

`targetLogLevel` would be one of these: TRACE, DEBUG, INFO, WARN, ERROR

We only use `ROOT` as Logger name, as we would want to change log level at application level.


HeartSaVioR:
Result of both GET and POST HTTP API endpoint would be below:

```
{
    "logLevel": "DEBUG",
    "epoch": 1510813008068
}
```

shahsank3t:
@HeartSaVioR Do we also have an API to change log level per component?

HeartSaVioR:
@shahsank3t 
I found technical issue on that so I discussed with @arunmahadevan, and we decided it is not possible to change log level per component. Changing log level should not be in component configuration, and current log level per component will not be also available.

arunmahadevan:
+1

PR #1018, opened 2017-11-18
ISSUE-1017 Place event first which is directly related to selected event

### LINKED ISSUE #1017: Place event first which is directly related to selected event
Currently we are highlighting the source and placing the event to the first for selected root event. There's subsequent requirement that placing the event to the first for selected root event other than source.

To add my idea, we can place events to the first on components which is directly related to select event, so that it can reflect complex topology like below:

```
S1 -> Projection -> Aggregation -> Join -> Sink
S                               ->
```

If we select the event which is emitted from S1, events in S1, Projection, Aggregation, Join will be reordered. Sink is not reordered given that the event is co-related (both from S1 and S2).

HeartSaVioR:
* reorder input and output events based on the rule:
  * 1: selected event
  * 2: events which contains selected event in root ids
  * 3: other events

No UI change needed.

This closes #1017


HeartSaVioR:
Please take a look at newly added test case: it represents the case which considering root id helps on reordering events.

arunmahadevan:
+1

Issue #1023, opened 2017-11-21
Provide HTTP API endpoints for merging component catalog information and metrics


HeartSaVioR:
We've been working on view mode of topology, which provides overall look of topology structure as well as runtime information (mostly metrics).

To merge topology structure with metrics, we would want to provide unified API endpoint which provides merged result of separate API (topology catalog, metrics).

Due to the request of flexible time window in view mode, metrics should be queried as time-series manner, which can't be done with Storm REST API. So all the API endpoints will only work properly when time-series DB is properly set to the namespace.

HeartSaVioR:
#1032

HeartSaVioR:
#1032 was merged. Closing.

PR #1025, opened 2017-11-21
ISSUE-1024 Refactor TopologyCatalogResource (also API endpoint paths)

### LINKED ISSUE #1024: Refactor TopologyCatalogResource (also API endpoint paths)
Currently TopologyCatalogResource serves more than catalog specific. 
I think two major things are out of catalog:

1. API endpoints for TopologyAction
2. "details" option, and corresponding options

We can move them to other new resource classes, with changing API endpoint path if needed.
Especially I'd rather not using flag option. I feel that was not good decision which we can separate API endpoint to clarify the meaning.
(I'd also add the API endpoints for #1023 rather than adding flag option again.)

@arunmahadevan @shahsank3t Please let me know your opinion. That may bring UI side change since API endpoint may also be changed.

HeartSaVioR:
* split up TopologyCatalogResource into three pieces
  * new resources: TopologyDashboardResource, TopologyActionResource
* change some of REST API endpoints which are moved to TopologyDashboardResource
* modify UI side as well

This closes #1024.

### DIFF
--- a/streams/service/src/main/java/com/hortonworks/streamline/streams/service/CatalogResourceUtil.java
+++ b/streams/service/src/main/java/com/hortonworks/streamline/streams/service/CatalogResourceUtil.java
@@ -47,13 +47,13 @@ public final class CatalogResourceUtil {
     }
 
     @JsonInclude(JsonInclude.Include.NON_NULL)
-    static class TopologyDetailedResponse {
+    static class TopologyDashboardResponse {
         private final Topology topology;
         private final TopologyRunningStatus running;
         private final String namespaceName;
         private TopologyRuntimeResponse runtime;
 
-        public TopologyDetailedResponse(Topology topology, TopologyRunningStatus running, String namespaceName) {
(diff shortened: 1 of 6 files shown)


HeartSaVioR:
Copying comment from #1024 given that I've not address it yet and would like to hear opinion before making progress.

https://github.com/hortonworks/streamline/issues/1024#issuecomment-345925613

> I'd also like to not mix up actions for runtime and catalog: import, export, clone has actions in the path on API endpoint, which may conflict with TopologyAction.
If we want to keep actions in the path for them, I'd rather like to change TopologyAction to TopologyRuntime, and also make changes on API endpoint paths and module names respectively, cause they're related to runtime.

HeartSaVioR:
Manually tested. 

I found another API endpoints (cluster, namespace) which also rely on flag option as well, but this issue was brought from #1023 so would want to address it later if we agree with removing flag option.

arunmahadevan:
+1

PR #1028, opened 2017-11-21
ISSUE-1026: Group test mode events in the bottom nav bar

### LINKED ISSUE #1026: Group test mode events in the bottom nav bar
Right now the individual root events are displayed in the bottom bar. This is not a very intuitive way for users to navigate the events because the same set of events can repeat at the sources.

1. Add an API to return group of related events
2. UI displays the bottom panel based on the groups of events.

arunmahadevan:
Return set of related source events in the root events API

closes #1026

### DIFF
--- a/streams/common/src/main/java/com/hortonworks/streamline/streams/common/event/correlation/CorrelatedEventsGrouper.java
+++ b/streams/common/src/main/java/com/hortonworks/streamline/streams/common/event/correlation/CorrelatedEventsGrouper.java
@@ -15,8 +15,11 @@
  **/
 package com.hortonworks.streamline.streams.common.event.correlation;
 
+import com.google.common.collect.HashMultimap;
+import com.google.common.collect.Multimap;
 import com.hortonworks.streamline.streams.common.event.EventInformation;
 
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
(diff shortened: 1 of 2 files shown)


arunmahadevan:
@shahsank3t , we need to display related group of events in the bottom nav bar. Can you try it out?

shahsank3t:
@arunmahadevan 
Sure will try it out today.

shahsank3t:
@arunmahadevan I tried this PR and the response looks pretty good to me.

arunmahadevan:
>(So if event 1~3 is grouped, which event to be based on might show different events placement.)

@HeartSaVioR , We will remove the "Source" highlighting feature. Lets say
S1 and S2 are sources and e1 - e3 is one of the groups, the display would be like

S1 [e1]
S2 [e2, e3]

With the grouped approach, the user cannot not then select some event at the sources (say e2) and expect the first source to automatically change the events displayed (say e1, e4). To see this user would have to click at the next group in the bottom panel (which would be [e1, e4, e2, e3] assuming e2 combined with e1 and e4 from source 1). Here I see some redundancy, but I am not sure how  this can avoided.

HeartSaVioR:
@arunmahadevan 
Please note that UI still needs to pick one of them in group and query to backend. (the patch only modifies API endpoint for listing root events)

So assuming below topology (that's so far uncommon but...):

```
Source1 -> Aggregate1 -> Join1 -> Sink1
Source2 -> Branch1    -/
                      -\ Aggregate2 -> Sink2
```

S1 [:A1,A2,A3]
Aggregate1 [A1,A2,A3:A4]
Join1 [A4,B4:AB1], [A4,B6:AB2] (or so)
Sink1 [AB1,AB2:]

S2 [:B1,B2,B3]
Branch1 [B1:B4 (join), B5(aggr)], [B2:B6 (join), B7 (aggr)], [B3:B8 (join), B9 (aggr)]
Aggregate2 [B5,B7,B9:B10]
Sink2 [B10:]

* Related root events of A1 : A1, A2, A3, B1 (root event of B4)
* Related root events of B1 : A1, A2, A3, B1, B2 (root event of B7), B3 (root event of B9)

We are displaying above groups to different groups - say [A1, A2, A3, B1] and [A1, A2, A3, B1, B2] - but showing events is still depending on which root event is base event.

In other words, in group [A1, A2, A3, B1], if we select A1 as base event in group, there will be no event in Aggregate2 and Sink2, but if we select B1 as base event in group, there will be events in Aggregate2 and Sink2. The API endpoint fills all the parent events (so covered backward) but doesn't fill forward events (doesn't cover forward) for other correlated root events.

arunmahadevan:
@HeartSaVioR , made some changes to include same set multiple times if they produce different resultant events on varying the root event. The returned list of eventsets would have the relevant root event id as the first event. 

Can you review the latest commit?

streams/common/src/main/java/com/hortonworks/streamline/streams/common/event/correlation/CorrelatedEventsGrouper.java
```diff
@@ -35,27 +37,29 @@ public CorrelatedEventsGrouper(List<EventInformation> events) {
...
      * @return the group of correlated events per source
      */
-    public Set<Set<String>> groupByRelatedSourceEvents(Set<String> sources) {
-        Set<Set<String>> result = new LinkedHashSet<>();
+    public List<Set<String>> groupByRelatedSourceEvents(Set<String> sources) {
+        Multimap<Set<String>, Set<String>> allEventsToSourceEvents = HashMultimap.create();
         Stream<String> rootEventIds = events.stream().filter(e -> e != null && e.getRootIds().isEmpty())
                 .map(EventInformation::getEventId);
         rootEventIds.forEach(rootEventId -> {
             Set<String> relatedSourceEvents = new LinkedHashSet<>();
```
HeartSaVioR: We don't need to collect related source events from allRelatedEvents, given that we are checking all the root events. So all the root events which have same related events should be grouped together, and naturally they should be correlated even we don't check it, because A, B, C are grouped, their same related events should have A, B, C.

HeartSaVioR: So it will work if you group root events which has same related events. If the group is [A, B, C], regardless of selection (one of A, B, C) the same events DAG will be shown.

arunmahadevan: I made some more changes to group all root events that result in same "allEvents". Please check.

HeartSaVioR:
cc. @sameer79 Please refer the last comment.

PR #1031, opened 2017-11-22
ISSUE-1030: Add runtime event sampling support

arunmahadevan:
- New APIs and support for enabling and disabling sampling of events at runtime
- Added APIs to fetch the sampled events

### DIFF
--- a/streams/common/src/main/java/com/hortonworks/streamline/streams/common/StreamlineEventImpl.java
+++ b/streams/common/src/main/java/com/hortonworks/streamline/streams/common/StreamlineEventImpl.java
@@ -15,6 +15,10 @@
  **/
 package com.hortonworks.streamline.streams.common;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.google.common.base.Joiner;
 import com.google.common.base.Predicate;
 import com.google.common.collect.Collections2;
@@ -22,6 +26,7 @@ import com.google.common.collect.ForwardingMap;
 import com.google.common.collect.ImmutableMap;
 import com.hortonworks.streamline.streams.StreamlineEvent;
 
+import java.io.IOException;
 import java.io.UnsupportedEncodingException;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -34,11 +39,14 @@ import java.util.UUID;
 /**
  * A default implementation of StreamlineEvent.
  */
+@JsonSerialize(as=StreamlineEventImpl.class)
 public final class StreamlineEventImpl extends ForwardingMap<String, Object> implements StreamlineEvent {
     // Default value chosen to be blank and not the default used in storm since wanted to keep it independent of storm.
     public final static String DEFAULT_SOURCE_STREAM = "default";
     // special event to trigger evaluation of group by
     public static final StreamlineEvent GROUP_BY_TRIGGER_EVENT = StreamlineEventImpl.builder().build();
+
+    public static final String TO_STRING_PREFIX = "StreamlineEventImpl";
 
     private final ImmutableMap<String, Object> header;
     private final String sourceStream;
@@ -50,6 +58,14 @@ public final class StreamlineEventImpl extends ForwardingMap<String, Object> imp
     @Override
     protected Map<String, Object> delegate() {
         return delegate;
+    }
+
+    private StreamlineEventImpl() {
+        header = null;
+        sourceStream = null;
+        auxiliaryFieldsAndValues = null;
+        dataSourceId = null;
+        delegate = null;
     }
 
     /**
@@ -310,15 +326,38 @@ public final class StreamlineEventImpl extends ForwardingMap<String, Object> imp
         throw new UnsupportedOperationException();
     }
 
+    public static StreamlineEvent fromString(String s) {
+        ObjectMapper mapper = new ObjectMapper();
+        try {
+            Map<String, Object> event = mapper.readValue(
+                    s.substring(s.indexOf(TO_STRING_PREFIX) + TO_STRING_PREFIX.length()),
+                    new TypeReference<Map<String, Object>>() {});
+            return StreamlineEventImpl.builder()
+                    .header((Map<String, Object>) event.get("header"))
+                    .sourceStream((String) event.get("sourceStream"))
+                    .auxiliaryFieldsAndValues((Map<String, Object>) event.get("auxiliaryFieldsAndValues"))
+                    .dataSourceId((String) event.get("dataSourceId"))
+                    .putAll((Map<String, Object>) event.get("fieldsAndValues"))
+                    .build();
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
     @Override
     public String toString() {
-        return "StreamlineEventImpl{" +
-                "header=" + header +
-                ", sourceStream='" + sourceStream + '\'' +
-                ", fieldsAndValues=" + super.toString() +
-                ", auxiliaryFieldsAndValues=" + auxiliaryFieldsAndValues +
-                ", dataSourceId='" + dataSourceId + '\'' +
-                ", id='" + id + '\'' +
-                '}';
+        ObjectMapper mapper = new ObjectMapper();
+        try {
+            Map<String, Object> event = new HashMap<>();
+            event.put("header", header);
+            event.put("sourceStream", sourceStream);
+            event.put("auxiliaryFieldsAndValues", auxiliaryFieldsAndValues);
+            event.put("dataSourceId", dataSourceId);
+            event.put("id", id);
+            event.put("fieldsAndValues", delegate);
+            return  TO_STRING_PREFIX + mapper.writeValueAsString(event);
+        } catch (JsonProcessingException ex) {
+            throw new RuntimeException(ex);
+        }
     }
 }
--- a/streams/common/src/main/java/com/hortonworks/streamline/streams/common/StreamlineEventImpl.java
+++ b/streams/common/src/main/java/com/hortonworks/streamline/streams/common/StreamlineEventImpl.java
@@ -46,7 +46,7 @@ public final class StreamlineEventImpl extends ForwardingMap<String, Object> imp
     // special event to trigger evaluation of group by
     public static final StreamlineEvent GROUP_BY_TRIGGER_EVENT = StreamlineEventImpl.builder().build();
 
-    public static final String TO_STRING_PREFIX = "StreamlineEventImpl";
+    public static final String TO_STRING_PREFIX = "StreamlineEvent";
 
     private final ImmutableMap<String, Object> header;
     private final String sourceStream;
--- a/streams/metrics/pom.xml
+++ b/streams/metrics/pom.xml
@@ -27,9 +27,5 @@
             <groupId>com.hortonworks.streamline</groupId>
             <artifactId>streamline-cluster</artifactId>
         </dependency>
-        <dependency>
-            <groupId>com.hortonworks.registries</groupId>
-            <artifactId>storage-core</artifactId>
-        </dependency>
     </dependencies>
 </project>
--- a/streams/runners/storm/sampling/src/main/java/com/hortonworks/streamline/streams/sampling/service/storm/StormTopologySamplingService.java
+++ b/streams/runners/storm/sampling/src/main/java/com/hortonworks/streamline/streams/sampling/service/storm/StormTopologySamplingService.java
@@ -1,0 +1,257 @@
+package com.hortonworks.streamline.streams.sampling.service.storm;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonRawValue;
+import com.hortonworks.streamline.streams.catalog.Topology;
+import com.hortonworks.streamline.streams.catalog.TopologyComponent;
+import com.hortonworks.streamline.streams.layout.TopologyLayoutConstants;
+import com.hortonworks.streamline.streams.sampling.service.TopologySampling;
+import com.hortonworks.streamline.streams.storm.common.StormRestAPIClient;
+import com.hortonworks.streamline.streams.storm.common.StormTopologyUtil;
+import org.apache.commons.lang.StringEscapeUtils;
+import org.apache.commons.lang.StringUtils;
+import org.glassfish.jersey.client.ClientConfig;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.security.auth.Subject;
+import javax.ws.rs.client.Client;
+import javax.ws.rs.client.ClientBuilder;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Date;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static com.hortonworks.streamline.streams.common.StreamlineEventImpl.TO_STRING_PREFIX;
+
+public class StormTopologySamplingService implements TopologySampling {
+    private static final Logger LOG = LoggerFactory.getLogger(StormTopologySamplingService.class);
+
+    private static final Pattern LOG_CONTENT_REGEX = Pattern.compile("<pre id=\"logContent\">(.*)</pre>", Pattern.DOTALL);
+    private static final Pattern PREV_START_REGEX = Pattern.compile("<a class=\"btn btn-default enabled\"[^<]*start=(\\d+)[^<]*Prev</a>", Pattern.DOTALL);
+    private static final Pattern NEXT_START_REGEX = Pattern.compile("<a class=\"btn btn-default enabled\"[^<]*start=(\\d+)[^<]*Next</a>", Pattern.DOTALL);
+    public static final int BYTES_TO_FETCH = 51200;
+
+    private StormRestAPIClient client;
+
+    public StormTopologySamplingService() {
+    }
+
+    @Override
+    public void init(Map<String, Object> conf) {
+        String stormApiRootUrl = null;
+        Subject subject = null;
+        if (conf != null) {
+            stormApiRootUrl = (String) conf.get(TopologyLayoutConstants.STORM_API_ROOT_URL_KEY);
+            subject = (Subject) conf.get(TopologyLayoutConstants.SUBJECT_OBJECT);
+        }
+        Client restClient = ClientBuilder.newClient(new ClientConfig());
+        this.client = new StormRestAPIClient(restClient, stormApiRootUrl, subject);
+    }
+
+    @Override
+    public boolean enableSampling(Topology topology, int pct, String asUser) {
+        String topologyId = StormTopologyUtil.findStormTopologyId(client, topology.getId(), asUser);
+        return client.enableSampling(topologyId, pct, asUser);
+    }
+
+    @Override
+    public boolean enableSampling(Topology topology, TopologyComponent component, int pct, String asUser) {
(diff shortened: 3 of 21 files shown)


arunmahadevan:
**Get sampling status at topology level**
GET http://HOST:PORT/api/v1/sampling/topologies/:TOPOLOGY_ID
```json
{
    "topologyId": 3,
    "pct": 10,
    "enabled": false,
    "success": true
}
```
**Enable sampling at topology level**
POST http://HOST:PORT/api/v1/sampling/topologies/3/enable/100
```json
{
    "topologyId": 3,
    "pct": 100,
    "success": true
}
```

**Disable sampling at topology level**
POST http://HOST:PORT/api/v1/sampling/topologies/3/disable


**Get sampling status at component level**
GET http://HOST:PORT/api/v1/sampling/topologies/3/component/28

**Enable sampling at component level**
POST http://HOST:PORT/api/v1/sampling/topologies/3/component/28/enable/100

**Disable sampling at component level**
POST http://HOST:PORT/api/v1/sampling/topologies/3/component/28/disable


**Get sampled Events (returns the next 10 events)**
http://HOST:PORT/api/v1/sampling/topologies/:TOPOLOGY_ID/component/:COMPONENT_ID/events
```json
{
    "topologyId": 3,
    "componentId": 28,
    "success": true,
    "events": {
        "length": 51200,
        "next": 47008,
        "events": [
            {
                "time": 1511339899000,
                "event": {
                    "dataSourceId": "28",
                    "fieldsAndValues": {
                        "user_id": "Sjb5e5-gKoLXueFDMc2R8Q",
                        "text": "...",
                        "review_id": "2_Ru_ASf75kU303rdQjFfQ",
                        "stars": 4,
                        "date": "2013-04-25",
                        "business_id": "KayYbHCt-RkbGcPdGOThNg",
                        "type": "review",
                        "votes": {
                            "funny": 0,
                            "useful": 0,
                            "cool": 0
                        }
                    },
                    "auxiliaryFieldsAndValues": {},
                    "header": {},
                    "id": "75d1eb87-cd98-43cf-9721-114337cea5ba",
                    "sourceStream": "default"
                }
            },
			..
			..
			]
		}
	}
```

**Get the next set of sampled events**	

Use the next and length from the result of previous query

GET http://HOST:PORT/api/v1/sampling/topologies/:TOPOLOGY_ID/component/:COMPONENT_ID/events?start=47008&length=51200

streams/runners/storm/sampling/src/main/java/com/hortonworks/streamline/streams/sampling/service/storm/StormTopologySamplingService.java
```diff
@@ -0,0 +1,257 @@
...
+        };
+    }
+
+    private void addSampledEvents(String lines, String queriedComponentId, List<SampledEvent> events) {
+        // Timestamp, Component name, Component task-id, MessageId (incase of anchoring), List of emitted values
+        int offset = 0;
+        for (String line : lines.split("\n")) {
+            line = StringEscapeUtils.unescapeHtml(line);
+            int componentStart = line.indexOf(',') + 1;
+            if (componentStart != -1) {
```
HeartSaVioR: Given that you add 1 after `indexOf`, componentStart is in worst case `0`, not `-1`

streams/runners/storm/sampling/src/main/java/com/hortonworks/streamline/streams/sampling/service/storm/StormTopologySamplingService.java
```diff
@@ -0,0 +1,257 @@
...
+                int componentEnd = line.indexOf(',', componentStart);
+                if (componentEnd != -1) {
+                    String componentName = line.substring(componentStart, componentEnd);
+                    try {
+                        if (componentName.equals(queriedComponentId)) {
+                            long time = Date.parse(line.substring(0, componentStart - 1));
+                            String eventStr = line.substring(line.indexOf(TO_STRING_PREFIX) + TO_STRING_PREFIX.length(), line.lastIndexOf("]"));
```
HeartSaVioR: Someone may think it is not necessary, but for me it's better to do defensive programming: finding TO_STRING_PREFIX and "]" before extracting eventStr. If it can't find any of them, leave log message and skip the line.

streams/runners/storm/sampling/src/main/java/com/hortonworks/streamline/streams/sampling/service/storm/StormTopologySamplingService.java
```diff
@@ -0,0 +1,257 @@
...
+        String componentId = component.getId() + "-" + component.getName();
+        Long cur;
+        if (qps.desc()) {
+            cur = qps.start();
+        } else {
+            cur = qps.start() == null ? Long.valueOf(0) : qps.start();
+        }
+        Long nextOffset;
+        Long prevOffset = null;
+        do {
```
HeartSaVioR: Looks like this implementation may not be able to handle corner case: the event being placed to two pages, in other words, the line is exposed partially because we are splitting page by bytes. That's not only the event is hidden, but also the parsing logic for line could be broken.

streams/runners/storm/sampling/src/main/java/com/hortonworks/streamline/streams/sampling/service/storm/StormTopologySamplingService.java
```diff
@@ -0,0 +1,257 @@
...
+                }
+                if (next >= 0) {
+                    length = qps.length();
+                } else if (qps.start() > 0 && qps.start() < qps.length()){
+                    length = events.get(0).getStartOffset() + qps.start().intValue() + 1;
+                } else {
+                    length = events.get(0).getStartOffset() + 1;
+                }
+                next = Math.max(next, 0);
+                Collections.reverse(events);
```
HeartSaVioR: This will not work with multiple pages. 

Suppose first page contains [1, 2, 3, 4, 5] and second page contains [6, 7, 8, 9, 10]. We're fetching page in reverse order so events will be [6, 7, 8, 9, 10, 1, 2, 3, 4, 5] and reversed events will be [5, 4, 3, 2, 1, 10, 9, 8, 7, 6].

Simple fix is adding events per page by reversed order (if `desc()` is true), so that events will be [10, 9, 8, 7, 6] for the first fetch, and [10, 9, 8, 7, 6, 5, 4, 3, 2, 1] for the second fetch.

streams/runners/storm/sampling/src/main/java/com/hortonworks/streamline/streams/sampling/service/storm/StormTopologySamplingService.java
```diff
@@ -0,0 +1,257 @@
...
+        Integer length = null;
+        if (!events.isEmpty()) {
+            if (events.size() > qps.count()) {
+                int i = qps.desc() ? events.size() - qps.count() : 0;
+                int j = qps.desc() ? events.size() : qps.count();
+                events = events.subList(i, j);
+            }
+            if (qps.desc()) {
+                if (qps.start() != null) {
+                    next = qps.start() + events.get(0).getStartOffset() - qps.length() + 1;
```
HeartSaVioR: This doesn't hold true with fetching multiple pages. You may need to play with `cur` instead of `qps.start()`, or modify `addSampledEvents()` to ensure each event has absolute offset rather than relative offset in page. IMHO latter way is clearer.

HeartSaVioR: If we could make event having absolute offset, I don't think we need to calculate with prevOffset too. And the case for calculating `length` should be simplified too, cause we just need to deal with `next` and `qps.length()`: `qps.start()` no needed.

streams/runners/storm/sampling/src/main/java/com/hortonworks/streamline/streams/sampling/service/storm/StormTopologySamplingService.java
HeartSaVioR: And I just found the case above.

shahsank3t:
> Get sampled Events (returns the latest 10 events)
http://HOST:PORT/api/v1/sampling/topologies/:TOPOLOGY_ID/component/:COMPONENT_ID/events

@arunmahadevan Can we have the similar type of API but the one that returns event from all components?

streams/runners/storm/sampling/src/main/java/com/hortonworks/streamline/streams/sampling/service/storm/StormTopologySamplingService.java
HeartSaVioR: Other case than splitted event, now we are relying on actual class (StreamlineEventImpl) rather than interface, so if someone has a need to use other than StreamlineEventImpl, the logic breaks. 
Unfortunately I need to have the implementation other than StreamlineEventImpl due to preserve ID while modifying event, and if we don't want to break the logic then before emitting we should convert it to StreamlineEventImpl again. Losing ID at that time may not matter since we are just emitting, but I'm nervous to add conversion logic to all the emitting places.

streams/common/src/main/java/com/hortonworks/streamline/streams/common/StreamlineEventImpl.java
```diff
@@ -310,15 +326,38 @@ public final void clear() {
         throw new UnsupportedOperationException();
     }
 
+    public static StreamlineEvent fromString(String s) {
+        ObjectMapper mapper = new ObjectMapper();
+        try {
+            Map<String, Object> event = mapper.readValue(
+                    s.substring(s.indexOf(TO_STRING_PREFIX) + TO_STRING_PREFIX.length()),
+                    new TypeReference<Map<String, Object>>() {});
+            return StreamlineEventImpl.builder()
```
HeartSaVioR: You may already notice, the information for event ID is lost while converting. Do we want to have a way to preserve event ID in StreamlineEventImpl at least internally (only available in class)?

streams/metrics/pom.xml
```diff
@@ -27,5 +27,9 @@
             <groupId>com.hortonworks.streamline</groupId>
             <artifactId>streamline-cluster</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.hortonworks.registries</groupId>
```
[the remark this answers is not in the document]
arunmahadevan: Will remove.

streams/runners/storm/sampling/src/main/java/com/hortonworks/streamline/streams/sampling/service/storm/StormTopologySamplingService.java
```diff
@@ -0,0 +1,257 @@
...
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static com.hortonworks.streamline.streams.common.StreamlineEventImpl.TO_STRING_PREFIX;
+
+public class StormTopologySamplingService implements TopologySampling {
+    private static final Logger LOG = LoggerFactory.getLogger(StormTopologySamplingService.class);
+
+    private static final Pattern LOG_CONTENT_REGEX = Pattern.compile("<pre id=\"logContent\">(.*)</pre>", Pattern.DOTALL);
```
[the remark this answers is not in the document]
arunmahadevan: Modified to use Jsoup to parse and navigate the elements similar to DOM.

streams/runners/storm/sampling/src/main/java/com/hortonworks/streamline/streams/sampling/service/storm/StormTopologySamplingService.java
arunmahadevan: yes theres a chance that the partial events at the end can be missed based on the offset and its really difficult to avoid this while handling byte offsets. I have added checks to skip such partial lines. Similar behavior can happen while viewing the event logs via Storm's log viewer as well. 

When we enhance the sampling with event co-relation and implement at SAM level, we would rather store the events in a DB which will not have this limitation and would be easy to query.

streams/runners/storm/sampling/src/main/java/com/hortonworks/streamline/streams/sampling/service/storm/StormTopologySamplingService.java
arunmahadevan: for now i can play with `cur`. During the initial query (without any start and length) the start offset is unknown, so would be difficult to find the absolute offset.

streams/runners/storm/sampling/src/main/java/com/hortonworks/streamline/streams/sampling/service/storm/StormTopologySamplingService.java
arunmahadevan: Nice catch. Will fix.

streams/runners/storm/sampling/src/main/java/com/hortonworks/streamline/streams/sampling/service/storm/StormTopologySamplingService.java
arunmahadevan: thanks for noticing.

streams/runners/storm/sampling/src/main/java/com/hortonworks/streamline/streams/sampling/service/storm/StormTopologySamplingService.java
arunmahadevan: Will change prefix to `StreamlineEvent` so that `IdPreservedStreamlineEvent` is also covered.

streams/common/src/main/java/com/hortonworks/streamline/streams/common/StreamlineEventImpl.java
arunmahadevan: In the sampling case, I am not re-building an event, just passing the string part from the logs as is.

arunmahadevan:
@HeartSaVioR , thanks for the review, addressed the comments. I need to do some more testing but you can take a look again.

HeartSaVioR:
The code looks better. 

I've done basic manual tests:

- Getting status for sampling and enabling & disabiling event sample: OK
- Retrieving events: OK

> /api/v1/sampling/topologies/11/component/31/events?desc=false

```
{
    "topologyId": 11,
    "componentId": 31,
    "success": true,
    "events": {
        "length": 51200,
        "next": 9381,
        "events": [
            {
                "time": 1512972778000,
                "event": {
                    "dataSourceId": "31",
                    "fieldsAndValues": {
                        "user_id": "PUFPaY9KxDAcGqfsorJp3Q",
                        "review_id": "Ya85v4eqdd6k9Od8HbQjyA",
                        "stars": 4,
                        "date": "2012-08-01",
                        "business_id": "5UmKMjUEUNdYWqANhGckJw",
                        "type": "review",
                        "votes": {
                            "funny": 0,
                            "useful": 0,
                            "cool": 0
                        }
                    },
                    "auxiliaryFieldsAndValues": {},
                    "header": {},
                    "id": "30c7c7b9-eaec-4cf3-91f3-7d77e3283bd2",
                    "sourceStream": "default"
                }
            },
...
```

> /api/v1/sampling/topologies/11/component/31/events?desc=false&start=9381&length=51200

```
{
    "topologyId": 11,
    "componentId": 31,
    "success": true,
    "events": {
        "length": 51200,
        "next": 17730,
        "events": [
            {
                "time": 1512972778000,
                "event": {
                    "dataSourceId": "31",
                    "fieldsAndValues": {
                        "user_id": "JPPhyFE-UE453zA6K0TVgw",
                        "review_id": "mjCJR33jvUNt41iJCxDU_g",
                        "stars": 4,
                        "date": "2014-11-28",
                        "business_id": "UsFtqoBl7naz8AVUBZMjQQ",
                        "type": "review",
                        "votes": {
                            "funny": 0,
                            "useful": 0,
                            "cool": 0
                        }
                    },
                    "auxiliaryFieldsAndValues": {},
                    "header": {},
                    "id": "928efcff-98fe-4555-82eb-6b7992c84f7b",
                    "sourceStream": "default"
                }
            },
...
```

> /api/v1/sampling/topologies/11/component/31/events?desc=true

```
{
    "topologyId": 11,
    "componentId": 31,
    "success": true,
    "events": {
        "length": 51200,
        "next": 59276,
        "events": [
            {
                "time": 1512973466000,
                "event": {
                    "dataSourceId": "31",
                    "fieldsAndValues": {
                        "user_id": "bcwr1bFov3PSa1FiGfpc9g",
                        "review_id": "UrukGX1emhSRe2fGdxdVPA",
                        "stars": 3,
                        "date": "2011-12-22",
                        "business_id": "KayYbHCt-RkbGcPdGOThNg",
                        "type": "review",
                        "votes": {
                            "funny": 0,
                            "useful": 0,
                            "cool": 0
                        }
                    },
                    "auxiliaryFieldsAndValues": {},
                    "header": {},
                    "id": "7f88308e-5f3e-47f3-b534-30309f16f0d9",
                    "sourceStream": "default"
                }
            },
...
```

> /api/v1/sampling/topologies/11/component/31/events?desc=true&start=59276&length=51200

```
{
    "topologyId": 11,
    "componentId": 31,
    "success": true,
    "events": {
        "length": 51200,
        "next": 54546,
        "events": [
            {
                "time": 1512973466000,
                "event": {
                    "dataSourceId": "31",
                    "fieldsAndValues": {
                        "user_id": "k0a-uii8yr__Rl-9ipl_RA",
                        "review_id": "e11K8g-0m66gDeRz3TgSIQ",
                        "stars": 5,
                        "date": "2015-11-06",
                        "business_id": "mVHrayjG3uZ_RLHkLj-AMg",
                        "type": "review",
                        "votes": {
                            "funny": 0,
                            "useful": 1,
                            "cool": 1
                        }
                    },
                    "auxiliaryFieldsAndValues": {},
                    "header": {},
                    "id": "37c44352-165d-4384-b3f1-ebde5a96275b",
                    "sourceStream": "default"
                }
            },
...
```

HeartSaVioR:
Btw, I'm assuming that topology.eventlogger.executors is set to 1. Does it work well if it is set to more than 1? Is there any chance for eventlogger executors to be distributed?

Also event logger is disabled by default, we may want to document that `topology.eventlogger.executors` should be set (1, or more if it doesn't break the funtionality).

arunmahadevan:
I will set the default topology setting to topology.eventlogger.executors=1. If its set to more than 1, the logs will get spread out to multiple logs, but it should still work since we do fields grouping on the component name. So all samples from a component would always end up on a single log.

arunmahadevan:
@HeartSaVioR , thanks for taking time to test out the API. I have added a SAM topology setting for number of event sampling tasks (default = 1). 

@shahsank3t , right now the api is at component level since the storm samples the events based on the component. Theres no storm API to get events from all components. With topology level, we need to aggregate events from multiple component logs, but its not clear how to navigate it, which ones to display first and so on. These issues would be fixed when we develop event-correlation. Till then I think its better to have the user select a component first and view the samples for that component.

arunmahadevan:
The new ask is to be able to group (or search) the samples based on a correlation Id. We have added correlation Id as a part of PR #1080 . 

So the scope of this PR is now to find a grok pattern to parse the event sample log in the current format (or make changes to Storm to use a compatible format if not feasible). After that we need to provide APIs on top of log search to search the samples based on query string.

HeartSaVioR:
From the last comment on https://github.com/hortonworks/streamline/pull/1080#issuecomment-351533430

I filed issue on Storm side: https://issues.apache.org/jira/browse/STORM-2854 and will file issue for implementing custom IEventLogger on Streamline side. 

Since new requirement and implementing custom IEventLogger is related, we can move the new requirement to new issue and handle them altogether.

@arunmahadevan 
I guess we could remove the event fetch part on this patch (since it would be replaced to Solr query) and merge this sooner. You are also welcome to leave this as it is and handle the new approach of event fetching by yourself. Either is fine for me, and I can follow up your decision.

arunmahadevan:
@HeartSaVioR , removed the part to fetch the events from storm event logs. I left a few placeholders so that it can be easily changed to match the new requirements. Please review and merge if it looks fine. The new logic can be submitted as a separate PR.

PR #1037, opened 2017-11-25
ISSUE-1033: Implement UI part of Log Search feature

satishwaghela:
This PR contains UI implementation of Log Search feature, to view this page use url: http://host:port/#/logsearch/:topologyid


HeartSaVioR:
@satishwaghela 
Sorry I missed one thing to add from UI side: currently we update `timeSeriesDB` in namespace to `AMBARI_METRICS` when `AMBARI_METRICS` is selected in namespace. We need to update `logSearchService` in namespace when `AMBARI_INFRA` is selected in namespace.
Please apply above requirement. Thanks in advance!

satishwaghela:
@HeartSaVioR 
I have addressed your comment. please review changes

HeartSaVioR:
Screenshots:

screen shot 2017-11-27 at 3 59 40 pm
screen shot 2017-11-27 at 4 00 43 pm

shahsank3t:
> 1. maybe better to have filtering columns, given that host:port might be somewhat long so that it occupies the area of log message.

Yes, we can add a filter dropdown above the table to show/hide columns and by default, we can hide host and port columns and let the user choose them when need be.

> 2. host and port may be better to in separated columns.

I agree on this.

cc: @satishwaghela @HeartSaVioR  @arunmahadevan

satishwaghela:
@HeartSaVioR @shahsank3t 
I addressed your comment. I have added column selection feature in which user can show/hide columns.

satishwaghela:
@HeartSaVioR 
We should not reduce margin because this is standard margin which we have used in all the pages so either we can use full screen means no margin like we did in application edit mode page or we can keep as it is.

HeartSaVioR:
@satishwaghela OK. I think we could use full screen to show as much information as we can, but let's keep it as it is unless we have consensus around changing this.

shahsank3t:
@arunmahadevan This PR contains changes only for Log Search page and not the enabling logs in view mode. Changes you requested should be for https://github.com/hortonworks/streamline/pull/1055

arunmahadevan:
OK then we can merge this and the enable-disable setting can be addressed in #1055

arunmahadevan:
+1

PR #1045, opened 2017-11-29
ISSUE-1044: Add support to fetch nested fields in DruidEventMapper

omkreddy:


omkreddy:
@arunmahadevan pinging for review

shahsank3t:
@omkreddy Do this need any change in UI?

omkreddy:
@Yes, we need UI change. I was thinking to raise new issue after merging this PR.

shahsank3t:
@omkreddy Ok cool.

omkreddy:
@shahsank3t  I will create new issue for UI change. We can  merge after this PR.

shahsank3t:
sounds good @omkreddy

HeartSaVioR:
Will this patch break Druid sink working with UI? If then let's file an issue for UI and handle both of them at once for not making master branch in broken state.

omkreddy:
changes are backward compatible.  We have raised  UI issue.
https://github.com/hortonworks/streamline/issues/1046

arunmahadevan:
Hi @priyank5485 , can you review the changes ?

bootstrap/components/sinks/v002__druid-sink-topology-component.json
```diff
@@ -0,0 +1,352 @@
...
+      {
+        "uiName": "Block On Full",
+        "fieldName": "blockOnFull",
+        "isOptional": true,
+        "tooltip": "Whether send will block (true) or throw an exception (false) when called while the outgoing queue is full",
+        "type": "boolean",
+        "defaultValue": true
+      },
+      {
+        "uiName": "Discarded Events Stream Id",
```
priyank5485: Does this eventually translate to driud bolt's output stream? If so, not sure it is really being used given that we dont allow any connections out of a sink in SAM. Also, the edges and connections between components are created by SAM. How does SAM leverage this user input value to name the output stream in flux yaml?

bootstrap/components/sinks/v002__druid-sink-topology-component.json
```diff
@@ -0,0 +1,352 @@
...
+                  "uiName":"fieldName",
+                  "fieldName": "fieldName",
+                  "tooltip": "fieldName",
+                  "isOptional": false,
+                  "type": "enumstring",
+                  "options": [],
+                  "hint": "inputFields"
+                },
+               {
+                 "uiName": "Parallelism",
```
priyank5485: This does not seem to be at the right level. Can you double check if it translates to the correct number of executors in the underlying storm topology? Same for v001 json.

streams/runners/storm/layout/src/main/java/com/hortonworks/streamline/streams/layout/storm/DruidEventMapper.java
```diff
@@ -37,7 +39,60 @@ public DruidEventMapper(String eventFiledName) {
...
+        }
+
+        @Override
+        protected Map<String, Object> delegate() {
+            return event;
+        }
+
+        @Override
+        public Object get(@Nullable Object key) {
+            Object value = super.get(key);
```
priyank5485: Can you see if you can replace code in this method using https://github.com/hortonworks/streamline/blob/master/streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/StreamlineRuntimeUtil.java#L24
That way we can have everything in one place.

priyank5485:
@omkreddy Reviewed and left some comments. I believe that main change in json is for dimensions where noNestedField is removed and hence UI will now also start passing field names with a dot convention for nesting. 

On a side note, I noticed that some druid related classes are probably in the wrong module. For example DruidBeamFactoryImpl and DruidEventMapper is in streamline-layout-storm. If thats a class thats only being used in the bolt then you can move it to the storm-runtime-storm. Thats the convention I have been following but it might not apply here.

bootstrap/components/sinks/v002__druid-sink-topology-component.json
omkreddy: Agree. Currently we are just exposing the underlying druid bolt property. This is not used by SAM.  If required will remove this from druid sink.  will do in another PR.

bootstrap/components/sinks/v002__druid-sink-topology-component.json
```diff
@@ -0,0 +1,352 @@
...
+                  "type": "enumstring",
+                  "options": [],
+                  "hint": "inputFields"
+                }
+             ]
+           }
+        ]
+      },
+      {
+        "uiName": "Parallelism",
```
omkreddy: Good catch.  moved the Parallelism field to appropriate place.

streams/runners/storm/layout/src/main/java/com/hortonworks/streamline/streams/layout/storm/DruidEventMapper.java
```diff
@@ -37,7 +40,31 @@ public DruidEventMapper(String eventFiledName) {
...
+
+        @Override
+        protected Map<String, Object> delegate() {
+            return event;
+        }
+
+        @Override
+        public Object get(@Nullable Object key) {
+            return StreamlineEventUtils.getFieldValue(event, (String)key);
+        }
```
omkreddy: Thanks for the suggestion.  Actually I was looking for the existing util method.  

I moved the  StreamlineRuntimeUtils to common module and renamed as StreamlineUtils. let me know if any concerns.

omkreddy:
@priyank5485  Thanks for the review. Updated few review comments.

Regarding moving classes to storm-runtime-storm: Agree looks like druid and opentsdb related classes are in streamline-layout-storm. Will do the cleanup after the release, to avoid any regression issues.

Since i was not moving the classes,  to access StreamlineRuntimeUtils, I moved the StreamlineRuntimeUtils to common module and renamed as StreamlineUtils. let me know if any concerns.

priyank5485:
@omkreddy Can you file a follow up issue to clean up the changes we discussed in this PR ?Regarding renaming and moving StreamlineRuntimeUtils is it done because druid related runtime classes are in layout and it does not depend on runtime? If so, I think we should move StreamlineUtil back to runtime module since its only accessed in runtime classes. Can you make a note of that in the follow up issue? 

As far as DruidEvent is concerned do we still need that inner class? It seems it was created just to get the nested fields. If so, do you think we should remove that class completely and just use the StreamlineUtils method wherever needed? If so, you can do it in this PR or handle in the follow up issue. I will leave it up to you.

Other than the two things mentioned in this PR I am +1 once you raise a follow up issue.

omkreddy:
@priyank5485  raised  issue here: #1070

Druid client library takes Map as input data event and key list (dimensions) to query the event. So we need pass the DruidEvent to the underlying druid library.

arunmahadevan:
+1

arunmahadevan:
Thanks @omkreddy for the patch. Merged to master.

PR #1053, opened 2017-12-01
ISSUE-1052: Update kafka source bundle json during migration

priyank5485:
Update kafka source bundle json during migration

### DIFF
--- a/bootstrap/components/sources/v002__kafka-source-topology-component.json
+++ b/bootstrap/components/sources/v002__kafka-source-topology-component.json
@@ -1,0 +1,332 @@
+{
+  "type": "SOURCE",
+  "name": "Kafka",
+  "subType": "KAFKA",
+  "streamingEngine": "STORM",
+  "builtin": true,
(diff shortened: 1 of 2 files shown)


priyank5485:
@raju-saravanan @satishd Can you please review this? Idea is that during migration from any older version to current or higher version we need a PUT on kafka source bundle json. Was not sure about naming of the file since it has v001 prefix. Should it be v002 as per convention? In my case, the two files are identical.

satishd:
@priyank5485 Right, you need to use PUT  as that resource already exists and it needs to be updated with latest config. 
We are following v00x convention of having respective version(x) as it will be easy to read/maintain. 
IMHO, better to follow that convention in this case also instead of having a comment in the script on why v001 is used.

priyank5485:
@satishd @raju-saravanan I have copied v001 to v002 for kafka source and used that file in PUT now as per Satish's suggestion. Squashed the commit since it does not make any worse.

raju-saravanan:
+1 LGTM, thanks @priyank5485 for taking care of this.

PR #1064, opened 2017-12-06
ISSUE-984: Change text mysql type to mediumtext for large schema

priyank5485:
Change text mysql type to mediumtext for large schema

### DIFF
--- a/bootstrap/sql/mysql/v010__modify_text_to_mediumtext.sql
+++ b/bootstrap/sql/mysql/v010__modify_text_to_mediumtext.sql
@@ -22,6 +22,15 @@ ALTER TABLE `topology_source` MODIFY `configData` MEDIUMTEXT NOT NULL;
 ALTER TABLE `topology_sink` MODIFY `configData` MEDIUMTEXT NOT NULL;
 ALTER TABLE `topology_processor` MODIFY `configData` MEDIUMTEXT NOT NULL;
 ALTER TABLE `topology_edge` MODIFY `streamGroupingsData` MEDIUMTEXT NOT NULL;
+ALTER TABLE `service_configuration` MODIFY `configuration` MEDIUMTEXT NOT NULL;
+ALTER TABLE `topology_editor_toolbar` MODIFY `data` MEDIUMTEXT NOT NULL;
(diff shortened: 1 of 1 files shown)


priyank5485:
@arunmahadevan I have changed a few other columns that i think might store schemas as well and have changed them. Please review when you get a chance.

arunmahadevan:
How about https://github.com/hortonworks/streamline/blob/master/bootstrap/sql/mysql/v001__create_tables.sql#L34 ?

Can you also test the migration apps created with old schema to new schema and see if it works ?

If the change from TEXT -> MEDIUMTEXT is backward compatible, lets change all the TEXT to MEDIUMTEXT.

priyank5485:
@arunmahadevan  I had considered pmml field but then decided against it. I will update the PR and add it as well.

I did not get the part about migration apps created from old schema to new schema. can you please elaborate? If you are talking about flyway migration scripts then i have already tested it. I first ran bootstrap-storage with migrate option but without the v010__ that has the alter statement. I then inserted some records in tables. And then I ran migrate again with v010__. After that i checked mysql tables to make sure the column types were correct for all. I also used the browser then to send a GET request to make sure that old data is being fetched and correct json is returned.

Regarding changing all TEXT to MEDIUMTEXT, the change is backwards compatible. However, I considered against it because most of the other places we use TEXT are for things like name or description. So it seemed an overkill. Granted that it takes only one more byte to save the same string using MEDIUMTEXT than TEXT if the string could fit in TEXT field so there is no increase in disk space usage. However, i read at a few places online that the libraries connecting to mysql server(jdbc driver in our case) allocate more memory for querying MEDIUMTEXT field since they dont know beforehand how long the data is going to be. That can overburden the heap and create some performance issues. Do we still want to change all fields from TEXT to MEDUIMTEXT? I prefer not to.

arunmahadevan:
@priyank5485 , as long as you have created some SAM app in the old schema and migrated and able to access the same it should be fine.

- "name" can be varchar
- "description" can be TEXT
- for anything else if we cannot anticipate the max value then going with MEDIUMTEXT would be better IMO. Can you review all the TEXT fields again?

arunmahadevan:
@priyank5485 , can you also test if you are able to deploy topologies with a large schema?

Issue #1069, opened 2017-12-07
Test mode throws IllegalStateException


arunmahadevan:
Topology details : http://hdf-3-1-build3.field.hortonworks.com:7777/#/applications/4/edit?_k=membfo

It throws an exception when trying to view any of the events by clicking on the bottom panel.

The stacktrace is below:

```
ERROR  [20:49:31.883] [dw-62453 - GET /api/v1/catalog/topologies/4/testhistories/55/events/correlated/bc9ba583-54a3-4f81-8cd2-b0511627a673] c.h.s.s.s.TopologyTestRunResource -  Got exception: [IllegalStateException] / message [Duplicate key EventInformation{timestamp=1512679550726, componentName='Split', streamId='projection_transform_stream_47', targetComponentName='ENRICH-WEATHER', eventId='563faea7-3335-4c9c-bca8-3c29ada88aba', rootIds=[bc9ba583-54a3-4f81-8cd2-b0511627a673, 037a8e52-62b8-4c2b-b82e-d10f3006fe40], parentIds=[52fa7470-ae64-4537-9dc7-3b03d0046c47], fieldsAndValues={splitJoinValue=1506437672640, week=39, eventTime=2017-09-26 14:54:32.64, eventSource=truck_geo_event, truckId=54, driverId=23, driverName=Jeff Markham, routeId=1, route=Peoria to Ceder Rapids Route 2, eventType=Normal, latitude=40.7, longitude=-89.52, correlationId=1, geoAddress=, speed=58}}] / related resource location: [com.hortonworks.streamline.streams.service.TopologyTestRunResource.getGroupedCorrelatedEventsOfTestRunTopologyHistory](TopologyTestRunResource.java:261)
[...]

PR #1072, opened 2017-12-09
ISSUE-1069 Handles events fanout in event logging on test mode

HeartSaVioR:
* support multiple target components in an event
* fix the logic based on above fact

Please refer comments on #1069 for more details.

### DIFF
--- a/streams/common/src/main/java/com/hortonworks/streamline/streams/common/event/EventInformation.java
+++ b/streams/common/src/main/java/com/hortonworks/streamline/streams/common/event/EventInformation.java
@@ -15,6 +15,7 @@
  **/
 package com.hortonworks.streamline.streams.common.event;
 
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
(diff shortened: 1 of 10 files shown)


HeartSaVioR:
@arunmahadevan
I covered the case via added/modified unit tests, but can't reproduce exactly same environment from #1069 cause it relies on non-builtin things so can't even import the topology.
Could you do some manual tests if you could handle the test cluster?

@shahsank3t
Could you check that UI needs fix for targetComponent -> targetComponents change? I guess we don't leverage it in UI page, but it may need to be fixed so would like to double-check.

HeartSaVioR:
Moved mentions to comment to avoid including them as commit message.

HeartSaVioR:
Just created simple example and tested manually.

screen shot 2017-12-10 at 8 36 14 pm
screen shot 2017-12-10 at 8 36 21 pm
screen shot 2017-12-10 at 8 36 26 pm

Issue #1071, opened 2017-12-09
Disable the logsearch button in UI if Logsearch and Ambari Infra is not available in the Environment


arunmahadevan:
We need to hide the logsearch button in view mode if the SAM Env does not have Ambari infra.

We cannot guarantee that the cluster will have Ambari Infra and logsearch especially when the user has already installed SAM and upgrades to HDF 3.1.

arunmahadevan:
@shahsank3t @HeartSaVioR , do we need any backend APIs for this or the UI can figure it out ?

shahsank3t:
Basically, if the user selects Ambari Infra while creating an environment and uses the same one to create an application, we should show LogSearch button else not. 
I'll make the required changes.

@arunmahadevan No backend API's or changes should be needed for this change.

shahsank3t:
Raised PR: https://github.com/hortonworks/streamline/pull/1073

arunmahadevan:
Merged to master

PR #1073, opened 2017-12-11
ISSUE-1071 Disable the logsearch button in UI if Logsearch and Ambari Infra is not available in the Environment

shahsank3t:
Also fixed react-select issues from Log Search and Topology view mode metrics


shahsank3t:
This closes https://github.com/hortonworks/streamline/issues/1071

arunmahadevan:
+1

PR #1082, opened 2017-12-13
ISSUE-1081: Fix the validation errors when reading from kafka message values when length is more than offset.

### LINKED ISSUE #1081: Errors reading from Kafka when dealing with large messages
SAM Kafka source throws below exception at runtime.

`
Java.lang.IndexOutOfBoundsException: end index (3225) must not be less than start index (4967) at org.apache.storm.shade.com.google.common.base.Preconditions.checkPositionIndexes(Preconditions.java:386) at com.hortonworks.streamline.streams.runtime.storm.spout.AvroKafkaSpoutTranslator$ByteBufferInputStream.read(AvroKafkaSpoutTranslator.java:108) at 
`

arunmahadevan:
Fix the wrong index validations and added tests.

Fixes #1081

### DIFF
--- a/streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/spout/AvroKafkaSpoutTranslator.java
+++ b/streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/spout/AvroKafkaSpoutTranslator.java
@@ -105,7 +105,7 @@ public class AvroKafkaSpoutTranslator implements RecordTranslator<Object, ByteBu
 
         public int read (byte[] bytes, int off, int len) throws IOException {
             Preconditions.checkNotNull(bytes, "Given byte array can not be null");
-            Preconditions.checkPositionIndexes(off, len, bytes.length);
+            Preconditions.checkPositionIndexes(off, off + len, bytes.length);
 
             if (!buf.hasRemaining()) {
                 return -1;
--- a/streams/runners/storm/runtime/src/test/java/com/hortonworks/streamline/streams/runtime/storm/spout/AvroKafkaSpoutTranslatorTest.java
+++ b/streams/runners/storm/runtime/src/test/java/com/hortonworks/streamline/streams/runtime/storm/spout/AvroKafkaSpoutTranslatorTest.java
@@ -1,0 +1,63 @@
+/**
+ * Copyright 2017 Hortonworks.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+
+ *   http://www.apache.org/licenses/LICENSE-2.0
+
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
+package com.hortonworks.streamline.streams.runtime.storm.spout;
+
+import org.junit.Test;
+
+import java.io.IOException;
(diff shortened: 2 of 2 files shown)


streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/spout/AvroKafkaSpoutTranslator.java
```diff
@@ -104,17 +104,18 @@ public int read () throws IOException {
         }
 
         public int read (byte[] bytes, int off, int len) throws IOException {
-            Preconditions.checkNotNull(bytes, "Given byte array can not be null");
-            Preconditions.checkPositionIndexes(off, len, bytes.length);
+            if (bytes == null) {
+                throw new NullPointerException();
+            } else if (off < 0 || len < 0 || len > bytes.length - off) {
+                throw new IndexOutOfBoundsException();
+            } else if (len == 0) {
```
satishd: It is good to have this condition to be checked only when ByteBuffer has any bytes to be read. So, It is better to leave where it is currently.

It seems newly added conditional checks follow what `InputStream#(byte b[], int off, int len)` does but the above suggestions are valid for this scenario.

streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/spout/AvroKafkaSpoutTranslator.java
```diff
@@ -104,17 +104,18 @@ public int read () throws IOException {
         }
 
         public int read (byte[] bytes, int off, int len) throws IOException {
-            Preconditions.checkNotNull(bytes, "Given byte array can not be null");
-            Preconditions.checkPositionIndexes(off, len, bytes.length);
+            if (bytes == null) {
+                throw new NullPointerException();
```
satishd: nit: Good to have error message in the Exception or leave the current null check as earlier which takes care of that.
```
Preconditions.checkNotNull(bytes, "Given byte array can not be null");
```

streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/spout/AvroKafkaSpoutTranslator.java
```diff
@@ -104,17 +104,18 @@ public int read () throws IOException {
...
+            } else if (off < 0 || len < 0 || len > bytes.length - off) {
+                throw new IndexOutOfBoundsException();
```
satishd: nit: good to have respective error message.

streams/runners/storm/runtime/src/test/java/com/hortonworks/streamline/streams/runtime/storm/spout/AvroKafkaSpoutTranslatorTest.java
```diff
@@ -0,0 +1,70 @@
...
+        ByteBuffer byteBuffer = ByteBuffer.allocate(1000);
+        AvroKafkaSpoutTranslator.ByteBufferInputStream is = new AvroKafkaSpoutTranslator.ByteBufferInputStream(byteBuffer);
+        byte[] dest = new byte[1000];
+        int res = is.read(dest, 100, 900);
+        assertEquals(900, res);
+    }
+
+    @Test
+    public void testByteBufferInputStreamNonZeroOffset2() throws Exception {
+        ByteBuffer byteBuffer = ByteBuffer.allocate(1000);
```
satishd: A method can be extracted which takes offset and len, which can be used in the two tests above.

arunmahadevan:
@satishd , thanks for the review. Addressed the comments.

satishd:
@arunmahadevan Thanks for addressing review comments, merged PR.

PR #1094, opened 2017-12-14
ISSUE-1093: Flatten the nested fields in DruidEvents

omkreddy:


omkreddy:
unfortunately current approaching is not working for nested fields.  Druid client library is converting the input Map object to json string.  we need to the flatten the map

@priyank5485  StreamlineRuntimeUtil is not used in DruidEvent now. I moved StreamlineRuntimeUtil back to streamline-runtime module. sorry for the confusion.

@arunmahadevan verified the patch with druid cluster.

streams/runners/storm/layout/src/main/java/com/hortonworks/streamline/streams/layout/storm/DruidEventMapper.java
```diff
@@ -55,20 +57,27 @@ public DruidEvent(StreamlineEvent event) {
             this.event = event;
         }
 
-        public static boolean isMap(Object map) {
-            return map instanceof Map;
-        }
-
         @Override
         protected Map<String, Object> delegate() {
```
arunmahadevan: the `delegate` is going to be invoked multiple times, once per map operation, so maybe you can save the flattened result once and re-use it.

streams/runners/storm/layout/src/main/java/com/hortonworks/streamline/streams/layout/storm/DruidEventMapper.java
```diff
@@ -55,20 +57,27 @@ public DruidEvent(StreamlineEvent event) {
...
-            return event;
+            Map<String, Object> updatedEvent = new LinkedHashMap<>();
+            event.forEach((key, value) -> {
+                flattenNestedFields(key, value, updatedEvent, new LinkedList<>(Arrays.asList(key)));
```
arunmahadevan: its better if we skip this method for non nested values to avoid unnecessary overhead. For non-nested, I assume you can just insert the value into the updatedEvent.

If you make the `parentKeyList` a plain String it doesnt have any overhead, so you can leave it as is.

streams/runners/storm/layout/src/main/java/com/hortonworks/streamline/streams/layout/storm/DruidEventMapper.java
```diff
@@ -55,20 +57,27 @@ public DruidEvent(StreamlineEvent event) {
...
-        @Override
-        public Object get(@Nullable Object key) {
-            Object value = StreamlineEventUtils.getFieldValue(event, (String)key);
-            LOG.debug("called for key: {}, value: {}", key, value);
-            return value;
+        private void flattenNestedFields(String key, Object value, Map<String, Object> updatedEvent, List<String> parentKeyList) {
+            if (value instanceof  Map) {
+                Map<String, Object> subMap = (Map<String, Object>) value;
+                subMap.forEach((childKey, childValue) -> {
+                    List<String> updatedParentKeyList = new LinkedList<>(parentKeyList);
```
arunmahadevan: you could make the `parentKey` a string and then directly append the childKey and pass it around to avoid creating the new list.

e.g. 

```java
subMap.forEach((childKey, childValue) -> {
  flattenNestedFields(childKey, childValue, updatedEvent, parentKey + "." + childKey);
});
```

streams/runners/storm/layout/src/main/java/com/hortonworks/streamline/streams/layout/storm/DruidEventMapper.java
arunmahadevan: first arg (key) is not required, or make be just make it the `parentKey`

omkreddy:
@arunmahadevan  Thanks for the review. addressed review comments. Pls take a look.

arunmahadevan:
+1

PR #1101, opened 2017-12-19
ISSUE-1097 UI changes to display and search sampled events

sameer79:
1] Implementation of Sample in view mode for an individual component.
2] A dedicated page for Sample, shows all component events.
3] Filtering events on multiple components.
4] Search applied to keyValues, headers, auxKeyValues using "searchString"
5] Search  applied to eventId, rootIds, parentIds using "searchEventId"
6] **This PR depends on** https://github.com/hortonworks/streamline/pull/1100

@arunmahadevan @HeartSaVioR @shahsank3t, please review and test using the depend PR, open for any suggestion.
 Thanks in advanced

PR #1106, opened 2017-12-21
ISSUE-1105 Fix log level status API to respond status 200 when topology is not running

HeartSaVioR:
* change response format to be similar with sampling status API

Response format of `/logconfig` (both GET/POST) will contain:
- success (always presented)
- enabled (only when success = true)
- logLevel (only when enabled = true)
- epoch (only when enabled = true)

This format is similar to sampling status API.

### DIFF
--- a/streams/actions/src/main/java/com/hortonworks/streamline/streams/actions/TopologyActions.java
+++ b/streams/actions/src/main/java/com/hortonworks/streamline/streams/actions/TopologyActions.java
@@ -107,18 +107,32 @@ public interface TopologyActions {
 
     class LogLevelInformation {
         private LogLevel logLevel;
-        private long epoch;
+        private Long epoch;
+        private boolean enabled;
 
-        public LogLevelInformation(LogLevel logLevel, long epoch) {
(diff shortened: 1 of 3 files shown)


HeartSaVioR:
@arunmahadevan @shahsank3t @sameer79
Please take a look. Given that this patch modifies the API endpoint paths which are related to #1101, this patch is effectively blocker for #1101.

@sameer79 I'm sorry, but I think this patch should be also applied to #1101. Please take a look at the API endpoint paths.

HeartSaVioR:
@shahsank3t 
> componentName as null

That's valid message: component is determined by thread name, and there're many background threads which also writes valuable log messages. Sometimes it's more helpful than executor thread.

@arunmahadevan 
If we want to hide log messages from non-executor threads plesae let me know.

arunmahadevan:
@HeartSaVioR , lets keep logs from all threads. The user can filter based on component name to see executor thread logs.

arunmahadevan:
+1

HeartSaVioR:
@arunmahadevan Rebased. Once I have +1 I'll just merge to avoid rebasing again.

PR #1110, opened 2017-12-22
ISSUE-1109 Provide execute latency into in processor and sink for time-series manner

### LINKED ISSUE #1109: Provide execute latency into in processor and sink for time-series manner
Quoting the issue content in https://github.com/hortonworks/streamline/issues/1108

> 3. Execute latency in the processors and sinks components. Right now it shows process latency which includes the time to ACK so does not capture the actual time it spends in the execute. (I am referring to execute and process latency based on the Storm).
>
> Execute latency: The average time a Tuple spends in the execute method. The execute method may complete without sending an Ack for the tuple.
>
> Process latency: The average time it takes to Ack a Tuple after it is first received. Bolts that join, aggregate or batch may not Ack a tuple until a number of other Tuples have been received.

We already provide "process latency" as "process time", hence we need to add "execute latency" as "execute time".

HeartSaVioR:
* add 'executeTime' into "misc" for processors and sinks
  * it will not be available in sources

This closes #1109

### DIFF
--- a/streams/runners/storm/metrics/src/main/java/com/hortonworks/streamline/streams/metrics/storm/topology/StormMappedMetric.java
+++ b/streams/runners/storm/metrics/src/main/java/com/hortonworks/streamline/streams/metrics/storm/topology/StormMappedMetric.java
@@ -20,7 +20,7 @@ import com.hortonworks.streamline.streams.metrics.TimeSeriesQuerier;
 /**
  * Metric name conversion table between Streamline and Storm. It also contains function information for aggregation.
  */
-enum StormMappedMetric {
+public enum StormMappedMetric {
     completeLatency("__complete-latency", TimeSeriesQuerier.AggregateFunction.AVG),
     inputRecords("__execute-count", TimeSeriesQuerier.AggregateFunction.SUM),
     outputRecords("__emit-count", TimeSeriesQuerier.AggregateFunction.SUM),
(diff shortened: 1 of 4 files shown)


HeartSaVioR:
The change is tiny. Most of changes are in test.

HeartSaVioR:
@arunmahadevan 

1. We don't show that thing in overviewMetrics and "process time" is represented as "latency" so if we want to include it as well, we should rename latency to back to "process time" as well so that we can have both and no one represent the latency. UI change is needed as well.

2. `execute latency` is out of whilelist in metrics consumer. I was requested to reduce metrics as much as possible so I excluded all of unused things for that time. I'm adding it in test cluster, but it means we are required to fix it from Ambari side as well if we want to get it by default. 
Could you file an issue and assign as well?

arunmahadevan:
>We don't show that thing in overviewMetrics and "process time" is represented as "latency" so if we want to include it as well, we should rename latency to back to "process time" as well so that we can have both and no one represent the latency. UI change is needed as well.

In storm its named "Process latency" and "Execute latency". I think we can follow this? Yes we need to rename in UI as well.

>I found execute latency is out of whilelist in metrics consumer. I was requested to reduce metrics as much as possible so I excluded all of unused things for that time. I'm adding it in test cluster, but it means we are required to fix it from Ambari side as well if we want to get it by default.
Could you file an issue and assign as well?

https://github.com/hortonworks/streamline/issues/1112

HeartSaVioR:
@arunmahadevan 
We need to restart topologies in test cluster to take effect. Oddly I can't open the Streamline (it shows blank page) in test server. Really appreciate if you could handle it.

And I think there's nothing to do in SAM side for handling whitelist. Please file an issue to Ambari side.

arunmahadevan:
@HeartSaVioR , raised Ambari issue - https://issues.apache.org/jira/browse/AMBARI-22696

You meant to restart topology to see executeTime or for something else ? I have restarted the topology.

HeartSaVioR:
@arunmahadevan Execute time should be shown in test cluster after restarting topology, cause I modified configuration in test cluster manually.

HeartSaVioR:
@arunmahadevan @shahsank3t 
Updated PR to add latency in overall metrics.

  * topology: latency / prevLatency
  * source: completeLatency / prevCompleteLatency
  * processor/sink: executeTime / prevExecuteTime, processTime / prevProcessTime 

Note that I just used *Time instead of *Latency since @shahsank3t would like to handle the rename from UI side.

arunmahadevan:
Thanks @HeartSaVioR , merged to master. Need to check the status of the Ambari patch to whitelist metrics.

PR #1096, opened 2017-12-18
ISSUE-1090 Provide backend API endpoint for fetching events indexed in Log Search service

HeartSaVioR:
* introduce new API endpoint: /topologies/:topologyId/events
  * mandatory params
    * topologyId
    * from
    * to
  * optional params
    * componentName
    * searchString
    * searchEventId
    * start
    * limit
    * ascending (true/false)
  * please refer to parameters in "/topologies/:topologyId/logs" (#1002)

`searchString` is applied to keyValues, headers, auxKeyValues (OR query).
`searchEventId` is applied to eventId, rootIds, parentIds (OR query).

This closes #1096

PR #1111, opened 2017-12-22
ISSUE-1107: UI issues in view mode

sameer79:
@arunmahadevan @HeartSaVioR @shahsank3t please review and comment.

### DIFF
--- a/webservice/src/main/resources/app/scripts/components/TopologyGraphComponent.jsx
+++ b/webservice/src/main/resources/app/scripts/components/TopologyGraphComponent.jsx
@@ -72,6 +72,7 @@ export default class TopologyGraphComponent extends Component {
     this.renderFlag = false;
     this.testRunActivated = props.testRunActivated;
     this.hideEventLog = props.hideEventLog;
+    this.componentLevelActionObj={};
   }
 
   componentWillReceiveProps(nextProps,previousProps){
@@ -513,6 +514,7 @@ export default class TopologyGraphComponent extends Component {
     } else {
       this.props.compSelectCallback('', null);
       this.viewModeNodeClickFlag = false;
+      this.triggerComponentLevelCallBack();
     }
   }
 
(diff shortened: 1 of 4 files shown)


arunmahadevan:
@sameer79 , the reported issues have been addressed. A couple more that will be good to fix.

1. In the Sampling screen selected "Event Id" and "Root Id". Able the view the values. However after I go back to view mode and come back to the sampling view, these settings are not saved. Same behavior while selecting "host" in the log search. Can we save these settings?

2. The samples are always displayed in the ascending order so to view the latest event, one needs to select "last 5 mins" and then navigate till the last event. I will be good to add an option to reverse sort. @HeartSaVioR does solr support this ? If so we can provide an extra "sort" order in the API and UI can use it.

HeartSaVioR:
@arunmahadevan @sameer79 
Please refer the parameters in PR's body: https://github.com/hortonworks/streamline/pull/1096
For events we have sort option, for log we don't have the sort option, but I can add it if we want.

arunmahadevan:
@HeartSaVioR , thanks for the info. It will be good to add for logs as well. but we can track that as a separate issue.

@sameer79 , so can you address the below ?
1. In the Sampling screen selected "Event Id" and "Root Id". Able the view the values. However after I go back to view mode and come back to the sampling view, these settings are not saved. Same behavior while selecting "host" in the log search. Can we save these settings?

2. Add a sort option in the UI to view the result in ascending or descending order.

shahsank3t:
@arunmahadevan @HeartSaVioR 
I'll be addressing these UI changes since @sameer79 is on vacation. I'll create another PR with this PR commit and my changes above that in short time and close this one.

shahsank3t:
Raised PR https://github.com/hortonworks/streamline/pull/1114 with @sameer79's changes and addressed the comments above. Closing this now.

Issue #1107, opened 2017-12-21
UI issues in view mode


arunmahadevan:
1. After entering sampling pct at topology level the message displayed is "Sampling enable successfully". We can change it to "Sampling enabled successfully" or just "Sampling enabled". Similarly for component level it could be "Component sampling enabled". Similar text changes for disabling.

2. For topology level, the sampling setting takes effect once the mouse is clicked outside the settings box. However for component level I need to hit return key. We could have similar behavior for both settings.

3. While changing the log level also the "Sampling enable successfully" message is displayed.

HeartSaVioR:
3. in case of confusion, please change the message accordingly like "Changing log level successfully"

sameer79:
Raised  PR : https://github.com/hortonworks/streamline/pull/1111

shahsank3t:
#1111 is closed.
Raised another PR to address this issue: https://github.com/hortonworks/streamline/pull/1114

arunmahadevan:
Merged #1114 to master

PR #1113, opened 2017-12-23
ISSUE-1108 View mode metrics changes

shahsank3t:
1. Removed "Process Time" and renamed "Latency" to "Overall Latency" in the bottom panel for the source.
2. Added "Execute Time" graph for processors and sinks in the bottom panel metrics.

### DIFF
--- a/webservice/src/main/resources/app/scripts/components/TopologyGraphComponent.jsx
+++ b/webservice/src/main/resources/app/scripts/components/TopologyGraphComponent.jsx
@@ -149,11 +149,12 @@ export default class TopologyGraphComponent extends Component {
     TAB_KEY: 9,
     ENTER_KEY: 13,
     rectangleWidth: 184,
+    newMetricsRectangleWidth: 225,
     rectangleHeight: 40,
     testDataRectHeight: 210,
     testNoDataRectHeight: 75,
-    metricsRectHeight: 100,
-    metricsGraphRectHeight: 235
+    metricsRectHeight: 112,
+    metricsGraphRectHeight: 247
   };
 
   componentDidUpdate() {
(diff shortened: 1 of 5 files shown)


arunmahadevan:
@shahsank3t the changes looks good. We just need to finalize the labels. I think we can stick to the names in the Storm UI for the sources.

1. In the source (DAG) - Complete latency
2. In the source (bottom panel) - Complete latency
3. In the processors/sinks (DAG) - Process time , Execute time
4. In the processors/sinks (bottom panel) - Process time , Execute time

@HeartSaVioR let me know what you think. For 3, 4 do you think storm's naming ("process latency" and "execute latency") is better ?

@shahsank3t Can you rename the labels for 1 and 2 ? I think you are waiting for backend changes for 3 ? If so lets wait and address 3  in the patch before merging it. 

I have this patch (plus changes in #1107) deployed here - http://ctr-e136-1513029738776-15460-01-000002:7777/

shahsank3t:
@arunmahadevan When showing topology level metrics count (in the bottom panel), for `latency`, should we keep it as is or will it be `Complete Latency` or `Process Time`?

arunmahadevan:
@shahsank3t , I think we need to explicitly mention "Complete latency" only for source. We can show it in the source, bottom bar (with up/down arrows) and also when the bottom panel is expanded (for sources).

For topology level, its kind of implicit that this is a topology level value, so we can just show "Latency" when all components are selected, both in the bottom bar and in the expanded bottom panel.

HeartSaVioR:
@arunmahadevan If we want to make Storm users more familiar with our labels, I'd rather making labels in sync with Storm. If we still want to open a chance with having another streaming engines (though I don't think that's possible becase it will require dropping so many features) I'd rather not. 
I think either is fine even for another streaming engines.

arunmahadevan:
@shahsank3t , can you rename the labels to make it in sync with Storm ?

1. In the source (DAG) - Complete latency
2. In the source (bottom panel) - Complete latency
3. In the processors/sinks (DAG) - Process latency , Execute latency
4. In the processors/sinks (bottom panel) - Process latency, Execute latency
5. All Components - Latency

shahsank3t:
@arunmahadevan @HeartSaVioR 
Renamed the fields and updated the PR.

arunmahadevan:
+1

PR #1114, opened 2017-12-23
ISSUE-1107: UI issues in view mode

shahsank3t:
This PR contains changes from @sameer79's PR (https://github.com/hortonworks/streamline/pull/1111) and addressed following comments by @arunmahadevan https://github.com/hortonworks/streamline/pull/1111#issuecomment-353690604

> 1. In the Sampling screen selected "Event Id" and "Root Id". Able the view the values. However after I go back to view mode and come back to the sampling view, these settings are not saved. Same behavior while selecting "host" in the log search. Can we save these settings?
> 2. Add a sort option in the UI to view the result in ascending or descending order.

### DIFF
--- a/webservice/src/main/resources/app/scripts/components/TopologyGraphComponent.jsx
+++ b/webservice/src/main/resources/app/scripts/components/TopologyGraphComponent.jsx
@@ -72,6 +72,7 @@ export default class TopologyGraphComponent extends Component {
     this.renderFlag = false;
     this.testRunActivated = props.testRunActivated;
     this.hideEventLog = props.hideEventLog;
+    this.componentLevelActionObj={};
   }
 
   componentWillReceiveProps(nextProps,previousProps){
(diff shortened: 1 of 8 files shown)


shahsank3t:
@arunmahadevan 
We are now saving the state of column filters (except for Date/Time, Component Name and Key values column -- even if a user hides, it will still be visible on coming back to the page) and also saving the state of sorting.

Just to keep in mind, we are using browser's `LocalStorage` to save the state of column filter/sorting, so the state will be maintained till the data is present else fallback to default setting (which user sees when he loads the page for the first time)

arunmahadevan:
@shahsank3t , thanks for addressing the changes. Can you make similar changes in the log search UI as well (for host and port) ?

shahsank3t:
@arunmahadevan Addressed your comment to add the similar change to log search page.

arunmahadevan:
Thanks @shahsank3t , @sameer79 , merged the commits to master.

PR #1012, opened 2017-11-15
ISSUE-1015 Quote `window` when defining tables to avoid reserved word conflicts

navbruce:
Leaving `window` unquoted causes table creation to fail when using MariaDB v10.2

PR #1123, opened 2017-12-30
ISSUE-1015: Quote window in the mysql create statements to handle mariadb 10.2

arunmahadevan:
I am raising another PR just to address the quoting of "window" since there has been no further updates on https://github.com/hortonworks/streamline/pull/1012 and we are facing issues installing SAM in mariadb 10.2.


arunmahadevan:
Merged it since the QE deployment jobs were broken.

PR #1128, opened 2018-01-04
ISSUE-1127: SAM HBase Sink throws an exception when opening Sink Dialog

### LINKED ISSUE #1127: SAM HBase Sink throws an exception when opening Sink Dialog
NPE is thrown in SAM while Hbase sink tries to fetch hints: 

```
ERROR 15:04:43.193 dw-4473 - GET /api/v1/catalog/streams/componentbundles/SINK/11/hints/namespaces/2 c.h.s.s.s.TopologyComponentBundleResource - Got exception: RuntimeException / message java.lang.NullPointerException / related resource location: com.hortonworks.streamline.streams.service.TopologyComponentBundleResource.getFieldHints(TopologyComponentBundleResource.java:520) 
java.lang.RuntimeException: java.lang.NullPointerException
at com.hortonworks.streamline.streams.cluster.bundle.impl.HBaseBundleHintProvider.getHintsOnCluster(HBaseBundleHintProvider.java:45)
at com.hortonworks.streamline.streams.cluster.bundle.AbstractBundleHintProvider.provide(AbstractBundleHintProvider.java:54)
at com.hortonworks.streamline.streams.service.TopologyComponentBundleResource.getFieldHints(TopologyComponentBundleResource.java:520)
```

arunmahadevan:
ISSUE-1127: In non-secure mode, we should still pass the hbaseMaster and hbaseMasterProcesses while instantiating the HBaseMetadataService

closes #1127

### DIFF
--- a/streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/service/metadata/HBaseMetadataService.java
+++ b/streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/service/metadata/HBaseMetadataService.java
@@ -153,7 +153,7 @@ public class HBaseMetadataService implements AutoCloseable {
                     .getAdmin(), securityContext, subject, user, hbaseMaster, hbaseMasterProcesses);
         } else {
             return new HBaseMetadataService(ConnectionFactory.createConnection(hbaseConfig).getAdmin(),
-                    null, null, null, hbaseMaster, hbaseMasterProcesses);
+                    securityContext, subject, null, hbaseMaster, hbaseMasterProcesses);
         }
     }
 
--- a/streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/service/metadata/HBaseMetadataService.java
+++ b/streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/service/metadata/HBaseMetadataService.java
@@ -20,12 +20,12 @@ import com.google.common.collect.ImmutableList;
 import com.hortonworks.streamline.common.function.SupplierException;
 import com.hortonworks.streamline.streams.cluster.catalog.Component;
 import com.hortonworks.streamline.streams.cluster.catalog.ComponentProcess;
+import com.hortonworks.streamline.streams.cluster.discovery.ambari.ComponentPropertyPattern;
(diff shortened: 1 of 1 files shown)


streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/service/metadata/HBaseMetadataService.java
```diff
@@ -153,7 +152,8 @@ public static HBaseMetadataService newInstance(Configuration hbaseConfig,
             return new HBaseMetadataService(ConnectionFactory.createConnection(hbaseConfig, user)
                     .getAdmin(), securityContext, subject, user, hbaseMaster, hbaseMasterProcesses);
         } else {
-            return newInstance(hbaseConfig);
+            return new HBaseMetadataService(ConnectionFactory.createConnection(hbaseConfig).getAdmin(),
```
HeartSaVioR: Why not passing securityContext and subject given that it is not fetched from this method? Looks like only `user` is guard against null.

And I'd like to remove 1 arg constructor as well and remove all unused static newInstance() so that we don't make same mistake again. May also need to modify test code though.

HeartSaVioR: Why not passing securityContext and subject given that it is not fetched from this method? Looks like only user is guard against null.

And I'd like to remove 1 arg constructor as well and remove all unused static newInstance() so that we don't make same mistake again. May also need to modify test code though.

arunmahadevan:
@HeartSaVioR , added the subject and securityContext to the ctor. I did not refactor the newInstance since it was leading to more changes in the test code and wanted to avoid changing it now.

HeartSaVioR:
@arunmahadevan 
If you don't want to address it for now, then let's file the issue. I believe if someone who doesn't have context on this has a chance to modify the code here, they will likely do mistake again. Even us after 6 months.

arunmahadevan:
@HeartSaVioR , filed https://github.com/hortonworks/streamline/issues/1131

HeartSaVioR:
Thanks @arunmahadevan for the patch. I merged into master.

PR #1133, opened 2018-01-05
ISSUE-1126 Fix transaction lock issue while refreshing cluster

### LINKED ISSUE #1126: Shows issue while refreshing service pool cluster
refreshing cluster fails

HeartSaVioR:
* breaks down cluster import logic to two parts
  * fetch service and corresponding components information from Ambari (can be done concurrently)
    * including applying custom logic
  * store information to DB (must be done in same thread)
    * this is to group operations in "a" transaction
    * the transaction includes cleanup operation as well

This closes #1126

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -171,6 +171,7 @@
         <druid.tranquility.version>0.8.2</druid.tranquility.version>
         <spring.version>4.3.6.RELEASE</spring.version>
         <flyway.version>4.2.0</flyway.version>
+        <jool.version>0.9.12</jool.version>
 
         <!-- Runtime Scope Dependencies -->
         <logback.version>1.1.7</logback.version>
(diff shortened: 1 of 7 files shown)


arunmahadevan:
@HeartSaVioR , for instance during topology deployment where we run it in a separate thread, I am not sure why we keep on begin and commit [here](https://github.com/hortonworks/streamline/blob/master/streams/actions/src/main/java/com/hortonworks/streamline/streams/actions/topology/service/TopologyActionsService.java#L118) and [here](https://github.com/hortonworks/streamline/blob/master/streams/actions/src/main/java/com/hortonworks/streamline/streams/actions/topology/service/TopologyActionsService.java#L128). I suspect this may not even produce the expected result. In the case of failures we rollback the txn and the last saved state [here](https://github.com/hortonworks/streamline/blob/master/streams/actions/src/main/java/com/hortonworks/streamline/streams/actions/topology/state/TopologyStates.java#L82) is lost.

In the case of deployment failures, I think we need to 
1. wrap the entire deploy topology in a single txn.
2. Return the failure state back deploy
3. rollback (to clean up any partial data in the db)
4. start another txn, set the failure status and commit so that the state is saved.

I will raise a separate issue for this.

arunmahadevan:
@HeartSaVioR , thanks for the patch. For similar behavior in other modules lets raise separate issues as required.

PR #1134, opened 2018-01-05
ISSUE-1131 Remove the unused ctors and static newInstance methods fro…

### LINKED ISSUE #1131: Remove the unused ctors and static newInstance methods from HbaseMetadataService
Remove the 1 arg constructor as well and remove all unused static newInstance() so that we don't pass null values for the args and cause NPE.

HeartSaVioR:
…m HBaseMetadataService

* also removes some tests which are effectively unused and out of sync
 
This closes #1131


HeartSaVioR:
I've removed two tests and associated configurations because they're integration tests we don't run them for fairly long time. The host information in configuration is also invalid hence even without `@Ignore` it will not run.

arunmahadevan:
@HeartSaVioR , thanks for the patch, merged to master.

PR #1147, opened 2018-01-09
ISSUE-1146 Writing failure state while applying topology action should not be rolled back

### LINKED ISSUE #1146: Failure state should not be rolled back while deployment is failing
This is extracted out of #1137 to address easier issue first.

Quoting @arunmahadevan's description:

During topology deployment where we run it in a separate thread, we keep on doing `beginTransaction` and `commit` [here](https://github.com/hortonworks/streamline/blob/master/streams/actions/src/main/java/com/hortonworks/streamline/streams/actions/topology/service/TopologyActionsService.java#L118) and [here](https://github.com/hortonworks/streamline/blob/master/streams/actions/src/main/java/com/hortonworks/streamline/streams/actions/topology/service/TopologyActionsService.java#L128). I suspect this may not even produce the expected result. 

In the case of failures we rollback the txn and the last saved state [here](https://github.com/hortonworks/streamline/blob/master/streams/actions/src/main/java/com/hortonworks/streamline/streams/actions/topology/state/TopologyStates.java#L82) is lost.

This issue addresses failure case: failure state should be committed instead of rolled back.

HeartSaVioR:
* add new exception which is marker exception for ignoring rollback transaction
* apply to deploy/kill/suspect/resume

This closes #1146

### DIFF
--- a/streams/actions/src/main/java/com/hortonworks/streamline/streams/actions/topology/service/TopologyActionsService.java
+++ b/streams/actions/src/main/java/com/hortonworks/streamline/streams/actions/topology/service/TopologyActionsService.java
@@ -42,6 +42,7 @@ import com.hortonworks.streamline.streams.catalog.topology.TopologyComponentBund
 import com.hortonworks.streamline.streams.catalog.topology.component.TopologyDagBuilder;
 import com.hortonworks.streamline.streams.cluster.container.ContainingNamespaceAwareContainer;
 import com.hortonworks.streamline.streams.cluster.service.EnvironmentService;
+import com.hortonworks.streamline.common.exception.IgnoreTransactionRollbackException;
 import com.hortonworks.streamline.streams.layout.component.OutputComponent;
 import com.hortonworks.streamline.streams.layout.component.StreamlineProcessor;
 import com.hortonworks.streamline.streams.layout.component.StreamlineSource;
@@ -124,15 +125,10 @@ public class TopologyActionsService implements ContainingNamespaceAwareContainer
         }
         LOG.debug("Deploying topology {}", topology);
         while (ctx.getState() != TopologyStates.TOPOLOGY_STATE_DEPLOYED) {
-            try {
-                transactionManager.beginTransaction(TransactionIsolation.DEFAULT);
-                LOG.debug("Current state {}", ctx.getStateName());
-                ctx.deploy();
-                transactionManager.commitTransaction();
-            } catch (Exception e){
-                transactionManager.rollbackTransaction();
-                throw e;
-            }
+            mutateTopologyContext(topologyContext -> {
+                LOG.debug("Current state {}", topologyContext.getStateName());
+                topologyContext.deploy();
+            }, ctx);
         }
         return null;
     }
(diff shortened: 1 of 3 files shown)


HeartSaVioR:
Looks like recent SR change breaks the build. I'm pulling the latest of SR and rebuild, and trying to apply the change to SAM.

streams/actions/src/main/java/com/hortonworks/streamline/streams/actions/topology/service/TopologyActionsService.java
```diff
@@ -355,4 +375,12 @@ public StreamCatalogService getCatalogService() {
         return catalogService;
     }
 
+    private Exception convertThrowableToException(Throwable e) {
+        if (e.getCause() instanceof Exception) {
+            return (Exception) e.getCause();
+        } else {
+            return new RuntimeException(e);
```
arunmahadevan: 1. Shouldn't it be `new RuntimeException(e.getCause())` ?
2. Does this handle if e is `IgnoreTransactionRollbackException` and e.getCause() is null?

streams/actions/src/main/java/com/hortonworks/streamline/streams/actions/topology/service/TopologyActionsService.java
```diff
@@ -152,15 +148,39 @@ public boolean killTestRunTopology(Topology topology, TopologyTestRunHistory his
...
+    }
+
+    private void mutateTopologyContext(ConsumerWithThrowsException<TopologyContext> mutatingFunction,
+                                       TopologyContext topologyContext) throws Exception {
+        try {
+            transactionManager.beginTransaction(TransactionIsolation.DEFAULT);
+            mutatingFunction.accept(topologyContext);
+        } catch (IgnoreTransactionRollbackException e) {
+            transactionManager.commitTransaction();
+            throw convertThrowableToException(e.getCause());
```
arunmahadevan: you already invoke the `getCause` in `convertThrowableToException` so shouldnt this be convertThrowableToException(e) ?

streams/actions/src/main/java/com/hortonworks/streamline/streams/actions/topology/service/TopologyActionsService.java
```diff
@@ -355,4 +375,12 @@ public StreamCatalogService getCatalogService() {
         return catalogService;
     }
 
+    private Exception convertThrowableToException(Throwable e) {
+        if (e.getCause() instanceof Exception) {
+            return (Exception) e.getCause();
```
arunmahadevan: Findbugs warning.
 
```
[INFO] Unchecked/unconfirmed cast from Throwable to Exception of return value in 
com.hortonworks.streamline.streams.actions.topology.service.TopologyActionsService.convertThrowableToExce
ption(Throwable) [com.hortonworks.streamline.streams.actions.topology.service.TopologyActionsService] At 
TopologyActionsService.java:[line 380] BC_UNCONFIRMED_CAST_OF_RETURN_VALUE
```

You could assign the `cause` to a local variable and use that in the condition and cast to avoid this.

HeartSaVioR: Sadly I missed this comment and just fixed it with same approach instead of annotation. Thanks for guiding.

streams/actions/src/main/java/com/hortonworks/streamline/streams/actions/topology/service/TopologyActionsService.java
HeartSaVioR: 1. Yes I just fixed it.
2. No, but given that IgnoreTransactionRollbackException is just a marker exception it should provide valid exception. If we want to ensure this, null check should be landed to the constructor of IgnoreTransactionRollbackException.

streams/actions/src/main/java/com/hortonworks/streamline/streams/actions/topology/service/TopologyActionsService.java
HeartSaVioR: Yeah I realized `convertThrowableException` should handle passed exception instead of caused exception in passed exception, given that there's no context for the behavior. I'll leave this as is and fix `convertThrowableException` to handle passed exception.

HeartSaVioR:
@arunmahadevan Thanks for the review. I addressed your comments.

arunmahadevan:
+1

arunmahadevan:
thanks @HeartSaVioR for the fix, merged to master.

PR #1151, opened 2018-01-10
ISSUE-1150 Populate protocol/port for kafka components even listeners with multiple protocols.

satishd:
- Populate protocol/port for kafka components even listeners with multiple protocols.
- Fixed issue to set right protocol when kafka is kerberized though ambari gives wrong information about listener protocol.

### DIFF
--- a/streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/discovery/ambari/ComponentPropertyPattern.java
+++ b/streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/discovery/ambari/ComponentPropertyPattern.java
@@ -48,7 +48,8 @@ public enum ComponentPropertyPattern {
   // KAFKA
   // protocol (plaintext, ssl, kerberos, etc?), host (can be empty), port
   // https://cwiki.apache.org/confluence/display/KAFKA/Multiple+Listeners+for+Kafka+Brokers
-  KAFKA_BROKER("listeners", Pattern.compile("([a-zA-Z]+)://[a-zA-Z0-9_\\-\\\\.]*:([0-9]+)")),
+  // no matching pattern as listeners can contains multiple URIS with different schemes/protocols
+  KAFKA_BROKER("listeners", null),
 
   // AMBARI_METRICS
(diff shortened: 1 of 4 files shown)


streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/discovery/ambari/ComponentPropertyPattern.java
```diff
@@ -48,7 +48,8 @@
   // KAFKA
   // protocol (plaintext, ssl, kerberos, etc?), host (can be empty), port
   // https://cwiki.apache.org/confluence/display/KAFKA/Multiple+Listeners+for+Kafka+Brokers
-  KAFKA_BROKER("listeners", Pattern.compile("([a-zA-Z]+)://[a-zA-Z0-9_\\-\\\\.]*:([0-9]+)")),
+  // no matching pattern as listeners can contains multiple URIS with different schemes/protocols
+  KAFKA_BROKER("listeners", null),
```
HeartSaVioR: minor: Let's either commenting that it will be handled separately while importing cluster or just removing KAFKA_BROKER from here.

satishd: I started with removing that field but it is used in other places like [here](https://github.com/hortonworks/streamline/blob/master/streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/register/impl/KafkaServiceRegistrar.java#L38
) and [here](https://github.com/hortonworks/streamline/blob/master/streams/cluster/src/main/java/com/hortonworks/streamline/streams/cluster/service/metadata/KafkaMetadataService.java#L63). I did not want to remove that and comeup with a different enum or string from some other class as part of this PR.

HeartSaVioR: OK thanks for checking. Let's leave it as it is.

satishd:
@HeartSaVioR Thanks for reviewing.

PR #1158, opened 2018-01-12
ISSUE-1157 Fix leaking transactions in topology action

HeartSaVioR:
* move "commit" to finally statement and always commit unless it was rolled back


HeartSaVioR:
I deployed the change to test cluster and it resolved the issue.

Given that this is blocker issue, let me get this reviewed after post-merge.

PR #1161, opened 2018-01-13
ISSUE-1159: Remove user related toolbar metadata before deleting user

priyank5485:
Remove user related toolbar metadata before deleting user

### DIFF
--- a/streams/authorizer/pom.xml
+++ b/streams/authorizer/pom.xml
@@ -25,6 +25,10 @@
             <artifactId>streamline-common</artifactId>
         </dependency>
         <dependency>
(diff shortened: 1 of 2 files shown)


arunmahadevan:
@priyank5485 , can we have the UI invoke a request to delete the topology editor tool bar first before issuing the delete request for the user.

I am not sure if adding streamling-catalog dependency to streamline-authorizer will create other dependency issues. Also it will prevent the authorizer module from being re-used in other projects like registry.

priyank5485:
@arunmahadevan I am okay with asking UI to send a delete request for topology editor data. I will close this one then and assign the issue to Sanket.

PR #1163, opened 2018-01-16
ISSUE-1162: Trim stream id prefix from Hive Mapper fields/partitionFields

omkreddy:


omkreddy:
@arunmahadevan  Pls take a look.  looks  newer builds  UI is passing streamId along with filedNames. This was fixed in Druid before. We many need to check other sinks

satishd:
@arunmahadevan @omkreddy Found the same issue with HDFS sink, I am not sure whether it is intended for UI to generate field names with stream ids for sinks. @shahsank3t is going to confirm on that.

shahsank3t:
@satishd @omkreddy  @arunmahadevan 

Just to provide a context why this issue came up:
We support multiple input streams for sinks and they can have same field name. So to avoid the confusion, we use `streamID:fieldName` pattern everywhere we select fields in sinks.

As per my observation, following is the list of Sinks and their respective fields in which UI saves/sends the fields in above format (that backend might need to check if they support as well):
1. Druid => dimensions (supported)
2. HDFS => outputFields (unsupported, as per @satishd comment)
3. HIVE => fields & partitionFields (supported in this PR)
4. JDBC => columns (need to verify)

You guys might want to double check my observation and look into it.

streams/runners/storm/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/storm/bolt/StreamlineHiveMapper.java
```diff
@@ -29,8 +29,8 @@
     public StreamlineHiveMapper(List<String> fields, List<String> partitionFields) {
         Objects.requireNonNull(fields, "Empty fields");
         Objects.requireNonNull(partitionFields, "Empty partitionFields");
-        this.fields = new ArrayList<>(fields);
-        this.partitionFields = new ArrayList<>(partitionFields);
+        this.fields = fields.stream().map(field -> field.substring(field.indexOf(":") + 1, field.length())).collect(Collectors.toList());
```
satishd: No need to have endIndex  as field.length() as `field.substring(startIndex)` is sufficient. You may want to do the same change for the next line.

arunmahadevan: It may be better to do this in the FluxComponent. 

I dont think theres any logic to handle multiple input streams in this mapper, so how would we handle multiple input streams here ?

arunmahadevan:
@omkreddy , @shahsank3t , whats the expectation when there are multiple input streams? I dont think our sinks are capable of handling it yet. E.g. If the hbase sink is mapping the events to one table and the events in each stream are having different schema, how can it be mapped to a single table ? If the sink /mapper is processing the events without looking at the input stream of the tuple, we will end up with rows having null values for columns which are not present in each stream?

satishd:
@shahsank3t @arunmahadevan It is better to fix UI not sending `streamId:` prefix as field names instead of fixing all the sinks because none of the sinks support multiple streams. We may even want to restrict user not to create two streams pointing to a Sink.

arunmahadevan:
@satishd , agree. None of the sinks may be capable of handling multiple streams right now. 

@shahsank3t , can we add a UI restriction to limit sinks to one input stream and use "fieldName" instead of "stream:fieldName"? We can relook at enabling multiple input streams in future.

Not sure why the "stream:fieldName" was added. Was it required by any sink?

If the user wants to handle multiple streams, they can configure multiple instances of the same sink with appropriate fields.

shahsank3t:
@arunmahadevan @satishd 
Yes, we can restrict multiple streams being connected to sinks on the UI and remove `stream:fieldName` as well.
Just double checking, we should be doing this for all the sinks or specific ones?

satishd:
@shahsank3t AFAIK, we should do this for all sinks. We can enhance sinks to support multiple streams later.

shahsank3t:
@satishd @arunmahadevan 
Filed https://github.com/hortonworks/streamline/issues/1170 to address the above changes.

shahsank3t:
@satishd  @arunmahadevan Raised PR https://github.com/hortonworks/streamline/pull/1171 to address the changes on UI side.

PR #1175, opened 2018-01-18
ISSUE-1174 Apply ManagedTransaction to fix transaction issue

### LINKED ISSUE #1174: Apply ManagedTransaction to fix transaction issue
@arunmahadevan found another bug in #1158 and I found another corner cases (more about method contract) as well, so we just went forward to address them to Registry project.

ISSUE: https://github.com/hortonworks/registry/issues/381
PR: https://github.com/hortonworks/registry/pull/383

Since the PR got merged today, we can leverage ManageTransaction to avoid doing mistake again.

HeartSaVioR:
* leverage ManagedTransaction in Registry to remove custom implementation
  * custom implementation has a bug indeed

Change-Id: I4885f628d17e374023009b3af6f964d72002c874

This closes #1174


HeartSaVioR:
@arunmahadevan 
Current implementation has a bug which you found earlier while talking with me. In current implementation, if rollback throws exception, it will try to commit.

PR #1179, opened 2018-01-19
ISSUE-1178: Pass nimbus.seeds to storm command line

priyank5485:
Pass nimbus.seeds to storm command line instead of deprecated nimbus.host

### DIFF
--- a/streams/runners/storm/actions/src/main/java/com/hortonworks/streamline/streams/actions/storm/topology/StormTopologyActionsImpl.java
+++ b/streams/runners/storm/actions/src/main/java/com/hortonworks/streamline/streams/actions/storm/topology/StormTopologyActionsImpl.java
@@ -497,10 +497,9 @@ public class StormTopologyActionsImpl implements TopologyActions {
     private List<String> getNimbusConf() {
         List<String> args = new ArrayList<>();
 
-        // FIXME: Can't find how to pass list to nimbus.seeds for Storm CLI
-        // Maybe we need to fix Storm to parse string when expected parameter type is list
         args.add("-c");
-        args.add("nimbus.host=" + nimbusSeeds.split(",")[0]);
+        // See method documentation for getNimbusSeedsForCommandLine
+        args.add(NIMBUS_SEEDS + "=" + getNimbusSeedsForCommandLine());
 
         args.add("-c");
         args.add("nimbus.port=" + String.valueOf(nimbusPort));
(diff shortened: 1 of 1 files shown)


priyank5485:
@HeartSaVioR I figured we dont need a change in storm for that. Spent a significant time trying to figure out how to pass nimbus nimbus.seeds from java to storm.py. Please review. 
@arunmahadevan The original problem of SAM failing to submit an application if one nimbus host is down in HA mode is solved. Meaning with this PR i do see storm client getting a Connection Refused on first host that is done. I even see a topology submission request received on other nimbus in the nimbus.seeds list. However, the client times out on the second host for some reason. Trying to figure that issue out. Might not be related to this change at all. But need to confirm. Just putting it out for early review. Will update.

streams/runners/storm/actions/src/main/java/com/hortonworks/streamline/streams/actions/storm/topology/StormTopologyActionsImpl.java
```diff
@@ -941,4 +940,21 @@ private LogLevelInformation convertLogLevelResponseToLogLevelInformation(LogLeve
...
+    //note that you need double quotes in above json representation and single quotes are not valid
+    //Furthermore shell removes any double quotes from command line arguments when it passes to python.
+    //More on that at http://teaching.idallen.com/cst8207/12w/notes/280_quotes.txt
+    //As a result we need to pass something like [\"a\", \"b\"] to the command line. Without escaping the double quotes python will get arguments as [a,b] and
+    //that will fail the json parsing in storm submitter java code. Escaping of double quote for command line argument is taken care by java
+    private String getNimbusSeedsForCommandLine() {
+        StringBuilder nimbusSeedsJsonString = new StringBuilder("[");
+        if ((nimbusSeeds != null) && !nimbusSeeds.isEmpty()) {
+            String[] seeds = nimbusSeeds.split(",");
+            for (String seed: seeds) {
```
HeartSaVioR: This would be safer if we use stream() and Collectors.joining() or `String.join()`. 

Currently it adds "," by end of array, and assuming you tested before then it doesn't break the functionality, but still not beauty. :)

priyank5485: @HeartSaVioR Will change stream() and join as you suggested. Comma at the end is valid for json string. I tested it and it works fine. In fact, i did not want that extra code after the loop to remove the last character and hence kept it as is.

arunmahadevan:
+1

arunmahadevan:
@priyank5485 , thanks for the patch. If you see issues with the failover in HA, can you raise a follow up patch?

priyank5485:
@arunmahadevan yes. Still trying to figure out the root cause of the issue. Will raise follow up patch

PR #1184, opened 2018-01-19
ISSUE-1182 Avoid relying on default resolution while querying to Amba…

### LINKED ISSUE #1182: Avoid relying on default resolution while querying to Ambari Metrics
We found some issues on time-series value as well as single aggregated value in view mode. There are several valid suspicions, but what we could try out is leveraging contexts from both Storm and AMS (implementation detail).

@arunmahadevan provided nice observation that changing precision to MINUTES (which is also an interval Storm reports built-in metrics) where default resolution would be SECONDS makes result more accurate. 

We can set period by ourselves (not relying on AMS), with aware of that AMS has limitation of (time range, period) pair. So we still need to have lower precision (HOURS, DAYS), but not using SECONDS.
(Maybe it can throw exception if time range is not greater than 1 min.)

Below is a part of AMS codebase which explains the limit:
```
if (totalRowsRequested > PhoenixHBaseAccessor.RESULTSET_LIMIT) {
      throw new PrecisionLimitExceededException("Requested " +  condition.getMetricNames().size() + " metrics for "
        + numHosts + " hosts in " + precision +  " precision for the time range of " + range/1000
        + " seconds. Estimated resultset size of " + totalRowsRequested + " is greater than the limit of "
        + PhoenixHBaseAccessor.RESULTSET_LIMIT + ". Request lower precision or fewer number of metrics or hosts." +
        " Alternatively, increase the limit value through ams-site:timeline.metrics.service.default.result.limit config");
    }
```

HeartSaVioR:
…ri Metrics

* provide precision parameter when querying to Ambari Metrics
  * based on AMS default, with some modification
  * https://cwiki.apache.org/confluence/display/AMBARI/Metrics+Collector+API+Specification
  * no seconds
  * minutes: > 1 min, < 7 days
  * hours: >= 7 days, < 30 days
  * days: > 30 days

Change-Id: I621b9d065a09e07d6fb3ec0fab8fcc28d3ccf346

This closes #1182

### DIFF
--- a/streams/runners/storm/metrics/src/main/java/com/hortonworks/streamline/streams/metrics/storm/ambari/AmbariMetricsServiceWithStormQuerier.java
+++ b/streams/runners/storm/metrics/src/main/java/com/hortonworks/streamline/streams/metrics/storm/ambari/AmbariMetricsServiceWithStormQuerier.java
@@ -58,6 +58,10 @@ public class AmbariMetricsServiceWithStormQuerier extends AbstractTimeSeriesQuer
     public static final String DEFAULT_APP_ID = "nimbus";
     public static final String WILDCARD_ALL_COMPONENTS = "%";
 
+    enum Precision {
+        SECONDS, MINUTES, HOURS, DAYS
+    }
+
     private Client client;
     private URI collectorApiUri;
     private String appId;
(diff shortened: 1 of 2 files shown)


HeartSaVioR:
@arunmahadevan 
Could you please pull the patch and do some manual tests to see it helps? This change is based on your observation and I didn't have time to test it exhaustively.

HeartSaVioR:
@arunmahadevan Also applied aggregation function along with seriesAggregateFunction. Please test it out and let me know it also helps.

HeartSaVioR:
Closing this as origin issue was resolved from another patches.

Issue #1188, opened 2018-01-23
Handle accessing undefined object causing configuration page to freeze


shahsank3t:

shahsank3t:
https://github.com/hortonworks/streamline/pull/1189 fixes this issue

arunmahadevan:
+1

arunmahadevan:
@shahsank3t , thanks for the fix. Merged the changes to master.

PR #1189, opened 2018-01-23
ISSUE-1188 Handle accessing undefined object causing configuration page to freeze

shahsank3t:
This PR fixes https://github.com/hortonworks/streamline/issues/1188


arunmahadevan:
+1

PR #1195, opened 2018-01-29
ISSUE-1194 : While upgrading remove dangling entries in service when a service pool was removed in 3.0.2

raju-saravanan:
ISSUE 1194: While upgrading remove dangling entries in service when a service pool was removed in 3.0.2


raju-saravanan:
@HeartSaVioR: I add the necessary scripts for Oracle, MySQL, and Postgres and made sure that the bootstrap-storage is able to run the scripts against their respective databases. Can you please double check if this resolves the issue pointed out in https://hortonworks.jira.com/browse/BUG-95345 :)

arunmahadevan:
@raju-saravanan , this may not work since the FK is added in [v003__add_service_and_cluster_fk.sql](https://github.com/hortonworks/streamline/blob/master/bootstrap/sql/mysql/v003__add_service_and_cluster_fk.sql) which will get executed before the DELETE.

Are you going to raise a separate PR for the other issue where mod_hosts should be of TEXT type ?

raju-saravanan:
@arunmahadevan: Ok. Even if we modify v003 for the fix, if the user upgrades from 3.1 to m10, then v003 will not run as the database change log will say the latest version is v012, but fixing v003 directly would help to upgrade from 3.0.2 to m10 as we haven't introduced flyway in 3.0.2.

arunmahadevan:
@raju-saravanan , yes it would not run for 3.1 to m10 upgrade. But I assume the v003 script would not anyway run with the dirty rows. So for HDF 3.02 to HDF 3.1 upgrade, we need to add docs to manually remove the rows before the upgrade. After that the upgrade from 3.1 to M10 should happen without any issues (even without the DELETE).

Raised https://github.com/hortonworks/streamline/issues/1196 for addressing another issue around upgrade.

HeartSaVioR:
@arunmahadevan @raju-saravanan 
My idea was adding queries to v003 (prior to add FKs), assuming that we have last chance to get it in HDF 3.1. If we lost the chance, I agree documentation is the only way to address the issue.

raju-saravanan:
@arunmahadevan and @HeartSaVioR : Flyway before running the SQL migration, compares checksum of the scripts in sql/mysql directory with the checksums available in the DATABASE_CHANGE_LOG for each and every script, this would prevent us from making changes to the script directly. There is a way to disable this check for the entire migration, but the correct way would be to disable this check only for the scripts we want to fix like v003 and v005. I will see if there is a way to do this, else I would have to disable this check for all the scripts.

arunmahadevan:
@raju-saravanan , since we could get this fix in HDF 3.1, the steps have to be run manually for any upgrades to HDF 3.1. 

If we update the v003 script, it would take care of future upgrades from versions <= 3.0.2 to versions > HDF 3.1 without any other changes, correct? Then I don't think we need to worry about checksums since the v003 script would fail only for HDF 3.1 (if there are dirty records) and needs to be handled manually anyways. For migrations from HDF 3.1 to future versions we are good since the dirty data would not be present due to the newly added foreign key constraint.

arunmahadevan:
@raju-saravanan , based on the previous comment, can you make the changes to v003 script itself if it makes sense ?

raju-saravanan:
@arunmahadevan: I will disable the checksum checks for SAM, from a developer's perspective I don't see much use of it. I will update this PR shortly.

raju-saravanan:
@arunmahadevan: I have fixed the scripts can you please review? This PR requires [this](https://github.com/hortonworks/registry/pull/399) PR from schema registry for it work.

arunmahadevan:
+1, will merge after https://github.com/hortonworks/registry/pull/399 is merged. 

The streamline pom.xml has dependency on 0.5.0 version of registry. So either we should release 0.5.1 of registry with the fix or change the streamline pom version to 0.5.0-SNAPSHOT. @satishd , how do we want to proceed?

PR #1200, opened 2018-02-01
ISSUE-1199 Leverage schema branching in Kafka sources to pull schemas  from non master branches.

### LINKED ISSUE #1199: Leverage schema branching in Kafka sources to pull schemas from non master branches.
SR has new life cycle management features like branching. This enables users to tryout schema versions from different branches in dev/staging/qe etc scenarios.

satishd:
Kafka source and sink leverages schema branching available with registry 0.5.0 release.

Closes #1199

### DIFF
--- a/bootstrap/components/sinks/kafka-sink-topology-component.json
+++ b/bootstrap/components/sinks/kafka-sink-topology-component.json
@@ -41,7 +41,7 @@
         "tooltip": "Schema branch to be used fetching for writer schema versions. Default branch is MASTER",
         "type": "enumstring",
         "options": [],
-        "hint": "schema,override",
+        "hint": "dependOn-topic,schemaBranch",
         "defaultValue": "MASTER"
       },
       {
@@ -51,7 +51,7 @@
         "tooltip": "Version of schema for topic to use for serializing the message. Default is the latest version for the schema",
         "type": "enumstring",
         "options": [],
-        "hint": "dependsOn-topic,dependsOn-schemaBranch,schemaVersion"
+        "hint": "dependsOn-schemaBranch,schemaVersion"
       },
       {
         "uiName": "Security Protocol",
--- a/bootstrap/components/sinks/kafka-sink-topology-component.json
+++ b/bootstrap/components/sinks/kafka-sink-topology-component.json
@@ -35,13 +35,23 @@
         "hint": "schema,override"
       },
       {
-        "uiName": "Writer schema version",
+        "uiName": "Writer Schema Branch",
+        "fieldName": "schemaBranch",
+        "isOptional": true,
+        "tooltip": "Schema branch to be used fetching for writer schema versions. Default branch is MASTER",
+        "type": "enumstring",
+        "options": [],
+        "hint": "schema,override",
+        "defaultValue": "MASTER"
+      },
+      {
+        "uiName": "Writer Schema Version",
         "fieldName": "writerSchemaVersion",
         "isOptional": false,
         "tooltip": "Version of schema for topic to use for serializing the message. Default is the latest version for the schema",
         "type": "enumstring",
         "options": [],
-        "hint": "dependsOn-topic,schemaVersion"
+        "hint": "dependsOn-topic,dependsOn-schemaBranch,schemaVersion"
       },
       {
         "uiName": "Security Protocol",
--- a/bootstrap/components/sinks/kafka-sink-topology-component.json
+++ b/bootstrap/components/sinks/kafka-sink-topology-component.json
@@ -41,8 +41,7 @@
         "tooltip": "Schema branch to be used fetching for writer schema versions. Default branch is MASTER",
         "type": "enumstring",
         "options": [],
-        "hint": "dependOn-topic,schemaBranch",
-        "defaultValue": "MASTER"
+        "hint": "dependsOn-topic,schemaBranch"
       },
       {
         "uiName": "Writer Schema Version",
--- a/bootstrap/components/sources/kafka-source-topology-component.json
+++ b/bootstrap/components/sources/kafka-source-topology-component.json
@@ -26,7 +26,7 @@
         "hint": "dependsOn-securityProtocol"
       },
       {
-        "uiName": "Kafka topic",
+        "uiName": "Kafka Topic",
         "fieldName": "topic",
         "isOptional": false,
         "tooltip": "Kafka topic to read data from",
@@ -35,13 +35,23 @@
         "hint": "schema,override"
       },
       {
-        "uiName": "Reader schema version",
+        "uiName": "Reader Schema Branch",
+        "fieldName": "schemaBranch",
+        "isOptional": true,
+        "tooltip": "Schema branch to be used fetching for reader schema versions. Default branch is MASTER",
+        "type": "enumstring",
+        "options": [],
+        "hint": "schema,override",
+        "defaultValue": "MASTER"
+      },
+      {
+        "uiName": "Reader Schema Version",
         "fieldName": "readerSchemaVersion",
         "isOptional": false,
         "tooltip": "Version of schema for topic to use for deserializing the message. Default is version used by kafka producer to write the message",
         "type": "enumstring",
         "options": [],
-        "hint": "dependsOn-topic,schemaVersion"
+        "hint": "dependsOn-topic,dependsOn-schemaBranch,schemaVersion"
       },
       {
         "uiName": "Consumer Group Id",
--- a/bootstrap/components/sources/kafka-source-topology-component.json
+++ b/bootstrap/components/sources/kafka-source-topology-component.json
@@ -41,8 +41,7 @@
         "tooltip": "Schema branch to be used fetching for reader schema versions. Default branch is MASTER",
         "type": "enumstring",
         "options": [],
-        "hint": "dependsOn-topic,schemaBranch",
-        "defaultValue": "MASTER"
+        "hint": "dependsOn-topic,schemaBranch"
       },
       {
         "uiName": "Reader Schema Version",
--- a/bootstrap/components/sources/kafka-source-topology-component.json
+++ b/bootstrap/components/sources/kafka-source-topology-component.json
@@ -37,11 +37,10 @@
       {
         "uiName": "Reader Schema Branch",
         "fieldName": "schemaBranch",
-        "isOptional": true,
         "tooltip": "Schema branch to be used fetching for reader schema versions. Default branch is MASTER",
         "type": "enumstring",
         "options": [],
-        "hint": "schema,override",
+        "hint": "dependsOn-Topic,schemaBranch",
         "defaultValue": "MASTER"
       },
       {
@@ -51,7 +50,7 @@
         "tooltip": "Version of schema for topic to use for deserializing the message. Default is version used by kafka producer to write the message",
         "type": "enumstring",
         "options": [],
-        "hint": "dependsOn-topic,dependsOn-schemaBranch,schemaVersion"
+        "hint": "dependsOn-schemaBranch,schemaVersion"
       },
       {
         "uiName": "Consumer Group Id",
--- a/pom.xml
+++ b/pom.xml
@@ -164,7 +164,7 @@
         <phoenix.version>4.7.0.2.5.0.0-1245</phoenix.version>
         <redis.lettuce.version>3.4.2.Final</redis.lettuce.version>
         <scala.version>2.10.5</scala.version>
-        <schema-registry.version>0.5.1-SNAPSHOT</schema-registry.version>
+        <schema-registry.version>0.5.1.3.1.1.0-6</schema-registry.version>
         <slf4j.version>1.7.12</slf4j.version>
         <storm.version>1.2.0.3.1.0.0-406</storm.version>
         <woodstox-core-asl.version>4.4.1</woodstox-core-asl.version>
--- a/pom.xml
+++ b/pom.xml
@@ -164,7 +164,7 @@
         <phoenix.version>4.7.0.2.5.0.0-1245</phoenix.version>
         <redis.lettuce.version>3.4.2.Final</redis.lettuce.version>
         <scala.version>2.10.5</scala.version>
-        <schema-registry.version>0.5.0</schema-registry.version>
+        <schema-registry.version>0.5.1-SNAPSHOT</schema-registry.version>
         <slf4j.version>1.7.12</slf4j.version>
         <storm.version>1.2.0.3.1.0.0-406</storm.version>
         <woodstox-core-asl.version>4.4.1</woodstox-core-asl.version>
--- a/streams/service/src/main/java/com/hortonworks/streamline/streams/service/SchemaResource.java
+++ b/streams/service/src/main/java/com/hortonworks/streamline/streams/service/SchemaResource.java
@@ -121,16 +121,14 @@ public class SchemaResource {
     private Response doGetAllSchemaVersionForBranch(String schemaName, String branchName) {
         try {
             LOG.info("Get all versions for schema : {}", schemaName);
-            List<Byte> states = Collections.singletonList(SchemaVersionLifecycleStates.ENABLED.getId());
-            Collection<SchemaVersionInfo> schemaVersionInfos = schemaRegistryClient.getAllVersions(effectiveBranchName(branchName), schemaName, states);
-            Collection<String> schemaVersions = new ArrayList<>();
+            Collection<SchemaVersionInfo> schemaVersionInfos = schemaRegistryClient.getAllVersions(effectiveBranchName(branchName), schemaName);
+            LOG.debug("Received schema versions [{}] from schema registry for schema: {}", schemaVersionInfos, schemaName);
+
             if (schemaVersionInfos != null && !schemaVersionInfos.isEmpty()) {
-                LOG.debug("Received schema versions from schema registry: {}", schemaVersionInfos);
-                schemaVersions = schemaVersionInfos.stream().map(x -> x.getVersion().toString()).collect(Collectors.toList());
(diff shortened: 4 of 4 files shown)


satishd:
@shahsank3t You may need to make few changes on UI w.r.t new configuration. You can review kafka source/sink changes and let me know if there any changes required.

API to get branches for given schema name.
```
GET /v1/schemas/{schemaName}/branches 
```

API to get versions for a given branch name.
```
GET /v1/schemas/{schemaName}/{branchName}/versions numbers.
```

satishd:
This change requires APIs from registry 0.5.1-SNAPSHOT which is in [HDF-3.1-maint](https://github.com/hortonworks/registry/tree/HDF-3.1-maint) branch.

bootstrap/components/sinks/kafka-sink-topology-component.json
```diff
@@ -35,13 +35,23 @@
...
       },
       {
-        "uiName": "Writer schema version",
+        "uiName": "Writer Schema Branch",
+        "fieldName": "schemaBranch",
+        "isOptional": true,
+        "tooltip": "Schema branch to be used fetching for writer schema versions. Default branch is MASTER",
+        "type": "enumstring",
+        "options": [],
+        "hint": "schema,override",
```
priyank5485: We probably need to change the hint to schemaBranch and add a dependOn-topic hint here so UI can fetch all the branches for a given topic/schema.

bootstrap/components/sinks/kafka-sink-topology-component.json
```diff
@@ -35,13 +35,23 @@
...
+      },
+      {
+        "uiName": "Writer Schema Version",
         "fieldName": "writerSchemaVersion",
         "isOptional": false,
         "tooltip": "Version of schema for topic to use for serializing the message. Default is the latest version for the schema",
         "type": "enumstring",
         "options": [],
-        "hint": "dependsOn-topic,schemaVersion"
+        "hint": "dependsOn-topic,dependsOn-schemaBranch,schemaVersion"
```
priyank5485: Probably remove dependsOn-topic here if its a selection in order. I would imagine that user will first pick topic that UI will use to get all branches and then once user selects the branch we get all versions for the branch.  In that case schemaVersion can now only depend on schemaBranch. However, I see that schemaBranch is optional field. In that case it makes sense.

streams/service/src/main/java/com/hortonworks/streamline/streams/service/SchemaResource.java
```diff
@@ -190,4 +229,27 @@ public Response getSchemaForVersion(@PathParam("schemaName") String schemaName,
...
     }
 
+    @GET
+    @Path("/{schemaName}/branches")
+    @Timed
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response getSchemaBranches(@PathParam("schemaName") String schemaName,
+                                      @Context SecurityContext securityContext) {
+        try {
+            Collection<SchemaBranch> schemaBranches = schemaRegistryClient.getSchemaBranches(schemaName);
```
priyank5485: Looks like we will always return MASTER branch. In that case may be make the schemaBranch UI field mandatory  and set defaultValue to MASTER ?

streams/service/src/main/java/com/hortonworks/streamline/streams/service/SchemaResource.java
```diff
@@ -109,18 +116,23 @@ public Response postStreamsSchema(StreamsSchemaInfo streamsSchemaInfo,
     @Produces(MediaType.APPLICATION_JSON)
     public Response getKafkaSourceSchema(@PathParam("topicName") String topicName,
```
priyank5485: This endpoint was probably never used before. I am okay to keep it as it is to return latest version on MASTER but may be we can just get rid of it?

streams/service/src/main/java/com/hortonworks/streamline/streams/service/SchemaResource.java
```diff
@@ -133,24 +145,37 @@ public Response getKafkaSourceSchema(@PathParam("topicName") String topicName,
         }
     }
 
+    @GET
+    @Path("/{topicName}/{branchName}")
+    @Timed
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response getKafkaSourceSchemaForBranch(@PathParam("topicName") String topicName,
```
priyank5485: We probably dont need this endpoint if user is forced to select a version from UI. Current behavior is user first selects a topic which populates the version drop down and then once a version is selected the schema is displayed in UI for that version.

priyank5485:
@satishd I left some comments. However, my understanding of how you designed it to work might not be correct. The way it currently works is once user selects a topic, UI fetches all the versions for that topic to display in the drop down and once a version is selected the schema is fetched and displayed in UI. I am guessing that now once a topic is selected all branches will be displayed first and then once a branch is selected all versions for that branch are displayed and once a version is selected the schema is displayed in the UI. If thats the case then please check out my comments and see if they make sense.

streams/service/src/main/java/com/hortonworks/streamline/streams/service/SchemaResource.java
satishd: I have added respective APIs for branching but it seems the current API of "/{topicName}" is also not used. In that case, we can remove this API.

satishd:
@priyank5485 Thanks for reviewing. 
This PR is intended to make the below flow and we are on the same page with that based on your comment.
topic ---> branches --> versions ---> schema

I was not sure about the UI semantics in source/sink yamls for which I asked Sanket to take look. I will update with the suggested changes.

satishd:
@priyank5485 Addressed review comments with the latest commit.

bootstrap/components/sinks/kafka-sink-topology-component.json
```diff
@@ -35,13 +35,23 @@
         "hint": "schema,override"
       },
       {
-        "uiName": "Writer schema version",
+        "uiName": "Writer Schema Branch",
+        "fieldName": "schemaBranch",
+        "isOptional": true,
```
priyank5485: @shahsank3t Do we need to set isOptional to false for UI to work correctly? If its optional then there is a chance that the branch user selected first is now deleted in SR. In that case we have seen UI silently failing on that component and getting runtime errors. Can you check the behavior of UI not failing in such scenarios?

bootstrap/components/sinks/kafka-sink-topology-component.json
```diff
@@ -35,13 +35,23 @@
...
+        "hint": "dependOn-topic,schemaBranch",
+        "defaultValue": "MASTER"
+      },
+      {
+        "uiName": "Writer Schema Version",
         "fieldName": "writerSchemaVersion",
         "isOptional": false,
         "tooltip": "Version of schema for topic to use for serializing the message. Default is the latest version for the schema",
         "type": "enumstring",
         "options": [],
```
priyank5485: @shahsank3t Can you also check the same for this? Meaning, if UI will work when a particular version of schema that user previously selected is not returned anymore in the list because it was deleted. In that case we need to make sure user is asked for an input again.

bootstrap/components/sources/kafka-source-topology-component.json
```diff
@@ -35,13 +35,22 @@
         "hint": "schema,override"
       },
       {
-        "uiName": "Reader schema version",
+        "uiName": "Reader Schema Branch",
+        "fieldName": "schemaBranch",
```
priyank5485: Same comment for isOptional field as i made in kafka sink

streams/service/src/main/java/com/hortonworks/streamline/streams/service/SchemaResource.java
```diff
@@ -102,55 +109,27 @@ public Response postStreamsSchema(StreamsSchemaInfo streamsSchemaInfo,
...
     public Response getAllSchemaVersions(@PathParam("schemaName") String schemaName,
                                          @Context SecurityContext securityContext) {
+        return doGetAllSchemaVersionForBranch(schemaName, null);
+    }
+
+    private Response doGetAllSchemaVersionForBranch(String schemaName, String branchName) {
         try {
             LOG.info("Get all versions for schema : {}", schemaName);
-            Collection<SchemaVersionInfo> schemaVersionInfos = schemaRegistryClient.getAllVersions(schemaName);
+            List<Byte> states = Collections.singletonList(SchemaVersionLifecycleStates.ENABLED.getId());
```
priyank5485: @satishd One other question i had after refreshing on SR branching feature is that we allow a schema branch version in a "in review" or other custom state to be used by producer/consumer. Here, we are just returning ENABLED schemas. Is there a reason for that or should we be allowing schema versions in those states?

satishd: @priyank5485 Only enabled schema versions are allowed, any other state is not considered ready to be consumed when a topology is configured. Users use non master branches to try out the changes by enabling that version and merge it to master only if they find that version is working as expected in their topology.

satishd:
@priyank5485 
>We are using only schema name and version and so we should not need any change since UI will force the user to select a schema version. But just making sure that we dont need branch name here since a version of a schema on a particular branch can be uniquely identified using schema name and version. 

Right. any versioned schema can be identified with schema name and version number and no need to have any branching information as version number is unique among all branches of a schema metadata. 

>The branch name for Kafka source and sink will still be persisted as a config field so UI can operate correctly whenever we open the config dialog in SAM for source or sink.

Right.

satishd:
@priyank5485 @shahsank3t [getKafkaSourceSchema API](https://github.com/hortonworks/streamline/pull/1200/commits/c7fb8127d41a5d9a338b6f3b1957ada6a6521f24#diff-a7543aedf525304690e445f76dec052cR117) is removed in latest commit based on a review comment. Do not we need that for UI if the version config is treated as optional?

priyank5485:
@satishd I am looking at kafka source and sink json and reader and writer schema version fields are both mandatory. We can keep them as mandatory. That api being deleted should not be a problem anyway since now the UI will always get the branches first for a topic(default MASTER) and force the user to pick a version from that. Do you still see the deleted api being used anywhere? Also, I am presuming that older SR schemas will all be migrated to MASTER branch as enabled in newer version.

streams/service/src/main/java/com/hortonworks/streamline/streams/service/SchemaResource.java
priyank5485: Based on this document https://docs.google.com/document/d/1qi3o9ADxkySZn-PZJiWTyfcvdBraW-dHGwMRm-xZeyA/edit# it seems that Archived, ChangeRequired, etc states are where it is available to producers/consumers. Given that SAM Kafka components are producers/consumers, should not we allow them to play with a schema on a branch which is under review?

Also, based on the document it seems that once a schema is enabled it can be merged to master. If its anyway mergeable to master then why expose the branches of a schema in SAM? We can go back to old UI and on server side we just get versions for the master branch. I guess the main question is "Are we allowing SAM users to try out schemas that are on a branch and under review?" If not and if we only want enabled schemas(that are anyway mergeable to master) then this PR should have only changes for server side.

satishd: @priyank5485 

>Also, based on the document it seems that once a schema is enabled it can be merged to master. If its anyway mergeable to master then why expose the branches of a schema in SAM? 

It depends on how are verification workflows are designed in the applications. There are different ways users want to use lifecycle states. That is the reason SR client APIs give flexible APIs to fetch schema versions with whatever states they are interested in. Default or custom review states can be considered outside of SAM and it is enabled only when it can be tried in applications like SAM. If that application fails with this schema then it can be disabled and eventually discarded. Enabled schema does not need to be always merged. Merge operation can be done based on the result of trying out on SAM or any other respective application.

So, it depends on how users want to try out branched schema versions. I am OK to allowing non-master branch schema versions with states other than archived/deleted state in SAM.

satishd:
>@satishd I am looking at kafka source and sink json and reader and writer schema version fields are both mandatory. We can keep them as mandatory. That api being deleted should not be a problem anyway since now the UI will always get the branches first for a topic(default MASTER) and force the user to pick a version from that. Do you still see the deleted api being used anywhere? Also, I am presuming that older SR schemas will all be migrated to MASTER branch as enabled in newer version.

@priyank5485 Right, older schemas in HDF-3.0.2 or earlier are mapped to MASTER branch in HDF-3.1 release. Even if users upgrade from HDF-3.0.x releases to HDF-3.1+ then those schemas will be migrated to MASTER branch.

satishd:
@priyank5485 Thanks for detailed review, addressed review comments with latest commit.

@shahsank3t You may want to make respective UI changes in another PR. I will merge this once @priyank5485 reviews the latest changes.

pom.xml
```diff
@@ -164,7 +164,7 @@
         <phoenix.version>4.7.0.2.5.0.0-1245</phoenix.version>
         <redis.lettuce.version>3.4.2.Final</redis.lettuce.version>
         <scala.version>2.10.5</scala.version>
-        <schema-registry.version>0.5.0</schema-registry.version>
+        <schema-registry.version>0.5.1.3.1.1.0-6</schema-registry.version>
```
satishd: This will be changed to 0.5.1 once it is released(plan to do that by this weekend viz 12th Feb)

priyank5485:
@satishd Thanks for the detailed explanation.This looks good. +1.
Just one last change which we might need to make is change schema branch name mandatory in source and sink jsons. I believe we will wait for @shahsank3t comments and let him decide and change it. We can check with him and merge this in as long as he can get the UI related changes in time.

bootstrap/components/sinks/kafka-sink-topology-component.json
shahsank3t: @priyank5485 This is a mandatory field, so if the previously selected version is no longer available in the dropdown list, user will be forced to pick another one by the UI

bootstrap/components/sources/kafka-source-topology-component.json
shahsank3t: yes, need to add `isOptional` field with false.

bootstrap/components/sources/kafka-source-topology-component.json
```diff
@@ -35,13 +35,22 @@
...
+        "tooltip": "Schema branch to be used fetching for reader schema versions. Default branch is MASTER",
+        "type": "enumstring",
+        "options": [],
+        "hint": "dependsOn-Topic,schemaBranch",
```
shahsank3t: @satishd 
minor nit: T in `dependsOn-Topic` should be lower case.

shahsank3t:
@priyank5485 @satishd 
Left couple of comments which are basically minor nits.

satishd:
@priyank5485 @shahsank3t Addressed review comments. You can start working on UI changes in a separate PR based out of this PR. We can close both PRs once it is verified.

bootstrap/components/sinks/kafka-sink-topology-component.json
```diff
@@ -35,13 +35,23 @@
...
       {
-        "uiName": "Writer schema version",
+        "uiName": "Writer Schema Branch",
+        "fieldName": "schemaBranch",
+        "isOptional": false,
+        "tooltip": "Schema branch to be used fetching for writer schema versions. Default branch is MASTER",
+        "type": "enumstring",
+        "options": [],
+        "hint": "dependOn-topic,schemaBranch",
+        "defaultValue": "MASTER"
```
shahsank3t: @satishd due to this "defaultValue", bootstrap.sh is failing to add source/sink. We have hardcode the default value of branch to "MASTER" in UI so if you want, can remove this field or please have a look why it's failing.

bootstrap/components/sources/kafka-source-topology-component.json
```diff
@@ -35,13 +35,23 @@
...
       {
-        "uiName": "Reader schema version",
+        "uiName": "Reader Schema Branch",
+        "fieldName": "schemaBranch",
+        "isOptional": false,
+        "tooltip": "Schema branch to be used fetching for reader schema versions. Default branch is MASTER",
+        "type": "enumstring",
+        "options": [],
+        "hint": "dependsOn-topic,schemaBranch",
+        "defaultValue": "MASTER"
```
shahsank3t: @satishd same thing over here, failing the source to be registered during bootstrap.

bootstrap/components/sinks/kafka-sink-topology-component.json
priyank5485: @satishd @shahsank3t Its failing because its expecting the default value to be one of the items from the options. But in this case because options is an empty list which is populated later by making an API call, its throwing an error. We can just remove the default value in the json spec since thats how other such fields work.

satishd: @priyank5485 @shahsank3t Right, we can remove default values for this mandatory field.

bootstrap/components/sinks/kafka-sink-topology-component.json
```diff
@@ -35,13 +35,22 @@
...
       },
       {
-        "uiName": "Writer schema version",
+        "uiName": "Writer Schema Branch",
+        "fieldName": "schemaBranch",
+        "isOptional": false,
+        "tooltip": "Schema branch to be used fetching for writer schema versions. Default branch is MASTER",
+        "type": "enumstring",
+        "options": [],
+        "hint": "dependOn-topic,schemaBranch"
```
joylyn: @satishd Need a minor change, the hint for " writer schema branch" is missing an 's' in "dependsOn-topic", hence I got an issue on UI. Could you make this change and verify ?

satishd: @joylyn Pushed the required change in latest commit.

PR #1203, opened 2018-02-06
ISSUE-1168 Support Kafka metrics in SAM

sameer79:
Kafka lag offset graph shows in the  Kafka-Source component node

kafkalagoncomponent

when Kafka-Source componenet is selected, Kafka lag offset graph shown in the bottom panel.

bottompanel

@arunmahadevan  @shahsank3t please review and comment

### DIFF
--- a/webservice/src/main/resources/app/scripts/containers/Streams/Metrics/TopologyComponentMetrics.jsx
+++ b/webservice/src/main/resources/app/scripts/containers/Streams/Metrics/TopologyComponentMetrics.jsx
@@ -16,7 +16,6 @@ import React,{Component} from 'react';
 import _ from 'lodash';
 import TimeSeriesChart from '../../../components/TimeSeriesChart';
 import d3 from 'd3';
-import MetricsREST from '../../../rest/MetricsREST';
 import Utils from '../../../utils/Utils';
 import ContentScrollableComponent from '../../../components/ContentScrollableComponent';
 
@@ -106,6 +105,7 @@ class TopologyComponentMetrics extends Component {
     const queueData = [];
     const processTimeData = [];
     const completeLatency = [];
+    const kafkaLagOffsetData=[];
 
     if(timeSeriesMetrics) {
       const {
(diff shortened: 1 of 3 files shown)


webservice/src/main/resources/app/scripts/containers/Streams/Metrics/TopologyComponentMetrics.jsx
```diff
@@ -239,6 +247,20 @@ class TopologyComponentMetrics extends Component {
               {this.state.loadingRecord ? loader : this.getGraph('Queue', queueData, 'step-before', showMetrics)}
             </div>
           </div>
+          {
+            compData.parentType === 'SOURCE' && compData.currentType === "Kafka"
+            ? <div className="component-metric-graph">
+                <div style={{textAlign: "left"}}>kafka Lag Offset</div>
```
shahsank3t: nit: k should be in uppercase

omkreddy:
maybe we should rename  "Kafka Lag offset" to "Kafka Consumer Lag" or "Kafka Offset Lag"

shahsank3t:
@omkreddy @arunmahadevan @HeartSaVioR 
http://jsoneditoronline.org/?id=23523623dd16056ad74460824b5822fd
Above is the response we get from the kafka metrics API. 

FYI: on the UI, we are currently using `lag` and `offset` from the response so there will be 2 lines in the graph that can be viewed. 

Let us know if that needs to be changed or whether we should also be using `logsize` from the API response. Thanks.

omkreddy:
In my opinion, "Lag" should be sufficient.

arunmahadevan:
@shahsank3t , the changes looks good. I think lag is more useful and for now that should be sufficient. Can you rename it to "Kafka offset lag" as @omkreddy suggested?

sameer79:
@omkreddy @arunmahadevan @shahsank3t  address the comment and squash commit as well, ready for merge.

arunmahadevan:
+1

PR #1204, opened 2018-02-06
ISSUE-1199-UI: Leverage schema branching in Kafka sources to pull sch…

joylyn:
…emas from non master branches.

Added UI changes for #1199


joylyn:
@satishd  I have updated this PR with a change in the case of older SR schemas which are mapped to MASTER branch.

PR #1221, opened 2018-02-21
ISSUE-1220 Exclude 0.0 values while aggregating metrics via SUM/AVG i…

### LINKED ISSUE #1220: Exclude 0.0 values while aggregating metrics via SUM/AVG in view mode
The result of average latency is sometimes far from actual value in Storm UI, which is due to account 0.0 values while calculating average. 

While it may not be accurate when we exclude valid 0.0 (we can't determine whether the value is valid or reflecting that there's no metric recorded in time period), but it may be much better than messing up with including them.

HeartSaVioR:
…n view mode

This closes #1220


arunmahadevan:
+1

Issue #1217, opened 2018-02-16
Projection processor should support UDF with one or more arguments


arunmahadevan:
Support projections with one or more arguments, constants and nested invocations. Some examples

- upper(name)
- concat(first, last)
- substring(str, 0, 10)
- lower(trim(text))

One option could be to have a code text editor with pop up and use the type information available in the UDF api to prompt for the arguments.

arunmahadevan:
@shahsank3t , you should be able to pass functions as "expr" objects within projections as follows:

```
"projections":[{"expr":"LOWER(driverName) AS DN"},{"expr":"driverId"}]
```

arunmahadevan:
Once #1224 is merged, we could introduce similar changes for projection processor.

sameer79:
Raised PR : https://github.com/hortonworks/streamline/pull/1255

arunmahadevan:
thanks @sameer79 , merged #1255 to master

PR #1224, opened 2018-02-22
ISSUE-1218 Rule processor should support writing complex conditions

sameer79:
@arunmahadevan  - There’s no UI validations on creating the sql query right now due to some Javascript limitations. We are looking into other turn around solutions (in any) to fix but meanwhile, either backend needs to handle the validation or user needs to be precise while creating the rule..

### DIFF
--- a/webservice/src/main/resources/app/scripts/components/CommonCodeMirror.jsx
+++ b/webservice/src/main/resources/app/scripts/components/CommonCodeMirror.jsx
@@ -1,0 +1,298 @@
+/**
+  * Copyright 2017 Hortonworks.
+  *
+  * Licensed under the Apache License, Version 2.0 (the "License");
+  * you may not use this file except in compliance with the License.
+  * You may obtain a copy of the License at
+  *   http://www.apache.org/licenses/LICENSE-2.0
+  * Unless required by applicable law or agreed to in writing, software
+  * distributed under the License is distributed on an "AS IS" BASIS,
+  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  * See the License for the specific language governing permissions and
+  * limitations under the License.
+**/
+import React, {Component} from 'react';
+import PropTypes from 'prop-types';
+import _ from 'lodash';
+import CodeMirror from 'codemirror';
+import 'codemirror/mode/javascript/javascript';
+import jsonlint from 'jsonlint';
+import sql from 'codemirror/addon/hint/sql-hint';
+import javascriptHint from 'codemirror/addon/hint/javascript-hint';
+import lint from 'codemirror/addon/lint/lint';
+import showHint from 'codemirror/addon/hint/show-hint';
+import placeholder from 'codemirror/addon/display/placeholder';
+import panel from 'codemirror/addon/display/panel';
+
+CodeMirror.registerHelper("lint", "json", function(text) {
+  var found = [];
+  var {parser} = jsonlint;
+  parser.parseError = function(str, hash) {
+    var loc = hash.loc;
+    found.push({
+      from: CodeMirror.Pos(loc.first_line - 1, loc.first_column),
+      to: CodeMirror.Pos(loc.last_line - 1, loc.last_column),
+      message: str
+    });
+  };
+  try {
+    jsonlint.parse(text);
+  } catch (e) {}
+  return found;
+});
+
+export default class CommonCodeMirror extends Component{
+  constructor(props){
+    super(props);
+    this.modeList = {
+      javascript : {
+        lineNumbers: true,
+        mode: "javascript",
+        styleActiveLine: true,
+        lint: true
+      },
+      json : {
+        lineNumbers: true,
+        mode: "application/json",
+        styleActiveLine: true,
+        gutters: ["CodeMirror-lint-markers"],
+        lint: true,
+        autofocus: true
+      },
+      sql : {
+        mode: "text/x-mysql",
+        indentWithTabs: true,
+        smartIndent: true,
+        lineNumbers: true,
+        matchBrackets : true,
+        extraKeys : {"'@'": "autocomplete"}
+      }
+    };
+    this.escapeHintChar = /[`~!@#$%^&0-9()_|\¿?;:'",\{\}\[\]\\]/gi;
+    this.state = {textValue : ''};
+    this.initialFlag= true;
+  }
+
+  componentWillReceiveProps(newProps) {
+    if (this.props.editMode && this.initialFlag) {
+      this.setState({textValue : newProps.value}, () => {
+        this.codeWrapper.setValue(newProps.value);
+        this.initialFlag= false;
+      });
+    }
+  }
+
+  componentDidMount = () => {
+    const {hintOptions,modeType,modeOptions,height,width,panelText} = this.props;
+    const that = this;
+    let tempMode = this.mergeModeTypeAndOptions(modeType,modeOptions);
+
+    !!panelText ? tempMode.theme = "default panel" : null;
+
+    this.codeWrapper = CodeMirror.fromTextArea(this.refs.codeDiv, tempMode);
+    this.codeWrapper.setSize(width || null, height || null);
+
+    if(!!panelText){
+      const node = this.creatPanel('top');
+      this.codeWrapper.addPanel(node, {position: 'top', stable: true});
+    }
+
+    if(that.props.modeType !== 'json'){
+      that.codeWrapper.on('inputRead', (cm, event) => {
+        const type =  that.props.modeType === undefined ? 'javascript' : that.props.modeType;
+        let orig = CodeMirror.hint[type];
+        const val = event.text[0].trim();
+        let inner = orig(cm) || {from: cm.getCursor(), to: cm.getCursor(), list: []};
+        if(!that.escapeHintChar.test(val) && val !== "" && _.isNaN(parseInt(val))){
+          CodeMirror.showHint(cm, (cm) => {
+            return that.populateHintOptions(cm,inner);
+          },{completeSingle: false});
+        }
+      });
+    }
+
+    this.codeWrapper.on('change', (cm, event) => {
+      that.props.callBack(cm.getValue());
+      this.removeHintPopUp();
+    });
+
+    CodeMirror.commands.autocomplete = (cm) => {
+      const {modeType,hintOptions} = that.props;
+      if(modeType !== 'json' && hintOptions !== undefined && hintOptions.length){
+        CodeMirror.showHint(cm,CodeMirror.hint.ownHint,{completeSingle: false});
+      } else if(modeType !== 'json' && hintOptions === undefined){
+        CodeMirror.showHint(cm,CodeMirror.hint[modeType],{completeSingle: false});
+      }
+    };
+
+    CodeMirror.hint.ownHint = (cm) => {
+      const {hintOptions} = this.props;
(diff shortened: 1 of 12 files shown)


sameer79:
screenshot of changes.

screenshot from 2018-02-22 18-26-29

arunmahadevan:
@sameer79 thanks making progress on this, it looks great. I played around a bit and adding some feedback below.

1. Is there a way to force the pop up to appear?  I think pressing a special char (e.g `@`) should bring it up. Maybe good to bring it up only when the special char is pressed so that it does not interfere with what user may want to type otherwise. See comment 4.

2. the pop up displays 'displayName' (e.g SUBSTRING) but once selected it puts the 'name' in the text box (e.g. SUBSTRING_FN) . I think we should keep the 'displayName' itself in the text box. The translation needs to happen in the UI or backend. If UI cannot handle it, raise a backend issue and assign it to me.

3. If possible display only the relevant arguments. E.g Since the argument for TRIM is a string, no need to display integer or long fields in the popup. If its too much effort ignore it for now.

4.  The autocompletion should be turned off while typing literals. E.g. I was not able to type `= 'ABC'`. It autocompletes to `'abABS`. Maybe showing the popup only on pressing the special char would solve this issue.

5. For now show only `AND`, `OR` sql key words and remove all the others like `AS`, `AT`, `IF` etc. These are not tested and not required for writing conditions.

6. Include the following comparison operators in the pop up. (>, <, >=, <=, =, <>)

7. The pop up does not differentiate between overloaded functions. For example it displays `SUBSTRING(STRING)` in the popup twice, but the UDFs that we have are following 1.SUBSTRING("STRING","INTEGER") and 2. SUBSTRING("STRING","INTEGER","INTEGER")

8. Have some way to display the function description so that users can know what are the argument types etc. Maybe when we hover over the specific function in the pop-up, display the description.

sameer79:
@arunmahadevan  @shahsank3t  Address the comment and PR is updated with the partial fix, details are below.

1.  As per the discussion with @shahsank3t, we are not using special char('@') to bring up the popup.
2.  Done.
3.  Looking into this issue.
4.  Done.
5.   Done.
6.   API to get the operator, reference to PR https://github.com/hortonworks/streamline/pull/1225 OR we need to hardcode it?
7.  Done.
8.  Looking into this issue.
9. Done

After all the issue gets fixed, ill squash the commit, open to any suggestion.

arunmahadevan:
@sameer79 , will retest after you make the changes. 

1. As long as the pop-up shows up only when its required it should be good. Otherwise we need to think of some way to show it only on a key press.

6. For now you can hardcode it to the following binary operators (>, <, >=, <=, =, <>, +, -, *, /)

sameer79:
@arunmahadevan  @shahsank3t  Address comments and update PR as well.

1] If possible display only the relevant arguments. E.g Since the argument for TRIM is a string, no need to display an integer or long fields in the popup. `Ignore for now`

2] Have some way to display the function description so that users can know what are the argument types etc. Maybe when we hover over the specific function in the pop-up, display the description. `Done`

3] Binary operators added to hints `Done`

If any use case or issues do let me know, thanks in advanced.

arunmahadevan:
>3] Binary operators added to hints Done

I see the conditional operators. Arithmetic (+, -, /, *) are missing. Can you check?

>Instead of "Create Query" and "Sql query goes here", it can be "Create condition" and "Sql condition goes here...". You could also add info about shortcuts in the default text if we decide to support it.

I still see the following:
screen shot 2018-02-27 at 7 03 08 pm

>the pop up displays 'displayName' (e.g SUBSTRING) but once selected it puts the 'name' in the text box (e.g. SUBSTRING_FN) . I think we should keep the 'displayName' itself in the text box. The translation needs to happen in the UI or backend. If UI cannot handle it, raise a backend issue and assign it to me.

Are you passing the query with display name ? If so some translation needs to happen at the backend. Can you raise an issue for this?


@HeartSaVioR, can you apply these changes and play around with the rules processor and provide some feedback?

sameer79:
> Instead of "Create Query" and "Sql query goes here", it can be "Create condition" and "Sql condition  goes here...". You could also add info about shortcuts in the default text if we decide to support it. 

 My bad the file got check-out somehow. I'll update that.
> I see the conditional operators. Arithmetic (+, -, /, *) are missing. Can you check?

Options are there  (+, -, /, *). 
Search works on "char" or "string"  present in the given hints, E.g if hints has ("+","+=")  and user type "+",  then it will show two options, but if it has ("+") and user type "+",  This won't trigger the popup and it gets selected from the hints.

> Are you passing the query with display name? If so some translation needs to happen at the backend. Can you raise an issue for this?

Yes, I have raised backend issue  https://github.com/hortonworks/streamline/issues/1229, and updated the PR as well.

HeartSaVioR:
Here's my 2 cents as a tester of the PR:

* Typing `f` (or `F`) autocompletes to `FLOOR` without any selection which is annoying.
  * It just completes the word when there's only one matched to current word being typed.
  * `f` is not just one to make me annoying: `h` goes to `CHAR_LENGTH`.
  * we would be better to show selection even there's only one matched, and hit `enter` to apply the selection.
* If possible, it would be better to open suggestion whenever we press combination of key: Apache Zeppelin uses `CTRL + .`
* It is bad UX to let user memorize fields: listing fields with types would be good to refer.
* If possible, it would be better to add `(` at the end when users select function to explicitly show they select a function.

It would be better if we mimic the UX of widely used editor, Apache Zeppelin in this case.

HeartSaVioR:
And I feel more convenient to use old rule definition dialog on simple case. Do we want to force users to input a part of SQL statement?

HeartSaVioR:
I'll do the manual test via running topology after we are OK with UX.

sameer79:
@arunmahadevan @shahsank3t @HeartSaVioR address the comments.

> we would be better to show selection even there's only one matched, and hit enter to apply the selection.

Done

> If possible, it would be better to open suggestion whenever we press combination of key: Apache Zeppelin uses CTRL + .

on '@' keypress shows all hints.

> If possible, it would be better to add ( at the end when users select function to explicitly show they select a function.

Done

> It is a bad UX to let user memorize fields: listing fields with types would be good to refer.

Reference of the previous screenshot, we are showing the Type below the fields name.

> End users also would be confused how much they need to input. We need to guide that user input will be matched as where statement and should not provide full SQL statement.

Shows pre-select query screenshot below: 


pre-select-query

let me know if I miss some use case, Thanks in advance

shahsank3t:
@sameer79 - can you add a tooltip icon on the top right of the text box saying that "Type @ to see all the available options" so the user doesn't have to assume or figure out which key needs to be pressed.

HeartSaVioR:
> It is a bad UX to let user memorize fields: listing fields with types would be good to refer.
>> Reference of the previous screenshot, we are showing the Type below the fields name.

I meant mainly about name. Users need to know any character to open filtered suggestion for fields. Happy to see `@` will provide suggestion, whereas for me showing schema from left side may help much. But it is just a 2 cent and I'm OK if others are OK with keeping it as it is.

Otherwise it looks good. I'll try to spend time to test manually but I'll be off tomorrow so anyone giving +1 would be OK to merge without my +1.

sameer79:
@shahsank3t  Address the comment and squash all commits.
updated the PR as well.

@HeartSaVioR @arunmahadevan @shahsank3t 
Reference to the below screenshot user can identify the functions, arguments... through the circle Highlighter in the hints options E.g.
- "F": Is the UDF function
- "A": Is an argument or field name
- "S": Is an SQL conditional Operator

screenshot

arunmahadevan:
@sameer79 ,

- Can you display cursor at the end to show where the next character is going appear ?
- I had an expression `CHAR_LENGTH(driverName) + 5 >= 100`. Placed the cursor before 'driverName' and selected `CONCAT` function and now the rest of the expression is wiped out and it becomes `CHAR_LENGTH(CONCAT(`
- Need to wait till https://github.com/hortonworks/streamline/issues/1229 is merged since some functions like CONCAT throws exceptions currently.

arunmahadevan:
@sameer79 , Looks like nested fields are not displayed in the editor. For example not able to see the "votes" field in the [yelp_review](https://github.com/hortonworks/registry/blob/master/examples/schema-registry/avro/src/main/resources/yelp_review.avsc) schema. I think we can display the top level field "votes" and if the user selects votes and presses "." display the first level nested fields and so on.

sameer79:
@arunmahadevan  @shahsank3t  @HeartSaVioR Address all comments and updated PR as well.

screenshot for nested fields below: 

nestedfields

arunmahadevan:
@sameer79 , thanks for fixing the issues. It mostly looks fine. One observation - After typing "UPPER(driverName" by selecting "driverName" from pop up, the cursor disappears.

Need to wait till #1229 gets merged before merging this.

@HeartSaVioR , do you have any additional comments?

arunmahadevan:
@sameer79 , can you make similar changes for "branch" processor as well ?

HeartSaVioR:
@arunmahadevan I just merged #1233 so that we can try end-to-end test with this patch.

sameer79:
@arunmahadevan  @shahsank3t @HeartSaVioR 
Done with the Branch processor change and some code cleanup.

FYI Both the processor has "No UI Validation"  for SQL query.

Test both the processors on your side as well, Thanks in advanced.

arunmahadevan:
>Not able to reproduce it. I assume you might be using the mouse click For the selection from the >popup. On a single click, it works perfectly, but on double click, it loses focus.

@sameer79 , the cursor disappears only in some cases. E.g. If I select "stars" from yelp_review schema or try to select a nested function like "LTRIM(LOWER(". I tried both "single click" and selecting via arrow keys and pressing "enter", but the cursor is not shown.

For branch processor, on editing a saved entry, the sql condition text area is blank. Can you fix this?

sameer79:
@arunmahadevan @shahsank3t  @HeartSaVioR 
>  For branch processor, on editing a saved entry, the sql condition text area is blank. Can you fix this?

Fix and updated PR as well.

> the cursor disappears only in some cases. E.g. If I select "stars" from yelp_review schema or try to select a nested function like "LTRIM(LOWER(". I tried both "single click" and selecting via arrow keys and pressing "enter", but the cursor is not shown.

Looking into this.

arunmahadevan:
@sameer79 let us know once you fix the cursor issue.

@HeartSaVioR , can you do a final pass and put any additional comments?

HeartSaVioR:
@arunmahadevan Will test it by today and leave review comment.

HeartSaVioR:
Sorry adding my 2 cents, fields may be better to be shown first in suggestion rather than last.

sameer79:
@HeartSaVioR @shahsank3t  @arunmahadevan 
addressed Jungtaek’s comment and regarding cursor issue, not able to reproduce on any browser.so not sure why @arunmahadevan is facing it.

@HeartSaVioR are you facing the same cursor issue?
 which @arunmahadevan has mentioned.

PR has been updated as well.

sameer79:
@shahsank3t  @arunmahadevan @HeartSaVioR 
 Resolve conflicts and rebase the PR as well.

shahsank3t:
@arunmahadevan can you please have a final look and merge if all's good, Thanks in advance.

arunmahadevan:
@sameer79 thanks for the patch. The changes are merged to master and HDF 3.2.

@shahsank3t , can we make similar changes in the projection processor as well so that its in sync? Heres the tracking issue - https://github.com/hortonworks/streamline/issues/1217

PR #1225, opened 2018-02-22
ISSUE-1219: Added arithmetic operator support for rule/projection processor

arunmahadevan:


streams/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/rule/condition/expression/GroovyExpression.java
```diff
@@ -65,6 +65,14 @@ protected String getOperator(Operator operator) {
                     return " >= ";
                 case LESS_THAN_EQUALS_TO:
                     return " <= ";
+                case MULTIPLY:
+                    return "*";
```
HeartSaVioR: nit: May be safer to wrap it with space like above operators. Also applies DIVIDE, PLUS, MINUS.

streams/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/rule/condition/expression/StormSqlExpression.java
```diff
@@ -239,6 +239,14 @@ protected String getOperator(Operator operator) {
                     return " >= ";
                 case LESS_THAN_EQUALS_TO:
                     return " <= ";
+                case MULTIPLY:
```
HeartSaVioR: nit: May be safer to wrap it with space like above operators. Also applies DIVIDE, PLUS, MINUS.

streams/runners/storm/runtime/src/test/java/com/hortonworks/streamline/streams/runtime/storm/bolt/rules/RulesBoltConditionTest.java
```diff
@@ -161,6 +162,50 @@ public void testProjectNoCondition() throws Exception {
         };
     }
 
+    @Test
+    public void testComplex1() throws Exception {
+        doTest(readFile("/streamline-complex-condition.json"), getWeather("SFO", 0, 0));
```
HeartSaVioR: It would be nice if we could add comment on json file to see actual query. It is super hard for human to interpret such rule processor configuration to query.
I'm assuming the query as `select temperature, humidity, city from teststream where temperature + humidity > 100 OR UPPER(city) = 'SFO'` since it is added in other test.

streams/runtime/src/main/java/com/hortonworks/streamline/streams/runtime/rule/condition/expression/GroovyExpression.java
arunmahadevan: nice observation, will do it.

streams/runners/storm/runtime/src/test/java/com/hortonworks/streamline/streams/runtime/storm/bolt/rules/RulesBoltConditionTest.java
arunmahadevan: will add

arunmahadevan:
@HeartSaVioR , addressed review comments, please go through.

PR #1239, opened 2018-03-05
ISSUE-1236: UI changes to enable test environment which don't require…

joylyn:
… importing cluster

### DIFF
--- a/webservice/src/main/resources/app/scripts/containers/Environment/EnvironmentContainer.jsx
+++ b/webservice/src/main/resources/app/scripts/containers/Environment/EnvironmentContainer.jsx
@@ -664,7 +664,9 @@ class EnvironmentContainer extends Component {
                 ? <NoData imgName={"applications"} searchVal={filterValue}/>
                 : entities.length !== 0
                   ? splitData[pageIndex].map((nameSpaceList, i) => {
-                    return <EnvironmentCards key={i} nameSpaceList={nameSpaceList} nameSpaceClicked={this.nameSpaceClicked} clusterDetails={clusterDetails} refIdArr={refIdArr} loader={loader} allACL={allACL}/>;
+                    if(!nameSpaceList.namespace.internal) {
+                      return <EnvironmentCards key={i} nameSpaceList={nameSpaceList} nameSpaceClicked={this.nameSpaceClicked} clusterDetails={clusterDetails} refIdArr={refIdArr} loader={loader} allACL={allACL}/>;
+                    }
                   })
                   : !this.initialFetch && entities.length === 0
                     ? <NoData imgName={"environments"} serviceFlag={checkServices}/>
--- a/webservice/src/main/resources/app/scripts/containers/Streams/TopologyEditor/TopologyEditorContainer.jsx
+++ b/webservice/src/main/resources/app/scripts/containers/Streams/TopologyEditor/TopologyEditorContainer.jsx
@@ -223,7 +223,7 @@ class TopologyEditorContainer extends Component {
           let defaultTimeSecVal = this.getDefaultTimeSec(this.topologyConfigData);
 
           Utils.sortArray(versions, 'name', true);
-          this.graphData.nodes = TopologyUtils.syncNodeData(sourcesNode, processorsNode, sinksNode, this.graphData.metaInfo, this.sourceConfigArr, this.processorConfigArr, this.sinkConfigArr);
+          this.graphData.nodes = TopologyUtils.syncNodeData(sourcesNode, processorsNode, sinksNode, this.graphData.metaInfo, this.sourceConfigArr, this.processorConfigArr, this.sinkConfigArr, this.notifyReconfigureCallback);
 
           this.graphData.uinamesList = [];
           this.graphData.nodes.map(node => {
@@ -928,9 +928,13 @@ class TopologyEditorContainer extends Component {
     });
     this.updateGraphMethod();
     if(errorMsg.length){
-      FSReactToastr.warning(
-        <CommonNotification flag="warning" content={"Re-evaluate the configuration for the nodes marked in \"Yellow\""}/>, '', toastOpt);
+      this.notifyReconfigureCallback();
     }
+  }
+
+  notifyReconfigureCallback(){
+    FSReactToastr.warning(
+      <CommonNotification flag="warning" content={"Re-evaluate the configuration for the nodes marked in \"Yellow\""}/>, '', toastOpt);
   }
 
   showEdgeConfigModal(topologyId, versionId, newEdge, edges, callback, node, streamName, grouping, groupingFields) {
--- a/webservice/src/main/resources/app/scripts/containers/Streams/TopologyListing/AddTopology.jsx
+++ b/webservice/src/main/resources/app/scripts/containers/Streams/TopologyListing/AddTopology.jsx
@@ -35,8 +35,8 @@ class AddTopology extends Component {
   constructor(props) {
     super(props);
     this.state = {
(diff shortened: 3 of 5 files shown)


webservice/src/main/resources/app/scripts/containers/Environment/EnvironmentContainer.jsx
```diff
@@ -152,7 +152,9 @@ class EnvironmentCards extends Component {
         <div className="service-box environment-box" data-id={namespace.id} ref={(ref) => this.nameSpaceRef = ref}>
           <div className="service-head clearfix">
             <h4 className="pull-left no-margin" title={namespace.name}>{Utils.ellipses(namespace.name, 15)}</h4>
-              <div className="pull-right">
+              {namespace.internal ?
+              '' :
```
shahsank3t: @joylyn instead of the blank string set it to null.

joylyn: @shahsank3t  Thanks for the review. I have made the change.

arunmahadevan:
We need to fix some issues to make this feature more usable:

1. The "Test Environment" shows up in the Environments. I think we should hide it and show it only in the dropdown while creating an app.

2. For an app in the "Test environment", I am not able to populate the output fields for a Kafka source so not able to proceed further with the app. At-least we should allow the user to manually enter the topic name and schema version in the Kafka source so that the output fields can be populated.

3. For an app created in the "Test environment" we should make most of the fields optional for sources and allow users to save and connect the source to downstream processors as long as the output fields can be determined. For non-test environment we can retain the existing behavior.

4. I tried a HDFS source hoping that here it will allow one to manually enter the schema name, but it doesn't work. I get a "Schema not found" error after selecting the schema version.

5. When I go to the view mode for an app created in the "Test Environment", an error is thrown.
"An exception with message [Streaming Engine STORM is not associated to the namespace Test Environment(-1)] was thrown while processing request."

6. How do we expect users to configure an app for the "Test environment", test it and later associate it with some real environment ? We need to provide an option to change the "environment" of an existing app. It may need more backend work but I think we should do it before we enable the "Test Environment" feature. cc @HeartSaVioR .

Once these issues are fixed, will try out the "test mode" and provide more feedback.

@HeartSaVioR  can you review and provide any necessary backend changes to support this?

HeartSaVioR:
@arunmahadevan 
My first intention would help ourselves do simple tests easier, not making `test environment` to **full featured** thing.

2,3,4 are beyond my intention and not trivial to make it. I don't expect `test environment` will enable topology editing with sources/sinks. Some of them could be addressed with TCP source/sink.

For 6, changing environment will force users to edit sources and sinks. Unfortunately we require same work for cloning topology and it will not be done smoothly. Same as export/import topology. Unless we automatically change configuration of each sources/sinks being adapt to new environment, we may end up making the step of migration more complicated. (we may need a dialog like `migration wizard`)

If you think they're mandatory, I'd rather consider enabling `debug mode` or `test mode` for us which enables `test environment` (test environment would not be seen when such mode is off), and make sure we are just aware about its limitation of `test environment` before using. Makes sense?

arunmahadevan:
>2,3,4 are beyond my intention and not trivial to make it. I don't expect test environment will enable >topology editing with sources/sinks. Some of them could be addressed with TCP source/sink.

for 2, 3, The intention is to be able to save a source (at least in the test environment) as long as the output schema is defined so that users can build the end-end flow. I am not sure why you think it would require major changes other than UI allowing to save the source and connect it to a downstream component as long as the output stream is defined.

4 seems to be an issue with HDFS source that needs to be fixed. I tried in a real environment and UI was still not able fetch the schema.

>For 6, changing environment will force users to edit sources and sinks.

Right now users can select a different environment while cloning an app. We don't automatically fill the configs based on the new environment and most of the fields would be blank. I think we can follow similar approach while allowing to edit an app and change a test-environment to a real environment. We have a "reconfigure" flag for component, we can force set this for all the components while changing an environment so that users are forced to edit each component, review the configs and save.

HeartSaVioR:
@arunmahadevan 

> for 2, 3, The intention is to be able to save a source (at least in the test environment) as long as the output schema is defined so that users can build the end-end flow. 

It should be possible with TCP source/sink.

> I am not sure why you think it would require major changes other than UI allowing to save the source and connect it to a downstream component as long as the output stream is defined.

Because if then UI should allow sources/sinks to be partially configured. For Kafka source, we list the topics based on the environment and force users to select one to match with SR. The behavior is much helpful for non-test environment but such behavior will not work with test environment.

> 4 seems to be an issue with HDFS source that needs to be fixed. I tried in a real environment and UI was still not able fetch the schema.
> 5. When I go to the view mode for an app created in the "Test Environment", an error is thrown.
"An exception with message [Streaming Engine STORM is not associated to the namespace Test Environment(-1)] was thrown while processing request."

Yeah 4 and 5 look like a bug, but 5 is the thing I found and fixed for the patch. I'll see if I can reproduce it.

> Right now users can select a different environment while cloning an app. We don't automatically fill the configs based on the new environment and most of the fields would be blank. I think we can follow similar approach while allowing to edit an app and change a test-environment to a real environment. We have a "reconfigure" flag for component, we can force set this for all the components while changing an environment so that users are forced to edit each component, review the configs and save.

Once we are thinking about same behavior as cloning, it may be OK. Makes sense.

arunmahadevan:
@HeartSaVioR , @joylyn , the backend changes got merged to master and HDF 3.2 branch, so can you address the pending issues so that we can merge the UI changes as well ?

I think we can ignore points 2, 3 for now. Once 4 is addressed, HDFS source can be used in the test environment for testing the flow. 6 is required so that users can switch from test to real env (and manually reconfigure all the components).

Can we address all the points except 2,3 ?

shahsank3t:
@arunmahadevan 
For 6, are we implying that user will clone test topology and after that UI will use `reconfigure` flag to highlight the source/sinks so users can reconfigure them again? Or we want some other flow to move from test to real env?

arunmahadevan:
@shahsank3t , we should have a separate update API to update the environment. Internally it can set the reconfigure flag for all the components and force the user to edit each component.

@HeartSaVioR , can we provide a back-end api for this?

shahsank3t:
@arunmahadevan 
> we should have a separate update API to update the environment. Internally it can set the reconfigure flag for all the components and force the user to edit each component.

From the UI/UX perspective, if I understand it, we will add another option (say: "Update Environment") below export/clone topology, in application listing page where the user can go ahead and change the environment, is that right?

HeartSaVioR:
@arunmahadevan @joylyn @shahsank3t 
#1246 for addressing 6. When updating topology entity, it will check whether namespace id is changed or not, and set all components to `reconfigure` if the namespace id is changed.

arunmahadevan:
>From the UI/UX perspective, if I understand it, we will add another option (say: "Update Environment") below export/clone topology, in application listing page where the user can go ahead and change the environment, is that right?

@shahsank3t, you can invoke the update topology API and change the namespace id. Internally it would set reconfigure flags on all components.

arunmahadevan:
@shahsank3t , can we also hide the 'Test Environment' from getting displayed in the environment. This needs to be there only in the environment drop down while creating an app.

joylyn:
@arunmahadevan I have made the requested changes to update the namespace id of a topology and to hide the 'test environment' from the environment listing page.

arunmahadevan:
+1, LGTM

@HeartSaVioR, can you take a final look ?

HeartSaVioR:
Thanks @joylyn for your great efforts. Merged into master.

PR #1243, opened 2018-03-16
ISSUE-1240: Handle schema branches for all sources rather than just Kafka

priyank5485:


priyank5485:
@shahsank3t @joylyn 

While handling ui changes in https://github.com/hortonworks/streamline/pull/1204/files  for backend changes in https://github.com/hortonworks/streamline/pull/1200 it seems that we missed a few things.

Can we fix the below?

1. It seems that default value of MASTER for schema branch is hardcoded for field name. Quoting an example for source at https://github.com/hortonworks/streamline/pull/1204/files#diff-700814ca73bcf79d2404b6826901cebfR115 Ditto for sink. Can we change the logic in that if clause such that we use hint instead of fieldName? So, if there are any fields for that source/sink that have a hint of schemaBranch and if they dont have a value then we assign the default value of master. 
2. The logic for getting all branches for a schema also seems to be hard coded just for Kafka. https://github.com/hortonworks/streamline/pull/1204/files#diff-700814ca73bcf79d2404b6826901cebfR145 ditto for sink. That is making all other sources fail. I dont see a call being made to fetch branches for other sources like hdfs, eventhub, etc after applying json file changes in this PR. Can we change that logic instead to use the hints schemaBranch and dependsOn-<field with hint schema> to make it generic and work for all ?

In general if you guys come across any UI api calls that are hardcoded using field names, please change it to use hints as well, unless its a special case.

@arunmahadevan The changes I have pushed in this PR are necessary and will not break anything. However they are not sufficient and we need UI fixes as mentioned above as well to fix this issue.

sameer79:
Raised  PR : https://github.com/hortonworks/streamline/pull/1244

priyank5485:
@shahsank3t @sameer79  Thanks for the UI fixes. @arunmahadevan I have tested that originally reported issue for ui sending request for undefined schema is fixed. Tested with hdfs and eventhubs. Please review and merge when you get a chance along with the other PR raised by @sameer79

arunmahadevan:
+1

PR #1252, opened 2018-03-28
ISSUE-1249 Treat existence of database/user as success while initiali…

### LINKED ISSUE #1249: Treat existence of database/user as success while initializing database
Currently, when running `bootstrap-storage-initenv.sh`, it throws error when target user and/or target database exists, which may be OK to end user, but a bit bad to automation script.  Better to just skip if it exists.

HeartSaVioR:
…zing database

* Major changes were done in https://github.com/hortonworks/registry/pull/429

This closes #1249


HeartSaVioR:
This requires pulling the latest change of SR. We may need to arrange dependency of SR version as well.

satishd:
@arunmahadevan AFAIK, our RE infrastructure does not generate snapshots. You can use versions like 0.5.1.3.1.1.0-39 for now till we release final version before HDF release.

PR #1254, opened 2018-03-29
ISSUE-1253 Apply the change of service name for Ambari Infra in Ambar…

### LINKED ISSUE #1253: Apply the change of service name for Ambari Infra in Ambari 2.7
Streamline recognizes services via service name while importing Ambari cluster. 

In Ambari 2.7, the name of Ambari Infra is changed from `AMBARI_INFRA` to `AMBARI_INFRA_SOLR`. We need to reflect such change to import Ambari Infra in Ambari 2.7.

HeartSaVioR:
…i 2.7

* The name of service 'AMBARI_INFRA' is changed to 'AMBARI_INFRA_SOLR'

This closes #1253


HeartSaVioR:
NOTE: After applying patch, SAM will not recognize Ambari Infra in Ambari 2.6. I guess we don't support range of versions for specific version of SAM, so I don't think this is going to be an issue.

PR #1255, opened 2018-03-29
ISSUE-1217 Projection processor should support UDF with one or more arguments

sameer79:
@arunmahadevan  @shahsank3t @HeartSaVioR  
projection processor support UDF with multiple arguments. as well as the function as an arguments

Basic error handling has been done for projection conditions.
But in my opinion, we should have the 'API' for validating the script.

please test it on your side as well.

### DIFF
--- a/webservice/src/main/resources/app/scripts/components/CommonCodeMirror.jsx
+++ b/webservice/src/main/resources/app/scripts/components/CommonCodeMirror.jsx
@@ -49,7 +49,9 @@ export default class CommonCodeMirror extends Component{
         lineNumbers: true,
         mode: "javascript",
         styleActiveLine: true,
-        lint: true
+        lint: true,
+        matchBrackets : true,
+        extraKeys : {"'@'": "autocomplete"}
       },
       json : {
         lineNumbers: true,
@@ -74,7 +76,7 @@ export default class CommonCodeMirror extends Component{
   }
 
   componentWillReceiveProps(newProps) {
-    if (this.props.editMode && this.initialFlag) {
+    if (newProps.editMode && this.initialFlag) {
       this.setState({textValue : newProps.value}, () => {
         this.codeWrapper.setValue(newProps.value);
         this.initialFlag= false;
@@ -186,11 +188,11 @@ export default class CommonCodeMirror extends Component{
   }
 
   getNestedFieldHints = (string) => {
-    const {hintOptions} = this.props;
+    const {hintOptions,modeType} = this.props;
     let arr=[];
     const nestedFunction = (hints) => {
       _.map(hints, (h) => {
-        if(!!string && string.includes(h.displayText) && h.fields){
+        if(!!string && string.includes(h.filterText || h.displayText) && h.fields){
           arr = this.filterByName(h.fields, string,'nested');
         }
         if(h.fields){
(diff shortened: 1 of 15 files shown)


sameer79:
screenshot below: 

prejection-changes



projection-error

arunmahadevan:
@sameer79 , a few comments

1. Change the title "Projection Conditions" to "Projection Expression"
2. Change the title "Fields Name" to "Output field name"
3. Instead of "Conditions goes here..." it should be "Expression goes here..."
4. The pop up displays only functions. We should also display the fields so that we can write expressions like f1 + f2 in the box. So we should also display arithmetic operators (+, -, /, *).
5. I selected the 3 arg version of "Substring", but after typing 2 arguments and comma, an error message "Function doesn't support more arguments" is displayed. I think you need to consider the overloaded functions before throwing the error.

Following are some examples of expressions that should be supported :

Expression | Output field name
----------- | -------------------
TRIM(drivername) | dn
x + y | f1
10/2 | five
(x + y) * n | f2
input_field3   | f3
SQRT(x + y) | sq


@HeartSaVioR , can you also take a look and provide feedback ?

sameer79:
@shahsank3t  @arunmahadevan  Yes  we expect API to return like
````
"projections":[
{"expr":"IDENTITY(truckId) AS t"},
{"expr":"TRIM(driverName) AS dn"},
{"expr":"eventTime"}]
````
That's the only backend change needed.

Coming to the above-mentioned points, we can handle points [1,2,3,5].
Need more clarity on point 4:

No | Expression | Output field name | ReturnType
------ |  ----------- | ------------------- | -----------------------------------------
1 | TRIM(drivername) | dn | TRIM
2 | x + y | f1 | ?
3 | 10/2 | five | ?
4 | (x + y) * n | f2 |  ?
5 | input_field3   | f3 | input_fields3
6 | SQRT(x + y) | sq | SQRT

In projection, we use the return type to show the outputStream like.
````
{
  name : "dn",
  type : function returnType OR arguments type
}
````
From the expression given in above table:
Row 2]  x+y  "INTEGER + DOUBLE" 
Row 3]  10/2  "INTEGER / INTEGER"  
Row 4]  (x + y) * n "(INTEGER + DOUBLE)*DOUBLE 

Row [2,3,4]  may return FLOAT OR DOUBLE. if the user used the higher number for expression.

Basically, need to understand how to identify what return type to be assigned in those case.

sameer79:
@arunmahadevan  @shahsank3t Address the comments, which includes the mentioned-points [1,2,3,5] and updated the PR as well.

arunmahadevan:
@sameer79 , you could use the Java type conversion rules to determine the type of the arithmetic expression.

If either operand is of type double, the other is converted to double.
Otherwise, if either operand is of type float, the other is converted to float.
Otherwise, if either operand is of type long, the other is converted to long.
Otherwise, both operands are converted to type int.

Reference - https://docs.oracle.com/javase/specs/jls/se8/html/jls-5.html#jls-5.6.2

Treat numeric literals as integer or double (if it has a decimal point)

sameer79:
@arunmahadevan @shahsank3t  @HeartSaVioR Address the comments and updated the PR as well.

arunmahadevan:
@sameer79 , when I select fields with underscore (e.g. projection expression: review_id , field name: f1), the output field type is not populated and I cannot save the details. If the field does not have an underscore or if the field is part of a function, it works.

When I pass uppercase "AS from the backend APIs, the UI is not showing up. Can you change the "As" to upper case "AS" in the expression (e.g. "expr": "TRIM(name) AS f1" instead of "expr": "TRIM(name As f1") and while handling it in the UI.

arunmahadevan:
@sameer79 , made some changes to get the existing projections work with the new format - https://github.com/hortonworks/streamline/pull/1262. This requires the UI to interpret "AS" (uppercase)

sameer79:
@arunmahadevan  address the comment and updated the PR as well.

arunmahadevan:
@sameer79 , thanks LGTM. +1

@HeartSaVioR , can you try it and see if it looks good?

PR #1261, opened 2018-04-04
ISSUE-1260: Changes to discover atlas and enable storm hook to notify

### LINKED ISSUE #1260: Streamline should discover Atlas and notify during topology deployment

arunmahadevan:
Discover and enable Storm Atlas notifier plugin if user has selected 'Atlas' in the environment.

This patch does not handle shipping the necessary atlas jars in the storm path. This can be handled in a separate patch by optionally shipping relevant atlas jars via streamline or have Ambari/Atlas to install the storm atlas hook jars in the HDF cluster.

Closes #1260


arunmahadevan:
@HeartSaVioR , manually placing the required atlas jars into storm's extlib works. Its not clear how these jars would be shipped to the HDF storm installation. I can create a patch to bundle the required atlas jars within streamline (using a separate build profile) and create symlinks from storm. I am also checking with Atlas team if they can provide a package for this instead of adding the jars in streamline distribution.

PR #1269, opened 2018-05-03
ISSUE-1268 Provide script to help download all transitive dependencies

### LINKED ISSUE #1268: Provide script to help download all transitive dependencies
We got some feedbacks for running Streamline in restricted environment, specific to restricted internet access.

Streamline relies on Aether (leveraging Storm's feature) to pull the transitive dependencies for the app, which requires access to the (official/HWX) Maven repositories. When the node which Streamline is running can't access to repositories, topology can't be submitted properly.

It can be mitigated via leveraging local cache of Maven: when we provide a way to download all transitive dependencies from another node, users can download them and copy to the node which is restricted to access the internet.

Downside of the approach is clear: users should do the steps for every updates on dependency. We don't announce the change of dependencies, so it is effectively all the updates.

HeartSaVioR:
* provide utility script to download all transitive dependencies
  * requires
    * running Streamline instance
    * binary distribution of Apache Storm

This closes #1268.

### DIFF
--- a/bootstrap/download-streamine-artifacts.py
+++ b/bootstrap/download-streamine-artifacts.py
@@ -23,162 +23,162 @@ import uuid
 # below modules are not built-in, so you need to install via pip or easy_install or so
 # recommend to create virtualenv and install with activating env.
 try:
-	import pureyaml
+    import pureyaml
 except ImportError:
-	print("ERROR: 'pureyaml' doesn't seem to be installed. Please ensure the module is installed.")
-	print("You can install 'pureyaml' via 'pip install pureyaml'.")
-	sys.exit(1)
+    print("ERROR: 'pureyaml' doesn't seem to be installed. Please ensure the module is installed.")
+    print("You can install 'pureyaml' via 'pip install pureyaml'.")
+    sys.exit(1)
 
 COMPONENT_BUNDLES_KIND_URL_POSTFIX = "/streams/componentbundles"
 
 
 class ConfigStruct:
-	def __init__(self, catalog_root_url, storm_home_dir, maven_repo_url, maven_local_repository_directory):
-		self.catalog_root_url = catalog_root_url
-		self.storm_home_dir = storm_home_dir
-		self.maven_repo_url = maven_repo_url
-		self.maven_local_repository_directory = maven_local_repository_directory
-		self.http_proxy_url = None
-		self.http_proxy_username = None
-		self.http_proxy_password = None
+    def __init__(self, catalog_root_url, storm_home_dir, maven_repo_url, maven_local_repository_directory):
+        self.catalog_root_url = catalog_root_url
+        self.storm_home_dir = storm_home_dir
+        self.maven_repo_url = maven_repo_url
+        self.maven_local_repository_directory = maven_local_repository_directory
+        self.http_proxy_url = None
+        self.http_proxy_username = None
+        self.http_proxy_password = None
 
-	def set_http_proxy(self, proxy_url, proxy_username = None, proxy_password = None):
-		self.http_proxy_url = proxy_url
-		self.http_proxy_username = proxy_username
-		self.http_proxy_password = proxy_password
+    def set_http_proxy(self, proxy_url, proxy_username = None, proxy_password = None):
+        self.http_proxy_url = proxy_url
+        self.http_proxy_username = proxy_username
+        self.http_proxy_password = proxy_password
 
-	def get_storm_submit_tool_classpath(self):
-		storm_submit_tool_dir = os.path.join(self.storm_home_dir, "external", "storm-submit-tools")
-		storm_submit_tool_classpath = os.path.join(storm_submit_tool_dir, "*")
-		return storm_submit_tool_classpath
+    def get_storm_submit_tool_classpath(self):
+        storm_submit_tool_dir = os.path.join(self.storm_home_dir, "external", "storm-submit-tools")
+        storm_submit_tool_classpath = os.path.join(storm_submit_tool_dir, "*")
+        return storm_submit_tool_classpath
 
-	def get_component_bundles_kind_url(self):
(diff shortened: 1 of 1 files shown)


bootstrap/download-streamine-artifacts.py
```diff
@@ -0,0 +1,178 @@
...
+import subprocess
+import copy
+
+# below modules are not built-in, so you need to install via pip or easy_install or so
+# recommend to create virtualenv and install with activating env.
+try:
+	import pureyaml
+except ImportError:
+	print("ERROR: 'pureyaml' doesn't seem to be installed. Please ensure the module is installed.")
+	print("You can install 'pureyaml' via 'pip pureyaml'.")
```
arunmahadevan: pip pureyaml -> pip install pureyaml

bootstrap/download-streamine-artifacts.py
```diff
@@ -0,0 +1,178 @@
...
+	
+	print("============== Writing downloaded information to file: %s" % output_file)
+	write_downloaded_information_to_output_file(output_file, downloaded_artifacts)
+
+	print("DONE!")
+
+	print("Please follow below steps:")
+	print("1. archive directory of either '.m2/repository' in home directory, or 'local-repo' in current directory based on location for downloaded artifacts.")
+	print("2. send archived file to the machines which streamline is installed and don't have access to internet.")
+	print("3. extract file to '.m2/repository' in home directory of streamline account. Please make the directory if it doesn't exist.")
```
arunmahadevan: Sometimes the artifacts are in local_repo directory. Does it always fall back to .m2/repository? How about the jar permission and ownership - will it have required permissions for the user streamline is running as ?

bootstrap/download-streamine-artifacts.py
HeartSaVioR: Nice catch! Will fix.

bootstrap/download-streamine-artifacts.py
HeartSaVioR: | .m2/repository | local-repo | download location |
| X | X | local-repo |
| O | X | .m2/repository |
| X | O | local-repo |
| O | O | .m2/repository |

So whenever '.m2/repository' directory exists, the directory is used. 'local-repo' is used for failing back.

For permission and ownership, I guess it belongs to the account which runs the script. So simplest approach is run this script as the account which runs streamline. Maybe just ensuring  '.m2/repository' exists should make it simpler. 

One thing we need to consider is, downloaded artifacts will be archived and copied to the other machine and extracted, so final ownership of archives will belong to the account which extracts the archive.

Could you suggest how to document this? I'm not sure how much the script should guide to users kindly.

HeartSaVioR:
Updated the patch: now it forces to download artifacts to 'local-repo' directory in cwd. Also updated the description.

bootstrap/download-streamine-artifacts.py
```diff
@@ -0,0 +1,183 @@
...
+	
+	print("============== Writing downloaded information to file: %s" % output_file)
+	write_downloaded_information_to_output_file(output_file, downloaded_artifacts)
+
+	print("DONE!")
+
+	print("Please follow below steps:")
+	print("1. archive 'local-repo' in current directory which downloaded artifacts are placed.")
+	print("2. send archived file to the machines which streamline is installed and don't have access to internet.")
+	print("3. extract file to 'local-repo' in root directory of streamline installation.")
```
arunmahadevan: @HeartSaVioR , In some installations I have seen the local_repo getting created under /tmp (e.g. /tmp/hsperfdata_streamline/local-repo) . 

Does aether create/looks for the local-repo directory under the process CWD by default? 

If so we should
1. Create an empty "local-repo" directory under "base dir" as part of streamline distribution with "rwx" for others.
2. Change the CWD of streamline to "base dir" in the launch script (E.g. `(cd $base_dir && exec ....)` )
3. Now users can place the jars under local-repo so that aether picks it up correctly.

HeartSaVioR: @arunmahadevan 

> In some installations I have seen the local_repo getting created under /tmp (e.g. /tmp/hsperfdata_streamline/local-repo).
> Does aether create/looks for the local-repo directory under the process CWD by default?

Sounds weird. Not sure why it is created under such directory, because it should use CWD of where it launches storm-submit-tools.

```
            String localMavenRepoPath = getOrDefaultLocalMavenRepositoryPath("local-repo");

            // create root directory if not exist
            Files.createDirectories(new File(localMavenRepoPath).toPath());
```

Note that we should consider which account runs `storm-submit-tool`, and which CWD is it. Since it is executed as a part of `storm submit`, it is affected how `storm` and `storm.py` modifies its CWD, so the location might be probably what we don't expect.

I'm avoiding to fix Storm stuff, but looks like I need to fix storm-submit-tool as well as storm.py script to receive custom local repository directory, so the local repository directory is decided as such priorities:

1. custom local repository directory if given
2. `.m2/repository` in account home directory if exists
3. CWD/local-repo

which requires both fixes of Storm and Streamline. This script can be updated to receive custom local repository directory afterwards, which users should know where is it.

HeartSaVioR:
Now the script receives maven local repository directory (depends on STORM-3069) and downloads transitive dependencies to given directory.

There's style nit (using tab instead of spaces) which I'll fix it after getting +1.

@arunmahadevan Could you review this again? Thanks in advance.

arunmahadevan:
+1 LGTM. Should we wait till the storm patch is merged to merge this ?

HeartSaVioR:
@arunmahadevan 
Yes and no, since the script is not mandatory thing and only used for end users who needs this.

HeartSaVioR:
Fixed indentation : tab -> 4 spaces.

HeartSaVioR:
Storm patch is merged so I guess we are good to go. I'll just take +1 on @arunmahadevan since I just fixed indentation via replacing `\t` -> `    `, not fixing others. Merging...

PR #1276, opened 2018-05-11
ISSUE-1275 Allow users to specify maven local repository directory for Storm topology submission

### LINKED ISSUE #1275: Allow users to specify maven local repository directory for Storm topology submission
This is needed in prior to resolve #1268 because current approach of finding maven local repository directory is sometimes non-deterministic (basically relying on CWD but sometimes we have seen odd path), so even we provide script to let users download transitive dependencies, users might have hard time to find where to copy and place them.

We should let SAM allow maven local repository directory as an option to streamline.yaml, and pass it as additional option while submitting topology via "storm submit".

HeartSaVioR:
* add new configuration in streams module: "mavenLocalRepositoryDirectory"
  * if it is presented, it is passed to "storm submit"
  * so that transitive dependencies are read and fetched from given directory

The new configuration only works with Storm version which supports custom maven local repository directory (STORM-3069), so please check it first before setting the configuration. 

This closes #1275


arunmahadevan:
+1

PR #1280, opened 2018-05-23
ISSUE-1279: Fix null handling in CustomProcessorBolt

arunmahadevan:
Will raise a separate one for master.


HeartSaVioR:
It would be nice to have test if we can spend short time on craft it. If it doesn't look like, forget about it.

arunmahadevan:
I had started it, will try to complete it if I can.

arunmahadevan:
@HeartSaVioR , added tests.

PR #1285, opened 2018-05-31
ISSUE-1284: Handle the case where the Ambari storm view URL query fails

arunmahadevan:


arunmahadevan:
@HeartSaVioR, can you take a look?

PR #1289, opened 2018-06-07
ISSUE-1288 SAM aggregate processor doesn't show the aggregation function

sameer79:


sameer79:
Aggregate Processor screenshot below:-

aggregate-screenshot

arunmahadevan:
@sameer79 , the changes are good. However we need to make sure that user passes only aggregate expressions (the truckId in the screenshot that you have uploaded is invalid). Right now it would fail at runtime so I have added some backend validations https://github.com/hortonworks/streamline/pull/1290

If possible can you also disable such expressions from the UI ?

sameer79:
The users will be forced to select a function before choosing the field and if a user just passes only a field the validation has been added to handle that case too.
Help has been added to bring up the pop up (like pressing @ in other processors)

@arunmahadevan  can you please verify the same, thanks in advanced

arunmahadevan:
+1

Issue #1293, opened 2018-06-14
Configured projection processor shows as blank


shahsank3t:
After saving the projection processor, when we open it again, the configuration is blank.

This issue is introduced due to the changes in https://github.com/hortonworks/streamline/issues/1288

cc: @arunmahadevan

PR #1294, opened 2018-06-14
ISSUE-1293 Configured projection processor shows as blank

shahsank3t:
This PR fixes https://github.com/hortonworks/streamline/issues/1293


shahsank3t:
@arunmahadevan this fix is required to be checked-in since it breaks the projection processor completely.

arunmahadevan:
+1

PR #1299, opened 2018-06-29
BUG-106423: HDF Streamline 'custom-udf-microtest' modules has different groupId than its parent module

tibkiss:


tibkiss:
Should be cherry-pick to HDF-3.2.0.0 branch too.

PR #1300, opened 2018-07-08
Registry version is updated to 0.5.2

satishd:


HeartSaVioR:
@satishd 
Could you explain how did you test with the change? Registry doesn't seem to respect semantic versioning, so manual test would be needed even for bugfix version upgrade.

satishd:
@HeartSaVioR I have not tested manually. If it is really needed, one of you can do that before merging this PR.

cc @arunmahadevan

HeartSaVioR:
@satishd Thanks for raising quick follow-up PR on registry change. Merged into master branch.

PR #1309, opened 2018-08-06
ISSUE-1308 Update registry version with 0.5.3

### LINKED ISSUE #1308: Update registry version to 0.5.3

satishd:
- Updated registry version with 0.5.3
- Excluded jsp-api dependencies to handle conflicts with dropwizard dependencies.
- Corrected repo url in root pom.xml

Closes #1308

### DIFF
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -78,6 +78,10 @@
                     <groupId>javax.servlet</groupId>
                     <artifactId>servlet-api</artifactId>
                 </exclusion>
+                <exclusion>
+                    <groupId>javax.servlet.jsp</groupId>
+                    <artifactId>jsp-api</artifactId>
+                </exclusion>
             </exclusions>
         </dependency>
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -12,7 +12,7 @@
     <description>
       Develop and deploy Streaming Analytics applications visually with bindings for streaming engine and multiple source/sinks, rich set of streaming operators and operational lifecycle management. Streaming Analytics Manager makes it easy to develop, monitor streaming applications and also provides analytics of data thats being processed by streaming application.
     </description>
-    <url>http://github.com/hortonworks/registry</url>
+    <url>http://github.com/hortonworks/streamline</url>
 
     <licenses>
         <license>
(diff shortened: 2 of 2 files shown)


HeartSaVioR:
I wouldn't upgrade unless one or both conditions are met:

1. The new version of SR introduces breaking change, and only the way for Streamline to be compatible with such version is upgrading.
(I'd be concerned if the breaking change is introduced in minor/bugfix version: this means SR requires all clients to make changes as well.)

2. Security issue identified in old version and the new version fixes the issue. (Putting CVE number or relevant issue would be nice to decide whether we need to upgrade or not)

If it meets the condition, please describe it to the body of PR so that we can determine why we should upgrade.

HeartSaVioR:
@satishd 
And I would say this is a good time to decouple storage module from Registry and have a new project. I hope to decouple anything once they are not relevant, so that we could upgrade the things which are only really needed, and we could have smaller areas to test once we upgrade.

satishd:
@HeartSaVioR There are no breaking changes but few bug fixes that went in with 0.5.3 as mentioned [here](https://github.com/hortonworks/registry/releases/tag/0.5.3).

Security vulnerabilities that were addressed in this release are
- CVE-2017-7536
- CVE-2017-1000487

This release also includes [commit](https://github.com/hortonworks/registry/commit/4a657d50b229526f6f5fa6c61abf04496e07a7e5) which came from https://github.com/hortonworks/registry/pull/464, which was reviewed by you.

HeartSaVioR:
@satishd 
What I've reviewed and verified is for storage module. As I skimmed release notes in SR 0.5.3, look like the things streamline should reflect are in storage module, unless CVE-2017-7536 affects SR client.

If SR client is not affected from CVE, I'd upgrade the version for only storage module and only verify the change via playing with bootstrap things. This would avoid non-trivial efforts like the end-to-end step for testing: install new SR, setup streamline and reflect new SR, set up end-to-end test environment, and test . This is my preferred approach, minimizing the impact on upgrading. 

If SR client should be upgraded as well (like affected to CVE) I'm happy to verify upgrade, though I wish that is also described in release note, so that other end users (maybe users for OSP version) of SR can indicate the issue.

cc. @arunmahadevan

satishd:
@HeartSaVioR You can close this PR and proceed with whatever approach is convenient to you. But you may need to be careful in having SR common module like common-auth are working with different upstream module versions.

HeartSaVioR:
@satishd 
So I sought commit list of common, common-auth, common-client and found CVE got resolved from there. I can't find this information until I sought the commit list. My bad I should be aware earlier, but I'm not sure it means I should need to also follow and watch SR.
(To be honest I'm watching SR but most of the things are not relevant of streamline so just adding clicks in inbox and miss such informational thing. I'm going to unwatch for now, and then I would just rely on release note.)

If they're extracted out of common project, I would be happy to watch and participate in (I'm not only an user of common project: I'm the author of storage-tool) because I would be less notified by Github as well as the chance of making modification of the project is small. It would also help to share the  responsibility of maintenance for common modules. Does it make sense to you?

PR #1312, opened 2018-09-04
ISSUE-1310 Unable to save projection when using select all Projection Fields

sameer79:
@arunmahadevan @shahsank3t  ready for review comment.

### DIFF
--- a/webservice/src/main/resources/app/scripts/utils/ProcessorUtils.js
+++ b/webservice/src/main/resources/app/scripts/utils/ProcessorUtils.js
@@ -720,14 +720,113 @@ const codeMirrorOptionsTemplate = (el,data) => {
 
 const webWorkerValidator = function(fieldsHintArr, functionListArr) {
   return `data:text/javascript;charset=US-ASCII,
-    let funcs = ${JSON.stringify(functionListArr)};
-    let arg = ${JSON.stringify(fieldsHintArr)};
-
     self.onmessage = function(msg) {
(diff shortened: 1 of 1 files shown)


arunmahadevan:
@sameer79 
Even with the fix, I am not able to save unless I select something in the "Projection Expression". This should not be mandatory and one should be able to save just by selecting one or more "Projection fields". Can you check?

arunmahadevan:
Actually both "Projection Expression" and "Projection Fields" should not be mandatory. We should be able to save as long as there is at-least one value in either "Projection Expression" or "Projection Fields".

shahsank3t:
@sameer79 - I believe you replaced `let` & `const` at some places with `var` keyword. Any specific reason not to change it everywhere?

shahsank3t:
@arunmahadevan - Changes you mentioned in the comment above are kind of UI enhancement which needs to be implemented. 
My suggestion is to create another Github issue for that enhancement and track from there and just verify the current issue here.

sameer79:
@shahsank3t address the comment and updated the PR as well.

arunmahadevan:
+1

@sameer79 @shahsank3t able to save projections with a subset of or "select all" fields in chrome and Firefox. We can address the issue of making the "Projection Expression" optional in a separate issue.

arunmahadevan:
Raised https://github.com/hortonworks/streamline/issues/1314 for followup.

PR #1315, opened 2018-09-05
ISSUE-1313 Not able to add rule in Rule processor if we use alphanumeric column name

shahsank3t:
@sameer79 - Was there any particular need for adding 0-9 into this regex? Seems like that's causing the issue. Could you please clarify if it's good to remove the number or is needed? Thanks.

@arunmahadevan - I have tested with the basic query and works fine for me now. Could you please try some complex query that can be run by the user so we can be sure of this fix? Thanks.


arunmahadevan:
Will wait for comments for @sameer79 before merging.

sameer79:
@arunmahadevan @shahsank3t As far as adding 0-9 into this regex is the concern. It was put to restrict the alphanumeric. so if the use case needs the alphanumeric values we can get rid of it.

Issue #1314, opened 2018-09-05
Make projection fields and expression optional in projection processor


arunmahadevan:
Right now "Projection Expression" is mandatory. Ideally all of the below cases should work.

Zero or more "Projection Fields"
Zero or more "Projection Expression"

But there must be
1. at-least one output field (via the fields or expression) AND
2. no duplicate output fields (by specifying same field name in projection expression)

PR #1318, opened 2018-09-07
ISSUE-1314: Make projection fields and expression optional in projection processor

joylyn:
Changes for #1314


joylyn:
@arunmahadevan 

> Not able to save with only "Projection Expressions"

I have now fixed this issue and updated the PR.

PR #1321, opened 2018-10-12
ISSUE-1320 add new Kafka consumer properties

bpgergo:


arunmahadevan:
Are all these properties valid for kafka client version - 0.10.2.1 (mentioned in the component json).

The current way of having to make code changes to add new properties does not scale. Also some properties may be applicable only to specific kafka versions. I am not sure if theres a way to add user defined properties (key-values) for a component. If not its better to add that.

PR #1324, opened 2018-10-18
ISSUE-1323: Add filter to set custom HTTP response headers

arunmahadevan:


webservice/src/main/java/com/hortonworks/streamline/webservice/StreamlineApplication.java
```diff
@@ -135,6 +135,25 @@ private void addServletFilters(StreamlineConfiguration configuration, Environmen
...
+        FilterRegistration.Dynamic dynamic = environment.servlets().addFilter(name, filter);
+        dynamic.setInitParameters(params);
+        dynamic.addMappingForUrlPatterns(EnumSet.allOf(DispatcherType.class), true, "/*");
+        LOG.info("Added servlet filter '{}' with configuration {}", filter, params);
+    }
+
+    // Add security headers. If needed these may be overridden via custom 'servletFilters'.
+    private void addSecurityHeaders(Environment environment) {
+        Map<String, String> params = new HashMap<>();
+        params.put("X-Frame-Options", "SAMEORIGIN");
```
HeartSaVioR: Would we be better to have configuration for these parameters and apply the values on config?

arunmahadevan: These can be overridden via config, but wanted to add values for basic security headers by default even if user forgets to add (and right now theres no way for users to add this via Ambari).

HeartSaVioR: I'm pretty sure it's better to add, but also wondering some customers wanting to tune or even disable. Would we also address this in Ambari side as well? Anyway +1 on streamline itself.

arunmahadevan: The headers can be tuned via configs in streamline.yaml. (have added commented samples to the yaml files). The hard-coded ones are standard security headers and should not be disabled. The values can be changed via the yaml. Will followup on exposing it in Ambari.

HernanHortonworks: @arunmahadevan 

Do you have the required changes on  streamline.yaml to bypass this?. Your help is highly appreciated here.

```
X-XSS-Protection HTTP Header missing on port 7778. 
X-Content-Type-Options HTTP Header missing on port 7778. 
Content-Security-Policy HTTP Header missing on port 7778. 
```

arunmahadevan: This patch adds these headers by default, no changes are required in streamline.yaml.

PR #1329, opened 2018-11-28
Support numeric user id

arunmahadevan:


arunmahadevan:
@HeartSaVioR thanks for reviewing. You are right, it requires the registry side changes as well.

PR #1336, opened 2019-02-12
MINOR : Update registry version and fix transaction isolation enum

raju-saravanan:
Updated registry version to "0.5.4.3.4.0.0-96" to resolve HDF build issues, will update once again after we are done with registry release. (which should happen in couple of days)


raju-saravanan:
Raised a PR (https://github.com/hortonworks/streamline/pull/1336) to fix this.

satishd:
thanks @raju-saravanan for the PR.
I could build successfully. You can change the version to 0.6.0 once it is released, which will be done in the next few days.

PR #1337, opened 2019-02-27
Updated registry version with 0.6.0

satishd:
Updated registry version with 0.6.0

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -164,7 +164,7 @@
         <phoenix.version>4.7.0.2.5.0.0-1245</phoenix.version>
         <redis.lettuce.version>3.4.2.Final</redis.lettuce.version>
         <scala.version>2.10.5</scala.version>
-        <schema-registry.version>0.5.3</schema-registry.version>
+        <schema-registry.version>0.6.0</schema-registry.version>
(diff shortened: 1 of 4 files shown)


satishd:
Once this is merged, it will be cherry picked into HDF-3.4.0.0

HeartSaVioR:
Otherwise I had to ask screenshot to verify it passes manual tests. I know it's also not easy one to SR developers, but either of one should be proceeded to apply this - if not, sadly we should blindly merge, worst situation at all.

satishd:
Avro version is not changed in registry since https://github.com/hortonworks/registry/commit/51e85480f1c1464ae718a29c90d52119173f2609 and it is the same version in earlier dependent version viz 0.5.3. It seems some hadoop related dependencies may have pulled other avro libs.

Agree that shaded client jar may be helpful, which can be added in next registry version. When are streamline integration tests run after HDF-3.4.0.0?

satishd:
@arunmahadevan  `TransactionIsolation.DEFAULT` is renamed to  `TransactionIsolation.JDBC_DEFAULT`. So, there are no semantic changes and no need to do any testing for that.

PR #1339, opened 2019-03-04
ISSUE-1338 Exclude Avro dependency from all possible client paths to not conflict with SR's Avro version

HeartSaVioR:
SR depends on Avro 1.8 whereas dependencies in Hadoop ecosystem depend on Avro 1.7 hence conflicted.
This patch fixes the issue to try best to leverage SR's Avro version in runtime topology.

This depends on https://github.com/hortonworks/registry/pull/541 - so we need to update SR's version here in the patch. I just didn't do this because it requires new release of SR.

### DIFF
--- a/bootstrap/components/sinks/hdfs-sink-topology-component.json
+++ b/bootstrap/components/sinks/hdfs-sink-topology-component.json
@@ -6,7 +6,7 @@
   "streamingEngine": "STORM",
   "fieldHintProviderClass": "com.hortonworks.streamline.streams.cluster.bundle.impl.HDFSBundleHintProvider",
   "transformationClass": "com.hortonworks.streamline.streams.layout.storm.HdfsBoltFluxComponent",
-  "mavenDeps": "org.apache.storm:storm-hdfs:STORM_VERSION^org.slf4j:slf4j-log4j12^org.apache.curator:curator-client^org.apache.curator:curator-framework",
+  "mavenDeps": "org.apache.storm:storm-hdfs:STORM_VERSION^org.slf4j:slf4j-log4j12^org.apache.curator:curator-client^org.apache.curator:curator-framework^org.apache.avro:avro",
   "topologyComponentUISpecification": {
     "fields": [
       {
--- a/pom.xml
+++ b/pom.xml
@@ -164,7 +164,8 @@
         <phoenix.version>4.7.0.2.5.0.0-1245</phoenix.version>
         <redis.lettuce.version>3.4.2.Final</redis.lettuce.version>
         <scala.version>2.10.5</scala.version>
-        <schema-registry.version>0.6.0</schema-registry.version>
+        <!-- FIXME: MUST BE FIXED WITH THE VERSION WHICH DOES NOT CONTAIN AVRO 1.7.7 -->
+        <schema-registry.version>0.6.1-SNAPSHOT-ISSUE-540</schema-registry.version>
         <slf4j.version>1.7.12</slf4j.version>
         <storm.version>1.2.2</storm.version>
         <woodstox-core-asl.version>4.4.1</woodstox-core-asl.version>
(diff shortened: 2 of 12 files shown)


HeartSaVioR:
cc. @satishd @raju-saravanan Please also refer here. I think this is clearest way for both SAM and SR.

bootstrap/components/sinks/hdfs-sink-topology-component.json
```diff
@@ -6,7 +6,7 @@
   "streamingEngine": "STORM",
   "fieldHintProviderClass": "com.hortonworks.streamline.streams.cluster.bundle.impl.HDFSBundleHintProvider",
   "transformationClass": "com.hortonworks.streamline.streams.layout.storm.HdfsBoltFluxComponent",
-  "mavenDeps": "org.apache.storm:storm-hdfs:STORM_VERSION^org.slf4j:slf4j-log4j12^org.apache.curator:curator-client^org.apache.curator:curator-framework",
```
arunmahadevan: Do we need to exclude from the hive and hbase source and sink as well ?

streams/catalog/pom.xml
```diff
@@ -53,14 +53,6 @@
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
         </dependency>
-        <dependency>
```
arunmahadevan: Catalog deps may not be shipped to storm runtime so is this required ?. If so, do we need to exclude hive-metastore from streamline-cluster as well? I am not sure if it would affect the functionality of SAM to look up hive tables.

HeartSaVioR: Even in webservice module it might make conflict between SR. Need to check dependency tree again. Btw, even it conflicts, maybe we have to ship hive-exec as 'core' classifier instead of just removing it, and exclude Avro there. Will check.

bootstrap/components/sinks/hdfs-sink-topology-component.json
HeartSaVioR: Actually we can't exclude Avro here since hive-exec "contains" Avro as uber jar. If we want to extract Avro as well we should address submit-tool in Storm to handle classifier as well, and add `hive-exec` as `core` classifier and exclude Avro from there.

HeartSaVioR: I've excluded Avro from HBase Sink as well. For Hive sink we still need to work on Storm submit-tool.

pom.xml
```diff
@@ -164,6 +164,7 @@
         <phoenix.version>4.7.0.2.5.0.0-1245</phoenix.version>
         <redis.lettuce.version>3.4.2.Final</redis.lettuce.version>
         <scala.version>2.10.5</scala.version>
+        <!-- FIXME: MUST BE FIXED WITH THE VERSION WHICH DOES NOT CONTAIN AVRO 1.7.7 -->
```
HeartSaVioR: Reminder: We shouldn't miss this before merging.

HeartSaVioR:
After the fix I only see Avro 1.8 in whole dependency tree for SAM. (For sure, it depends on hortonworks/registry#541.)

bootstrap/components/sinks/hdfs-sink-topology-component.json
HeartSaVioR: Never mind. storm-submit-tool is hopefully flexible enough to handle classifier. Addressed in 45e3fa46d729d37447887bf63dcafe7ae6b21f93

satishd:
@HeartSaVioR @arunmahadevan  I guess this PR can be merged as the related PRs in the registry are merged and waiting for https://github.com/hortonworks/registry/pull/544 to be merged.

arunmahadevan:
I guess we need to update the new registry version with https://github.com/hortonworks/registry/pull/541 in the pom.xml before merging this.

@satishd, do you plan to release a new registry version soon?

HeartSaVioR:
@arunmahadevan I just updated the SR version and reflected the change I brought in SR. Could you please take final review? Thanks!

arunmahadevan:
Merged to master.

I am unable to merge the master to HDF-3.4.0.0 branch. It seems that branch has been created out of an older branch (HDF 3.2 ?) while there are multiple fixes in master after that. I think we should merge the latest master with HDF 3.4.

arunmahadevan:
I reset HDF-3.4 to the last merge commit (67263f1a0230f1849593595284271a17ca750eb7) and then merged latest master into it.

PR #1341, opened 2019-03-07
[WIP] ISSUE-1340 Upgrade jackson and relevant dependencies (directly and indirectly)

HeartSaVioR:
This patch brings change of huge dependencies so should be tested thoughtfully.

The build passed with custom build of SR, and I'll upgrade once the version is available in HWX repo. After that we can check whether the build passes in Travis CI.


HeartSaVioR:
Since it touches various dependencies we may not feel safe to just rely on unit test. We may need to step back and not rush it...

Which tests should we perform before merge this? If we feel safe to rely on system tests we need to merge so we have contradiction here.

pom.xml
```diff
@@ -164,15 +165,16 @@
         <phoenix.version>4.7.0.2.5.0.0-1245</phoenix.version>
         <redis.lettuce.version>3.4.2.Final</redis.lettuce.version>
         <scala.version>2.10.5</scala.version>
-        <!-- Ideally we need to use official version of SR: wait for new release -->
-        <schema-registry.version>0.6.0.3.4.0.0-145</schema-registry.version>
+        <!-- This version should be changed to actual HDF-3.4.0.0-146 version when it is published -->
+        <schema-registry.version>0.6.1-SNAPSHOT-HDF-3.4.0.0-146-maybe</schema-registry.version>
```
HeartSaVioR: Should be changed to 3.4.0.0-146 (or higher) or new official SR version.

pom.xml
```diff
@@ -146,16 +146,17 @@
         <commons-lang.version>2.6</commons-lang.version>
         <commons-lang3.version>3.4</commons-lang3.version>
         <commons.version>2.5</commons.version>
-        <curator.version>2.11.0</curator.version>
-        <dropwizard.version>1.0.5</dropwizard.version>
+        <curator.version>4.0.1</curator.version>
+        <curator.test.version>2.12.0</curator.test.version>
```
arunmahadevan: is it ok to have different version just for test ?

pom.xml
arunmahadevan: Agree to keep the HDF version until the new SR version is pushed.

pom.xml
HeartSaVioR: I picked it up from Registry project, but it doesn't look like needed for Streamline. Will remove.

PR #1342, opened 2019-03-07
MINOR Apply actual latest dependencies we are depending on

HeartSaVioR:
We don't maintain the dependencies in sync with actual dependencies we are depending on, so the build succeeds on master branch whereas the build fails on internally.

This is a huge version update on Hadoop ecosystem, but most of our users are based on HDP/HDF products so if users depend on the right HDP/HDF version pair that should be OK.


HeartSaVioR:
I'll merge this once the build passes, since internal build is failing now.

HeartSaVioR:
Also merged into HDF-3.4.0.0 branch as well.

PR #1343, opened 2019-03-12
ISSUE-1344: Add missing classname argument in download-streamine-artifacts.py

### LINKED ISSUE #1344: Fix issue with download-streamine-artifacts.py
While trying to run download-streamine-artifacts.py, it throws below error.

Unrecognized option: --artifactRepositories
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.

arunmahadevan:
While trying to run download-streamine-artifacts.py, it throws below error.

Unrecognized option: --artifactRepositories
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.

It appears that the classname is missing in the java command that the script is trying to execute

#fixes #1344

### DIFF
--- a/bootstrap/download-streamine-artifacts.py
+++ b/bootstrap/download-streamine-artifacts.py
@@ -60,6 +60,7 @@ class ConfigStruct:
 
     def build_storm_submit_tool_command(self, artifacts):
         submit_tool_cmd_list = ["java", "-cp", self.get_storm_submit_tool_classpath()]
+        submit_tool_cmd_list.extend(["org.apache.storm.submit.command.DependencyResolverMain"])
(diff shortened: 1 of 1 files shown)


arunmahadevan:
@HeartSaVioR , I am not sure how the script is supposed to work, but looks like the classname argument is missing. Let me know if this is valid and I will create a github issue to associate with the PR.

HeartSaVioR:
@arunmahadevan Very weird... I ran script couple of times before raising pull request... I might be worked with copy of script... I guess we can file an issue and proceed. Thanks for the fix.

PR #1354, opened 2019-07-17
ISSUE-1353 Upgrade Schema Registry to 0.7.0

HeartSaVioR:
We've been depending on HDF specific version of SR which was a bit outdated and eventually be removed in nexus.

This patch upgrades SR to 0.7.0 which is newer official release and newer HDF versions are also relying on.


HeartSaVioR:
This is blocking PR CIs to pass. Once it passes CI, I'll merge this.

